PR #32, opened 2015-03-24
Add a setImmediate based .load() function to monomers for simpler API and recursion.

garlicnation:
Adds .load() for help recursing inside _parseHTML and to provide a cleaner API for using monomers.

By breaking up the recursion inside .load(), all parse work is moved to the end of the current task using a setImmediate polyfill.


azakus:
LGTM

PR #35, opened 2015-03-26
Change to espree for parsing.

garlicnation:
Parsing tests continue to pass.

The failure in safari has changed, but it seems to be core-asset related. I'll resolve in another PR.


justinfagnani:
:+1:

PR #39, opened 2015-03-30
Rename monomers to analyzer. Update jsdoc a bit.

garlicnation:
See: #37


azakus:
LGTM. I think we can just cut out Monomers entirely.

garlicnation:
I'll leave it in for a day or two until people fix up their code. Don't see any reason to break master.

PR #43, opened 2015-04-02
Restrict published contents

nevir:
Only include runtime code used by node projects


garlicnation:
Sweet! Can we make bower only include hydrolysis.{html,js}?

nevir:
Yeah I think bower has something similar

PR #44, opened 2015-04-02
Make fs resolver root config a prefix

azakus:
This allows for absolute urls with '/' to be resolved correctly.

Also fixes tests, which broke when the reorg occured for file loader


garlicnation:
LGTM. Can you squash and merge?

PR #57, opened 2015-04-11
Modify analyzer and noopresolver to support vulcanize.

garlicnation:


azakus:
LGTM. Should add a test for strings in NoopResolver tests.

PR #70, opened 2015-04-20
Rework isHtmlImportNode to use dom5 predicates.

garlicnation:
@jklein24 fyi.


jklein24:
Cool!

azakus:
LGTM. Minor nit, you could pull the predicate function out from the `isHtmlImportNode` and it would not make a new one every time.

azakus:
Took the initiative and added style finding with dom5 predicates in this PR, as well as rewrote to make a static predicate matcher.

PR #79, opened 2015-04-29
Update bower.json

GabrielDuque:


garlicnation:
This introduces a syntax error into bower.json.

PR #85, opened 2015-05-05
Support for behaviors. Closes #72.

### LINKED ISSUE #72: Support for behaviors
We should extract out a separate set of behaviors to go along side features.

nevir:


azakus:
Add a test for behaviors with behavior dependencies such as this one: https://github.com/PolymerElements/paper-behaviors/blob/master/paper-radio-button-behavior.html#L36-L40

nevir:
Hmm, let's do that as a separate PR. Isn't clear to me how we want to document that style of behavior?

PR #89, opened 2015-05-12
Event support

garlicnation:


lib/ast-utils/analyze-properties.js
```diff
@@ -65,7 +65,7 @@ var analyzeProperties = function(node) {
               prop.default = val.value;
             }
             else {
-              ;// TODO(atotic): need to make strings out of UnaryExpression, FunctionExpression
+              prop.default = "can't display";// TODO(atotic): need to make strings out of UnaryExpression, FunctionExpression
```
garlicnation: Leave default unset when we can't detect it. The API consumer can decide what to do with an undefined default.

garlicnation:
Do you know why the test is still failing? If it's wrong, it needs to be updated before we can merg.

PR #101, opened 2015-05-21
Find HTML comment above <script>

garlicnation:


hydrolysis-analyzer.html
```diff
@@ -91,6 +91,7 @@
       var options = {
         clean:  this.clean,
         filter: this.transitive ? function() { return false; } : null,
+        attachAst: true,
```
justinfagnani: what does attachAst mean?

lib/ast-utils/docs.js
```diff
@@ -107,7 +107,7 @@ function annotateElementHeader(descriptor) {
  */
 function annotateElement(descriptor) {
   if (!descriptor.desc && descriptor.type === 'element') {
-    descriptor.desc = _findElementDocs(descriptor.is, descriptor.domModule);
+    descriptor.desc = _findElementDocs(descriptor.is, descriptor.domModule, descriptor.scriptElement);
```
justinfagnani: line length

lib/ast-utils/docs.js
```diff
@@ -319,41 +319,51 @@ function cleanProperty(property) {
...
-  }
+function _findElementDocs(elementId, domModule, scriptElement) {
   // Note that we concatenate docs from all sources if we find them.
   var found = [];
 
-  // Do we have a HTML comment on the `<dom-module>`?
+  // Do we have a HTML comment on the `<dom-module>` or `<script>`?
   //
   // Confusingly, with our current style, the comment will be attached to
   // `<head>`, rather than being a sibling to the `<dom-module>`
```
justinfagnani: what's this? does this apply to `<script>` as well?

hydrolysis-analyzer.html
garlicnation: It hints hydrolysis to add AST references to a variety of places. I am convinced that there aren't any use cases that would want this off, now that I've tossed naive JSON serializability out of the window.

lib/ast-utils/docs.js
garlicnation: Due to the rules around where HTML comments end up in the AST, we have to go on a wild goose chase to find element docs. They can be: 
1) A sibling to the dom-module and part of body
2) The last child of head
2) the last child of the document

lib/ast-utils/docs.js
garlicnation: Fixed.

lib/ast-utils/docs.js
justinfagnani: are there parse5 rules, or html rules? can we file bugs?

justinfagnani:
lgtm

lib/ast-utils/docs.js
garlicnation: HTML rules.

PR #105, opened 2015-05-22
parse es6

garlicnation:


lib/ast-utils/js-parse.js
```diff
@@ -46,9 +46,78 @@ function traverse(visitorRegistries) {
...
+        spread: true,
+
+        // enable parsing classes
+        classes: true,
+
+        // enable parsing of modules
+        modules: true,
+
+        // enable React JSX parsing
+        jsx: true,
```
azakus: Do we need this?

garlicnation:
#127 replaces this.

PR #115, opened 2015-05-28
fix #114, better error parsing

### LINKED ISSUE #114: Better error reporting on bad js
Currently, you just get "error on line 33". polymer-catalog had a hard time figuring out which of the 100s of files it was.

atotic:


atotic:
I do not have a the javascript node linter has. 
The href gives you the filename.
The exception already has the line number in its message.

garlicnation:
You don't need the javascript node if the error has line/column information available as a property, just the HTML script element(or lack therof in a javascript file). If it's just in the error message, then what you've got is the best we can do for now.

atotic:
k. I can't be sure that the error throw will contain the information. This is what message looks like now:
Error parsing <script> tag http://localhost:8080/components/paper-item/paper-icon-item.html line:12 column:5 Error: Line 12: Unexpected identifier.
PTAL

PR #117, opened 2015-05-29
Move check of "loaded" into the metadata promise

### LINKED ISSUE #116: Import Traversal Ordering does not follow spec
Take this layout

- a
  - b
    - d
      - c
  - c

In the browser, the order should be "c, d, b, a", but hydrolysis will count "c" loading before "d" because of this looping behavior: https://github.com/Polymer/hydrolysis/blob/master/lib/analyzer.js#L400-L414

azakus:
Deeper branches of siblings can "load" before the current import
TODO: add tests
Fixes #116 Import Traversal Ordering does not follow spec


azakus:
TBR

garlicnation:
LGTM.

PR #118, opened 2015-05-31
FIx multiple dom-module doc bug

atotic:
New code differs from old significantly. I am not sure if this is the right thing to do. As is:

if dom-module is defined, uses first comment before dom-module
if dom-module is not defined, uses first comment before script
skips comments with @license


lib/ast-utils/docs.js
```diff
@@ -337,41 +341,23 @@ function _findElementDocs(elementId, domModule, scriptElement) {
...
+  return found
+    .filter(function(comment) {
+      // skip @license comments
+      if (comment.data && comment.data.indexOf('@license' === -1)) {
+        return true;
+      }
+      else {
+        return false;
+      }
+      })
```
garlicnation: Nit: unindent 2 spaces.

atotic:
PTAL

PR #126, opened 2015-06-10
adapted for publishing on bower

danielbaak:
As polymer is only distributed over bower (and zip files) it only makes sense to enable users to install hydrolysis over bower as well. 

Changes in bower.json simply make the package public (i.e. publishable) and define hydrolysis.js as the main file (attribute required in bower.json).

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,6 @@
 {
   "name": "hydrolysis",
-  "private": true,
+  "main": "hydrolysis.js",
   "ignore": [
     "lib/",
     "examples/",


danielbaak:
I signed it!

garlicnation:
Can you rebase+squash and remove the version bump? I like doing version bumps as independent commits directly on master along with the new tag.

danielbaak:
@garlicnation Thanks for the reminder ... I have to admit I was a bit ashamed of the dirty pull request. I should have taken the time to clean rebase it from the start.

PR #130, opened 2015-06-15
vulcanize-like loaded tree API

garlicnation:


azakus:
The `_inline` functions would need a transform function to be used by vulcanize, mostly to handle path resolving for the new locations.

The transform function would probably need an api of `transformFn(content, contentHref, docHref)` for path resolution.

PR #138, opened 2015-06-23
Copy properties from behaviors onto components and other behaviors.

garlicnation:


lib/analyzer.js
```diff
@@ -415,6 +424,19 @@ Analyzer.prototype._getDependencies = function _getDependencies(href, found, tra
   });
 };
 
+function matchesDocumentFolder(descriptor, href) {
+  console.log(descriptor.__ownerDocument, href);
+  return descriptor.__ownerDocument;
+}
```
rictic: Is this function used anywhere?

lib/analyzer.js
```diff
@@ -415,6 +424,19 @@ Analyzer.prototype._getDependencies = function _getDependencies(href, found, tra
...
+
+Analyzer.prototype.elementsForFolder = function elementsForFolder(href) {
+  return this.elements;
+};
+
+Analyzer.prototype.behaviorsForFolder = function behaviorsForFolder(href) {
+  return this.behaviors;
+};
```
rictic: Should these methods reference the href somehow?

lib/analyzer.js
garlicnation: No, removed it.

lib/analyzer.js
```diff
@@ -415,6 +424,33 @@ Analyzer.prototype._getDependencies = function _getDependencies(href, found, tra
...
+  }
+  var searchPath = searchDoc.pathname;
+  var lastSlash = searchPath.lastIndexOf("/");
+  if (lastSlash > 0) {
+    searchPath = searchPath.slice(0, searchPath.lastIndexOf("/"));
+  }
+  return descriptorDoc.pathname.startsWith(searchPath);
+}
+
+Analyzer.prototype.elementsForFolder = function elementsForFolder(href) {
```
garlicnation: These need tests, and still aren't used anywhere, but should be more robust.

garlicnation:
screen shot 2015-08-12 at 12 58 07 pm

lib/analyzer.js
```diff
@@ -415,6 +424,33 @@ Analyzer.prototype._getDependencies = function _getDependencies(href, found, tra
   });
 };
 
+function matchesDocumentFolder(descriptor, href) {
+  console.log(descriptor.__ownerDocument, href);
```
rictic: Consider removing the console.log if it's just for debugging.

lib/analyzer.js
```diff
@@ -415,6 +424,33 @@ Analyzer.prototype._getDependencies = function _getDependencies(href, found, tra
...
+  var descriptorDoc = url.parse(descriptor.__ownerDocument);
+  var searchDoc = url.parse(href);
```
rictic: consider naming these descriptorUrl and searchUrl rather than *Doc?

lib/analyzer.js
```diff
@@ -415,6 +424,33 @@ Analyzer.prototype._getDependencies = function _getDependencies(href, found, tra
...
+  if (!searchDoc || !searchDoc.pathname) {
+    return false;
+  }
+  var searchPath = searchDoc.pathname;
+  var lastSlash = searchPath.lastIndexOf("/");
+  if (lastSlash > 0) {
+    searchPath = searchPath.slice(0, searchPath.lastIndexOf("/"));
```
rictic: consider reusing lastSlash here instead of rerunning searchPath.lastIndexOf

lib/analyzer.js
```diff
@@ -605,9 +641,18 @@ Analyzer.prototype.annotate = function annotate() {
     this.elements.unshift(featureEl);
     this.elementsByTagName[featureEl.is] = featureEl;
   }
-
-  this.elements.forEach(docs.annotateElement);
-  this.behaviors.forEach(docs.annotateElement); // Same shape.
+  var behaviorsByName = this.behaviorsByName;
+  var elementHelper = function(descriptor){
+    console.log("In element helper", Object.keys(behaviorsByName));
```
rictic: Consider removing this console.log as well

lib/ast-utils/docs.js
```diff
@@ -106,7 +154,8 @@ function annotateElementHeader(descriptor) {
  * @param {Object} descriptor The element descriptor.
  * @return {Object} The descriptor that was given.
  */
-function annotateElement(descriptor) {
+function annotateElement(descriptor, behaviorsByName) {
+  console.log(descriptor.is);
```
rictic: Consider removing if for debugging and not part of intended output

test/js-parser.html
```diff
@@ -202,48 +202,56 @@
...
-        var loader = new hyd.Loader();
-        var resolver = new hyd.XHRResolver();
-        loader.addResolver(resolver);
-        loader.request("static/js-behaviors.js").then(function(content){
-          parsed = hyd._jsParse(content);
-          byName = {};
-          parsed.behaviors.forEach(function(behavior) {
-            byName[behavior.is] = behavior;
-          });
+        hyd.Analyzer.analyze("static/html-behaviors.html").then(function(analyzed){
```
rictic: With mocha you can just return a promise in a test rather than using done(). It will then wait on the promise to resolve successfully, much the same way that it now waits for done() to be called with no arguments. 

I bet this will work for setup as well.

rictic:
:+1: After the above comments.

lib/ast-utils/docs.js
```diff
@@ -96,6 +96,51 @@ function annotateElementHeader(descriptor) {
...
 }
 
+function matchByName(propa, propb) {
+  return propa.name == propb.name;
+}
+
+function copyProperties(from, to, behaviorsByName) {
+  if (from.properties) {
+    from.properties.forEach(function(fromProp){
+      if (fromProp.is in to.properties && !to.properties[fromProp.is].__fromBehavior) {
```
rictic: If `from.properties` and `to.properties` are both arrays then I think this is unlikely to ever be true, as the `in` operator checks keys, which for an array will be its indexes.

PR #147, opened 2015-08-01
Parsing HTML no longer fails catestrophically on empty script tags

joineral32:
We were using iron-component-page and encountered a problem where our iron-component-page would be completely inoperable if a Polymer element contained an empty script tag. I tracked this issue down to Hydrolysis trying to access script.childNodes[0] regardless of whether or not script.childNodes was empty, and passing the resulting undefined into the jsParse method in the empty case.


joineral32:
I signed it!

PR #148, opened 2015-08-01
Empty script tags no longer cause the analyzer to fail

joineral32:
We were using iron-component-page and encountered a problem where our iron-component-page would be completely inoperable if a Polymer element contained an empty script tag. I tracked this issue down to Hydrolysis trying to access script.childNodes[0] regardless of whether or not script.childNodes was empty, and passing the resulting undefined into the jsParse method in the empty case.


adewinter:
Nice!

lib/analyzer.js
```diff
@@ -303,7 +303,7 @@ Analyzer.prototype._processScript = function _processScript(script, href) {
   var parsedJs;
   if (!src) {
     try {
-      parsedJs = jsParse(script.childNodes[0].value);
+      parsedJs = jsParse((script.childNodes.length) ? script.childNodes[0].value : '/* */');
```
garlicnation: Why an empty comment block instead of just the empty string?

joineral32:
It is now using an empty string rather than an empty comment block.

garlicnation:
If you squash this into one commit and remove `npm-debug.log`, I'm happy to merge it.

Thanks for the contribution!

joineral32:
Squashed commits, rebased onto upstream and reran build. This should be all set. Thanks!

garlicnation:
Merged in #154

PR #154, opened 2015-08-10
merge changes from joineral32

garlicnation:


joineral32:
I am okay with these commits being contributed.

PR #157, opened 2015-08-24
Fix slash character compatibility in the default filter

jindav:
On Windows the file parsing, when using the default configuration, filters out scripts because the base path uses forward slash in the script tag and backslash on the filesystem.


jindav:
I signed it!

hypnoce:
+1

garlicnation:
Please rebase against master and confirm this passes the full test suite.

jindav:
Branch is up-to-date, ci tests are green. Should be safe to merge. Thanks and sorry for the delay.

PR #159, opened 2015-09-08
Fix #158

### LINKED ISSUE #158: Documentation for a behavior is duplicated when trying to document extended behaviors
_From @jjones-sg on August 12, 2015 18:2_

When trying to document a behavior that extends existing behaviors I end up using the @polymerBehavior annotation in two places; once above the implementation of the behavior, and once before declaring the array with the existing behaviors. So basically code is as follows:

```
/** @polymerBehavior BehaviorName */
BehaviorImpl = {
...
}

/** @polymerBehavior **/
BehaviorName = [ExistingBehavior, BehaviorImpl];
```
From looking at the Polymer code base I have concluded that this is the correct way to annotate this.

The problem arises when I use iron-component-page to create a documentation viewer for all of my elements and behaviors. In the select in the top left corner my behavior is listed twice because of the duplicate annotation. When I try moving the annotations around, I either get a single incomplete documentation page (properties or behaviors will be missing), or multiple complete ones depending on how the annotations are set up.

Ideally I would like a single, complete documentation page.

Is there anyway to do that in the current annotation scheme?

My iron-component-page looks like:
```
<iron-component-page src="elements.html"></iron-component-page>
```

Where elements.html is simply a series of html imports pulling in the elements and behaviors.

_Copied from original issue: PolymerElements/iron-component-page#46_

atotic:
The bug symptom:
- iron-component-page would show 2 copies of a behavior
- only happens when behavior definition was split into 2 parts, impl + array

Cause:
- hydrolysis was merging behaviors, but still reinserting the merged behavior, so we ended up with 2 copies.

Fix:
- do not insert behavior into the list if already there

### DIFF
--- a/lib/ast-utils/behavior-finder.js
+++ b/lib/ast-utils/behavior-finder.js
@@ -180,7 +180,8 @@ module.exports = function behaviorFinder() {
       if (isSimpleBehaviorArray(behaviorExpression(node))) {
         // TODO(ajo): Add a test to confirm the presence of `properties`.
         if (!currentBehavior.properties) currentBehavior.properties = [];
-        behaviors.push(currentBehavior);
+        if (behaviors.indexOf(currentBehavior) === -1)
+          behaviors.push(currentBehavior);
         currentBehavior = null;
       }
     },


garlicnation:
Taking a look at the build process and catalog today, will likely merge this afternoon.

garlicnation:
Merged! Thanks!
https://github.com/Polymer/hydrolysis/commit/7c4cd399ddc0a0fb6cf79e60e3852278b3e1232b

PR #160, opened 2015-09-09
🐐 Add missing wct dep

samccone:
since we are using wct, but did not have it defined in our dependencies npm test would error out unless a user had web-component-tester installed globally or had manually installed it locally.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,8 @@
     "jsdoc-to-markdown": "^1.0.3",
     "jshint": "^2.7.0",
     "mocha": "^2.2.4",
-    "watch": "latest"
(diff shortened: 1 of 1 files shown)


addyosmani:
`wct` is referenced in the test section of https://github.com/Polymer/hydrolysis/blob/master/package.json. With that in mind this change lgtm, but will defer to the repo owner.

addyosmani:
`wct` is referenced in the test section of https://github.com/Polymer/hydrolysis/blob/master/package.json. With that in mind this change lgtm, but will defer to the repo owner.

garlicnation:
LGTM.

PR #166, opened 2015-09-22
store whole js document ASTs

garlicnation:


lib/analyzer.js
```diff
@@ -39,6 +39,20 @@ var EMPTY_METADATA = {elements: [], features: [], behaviors: []};
  */
 
 /**
+ * espree's representation of a parsed html document
+ * @typedef {Object} JSAST
```
mikesamuel: How about, instead of using `{Object}` here, something like the below?

```js
/**
 * An AST Node as defined by
 * [SpiderMonkey's reflect API](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/Parser_API#Node_objects)
 * and as produced by esmangle.
 * @constructor
 */
function Node() {}
/** @type {string} */
Node.prototype.type;
/** @type {?SourceLocation} */
Node.prototype.loc;

/**
 * Also defined by SpiderMonkey's reflect API.
 * @constructor
 */
function SourceLocation() {}
/** @type {string | null} */
SourceLocation.prototype.source;
/** @type {Position | null} */
SourceLocation.prototype.start;
/** @type {Position | null} */
SourceLocation.prototype.end;

/**
 * Also defined by SpiderMonkey's reflect API.
 * @constructor
 */
function Position() {}
/** one-indexed  @type {number} */
Position.prototype.line;
/** zero-indexed @type {number} */
Position.prototype.column;
```

lib/analyzer.js
```diff
@@ -143,6 +155,17 @@ var Analyzer = function Analyzer(attachAST,
   this.parsedDocuments = {};
 
   /**
+   * A map, keyed by path, of JS script ASTs.
+   *
+   * If the path is an HTML file with multiple scripts, the entry will be an array of scripts.
+   *
+   * @type {Object<string,Array<ParsedJS>}
```
mikesamuel: I think this is missing a `>`.

lib/analyzer.js
```diff
@@ -351,6 +373,17 @@ Analyzer.prototype._processScript = function _processScript(script, href) {
       }.bind(this));
       this.behaviors = this.behaviors.concat(parsedJs.behaviors);
     }
+    if (!this.parsedScripts[href]) {
```
mikesamuel: `If !(Object.hasOwnProperty.call(this.parsedScripts, href))`

If `href` is `"constructor"`, we shouldn't go adding properties to the global Object function.

lib/analyzer.js
```diff
@@ -615,8 +648,8 @@ Analyzer.prototype.getLoadedAst = function getLoadedAst(href, loaded) {
  * @return {Object}
  */
 Analyzer.prototype.nodeWalkDocuments = function nodeWalkDocuments(predicate) {
-  for (var href in this._parsedDocuments) {
-    var match = dom5.nodeWalk(this._parsedDocuments[href].ast, predicate);
+  for (var href in this.parsedDocuments) {
+    var match = dom5.nodeWalk(this.parsedDocuments[href], predicate);
```
azakus: Shouldn't this be `this.parsedDocuments[href].ast`?

lib/analyzer.js
```diff
@@ -631,8 +664,8 @@ Analyzer.prototype.nodeWalkDocuments = function nodeWalkDocuments(predicate) {
  */
 Analyzer.prototype.nodeWalkAllDocuments = function nodeWalkDocuments(predicate) {
   var results = [];
-  for (var href in this._parsedDocuments) {
-    var newNodes = dom5.nodeWalkAll(this._parsedDocuments[href].ast, predicate);
+  for (var href in this.parsedDocuments) {
+    var newNodes = dom5.nodeWalkAll(this.parsedDocuments[href].ast, predicate);
```
azakus: Shouldn't this be `this.parsedDocuments[href]` minus the `.ast`?

azakus:
LGTM minus the comment.

PR #168, opened 2015-09-25
Support absolute paths in FSResolver

nomego:
This patch checks if the `local` path is an absolute path, and if so, does not prefix it with the `root` path.
`polylint` for example always supplies both, causing absolute paths to fail.
Since SublimeLinter mostly uses an input stream (not gotten it to work) or an absolute path, merging this would make PolymerLabs/polylint#22 a lot easier to accomplish.

### DIFF
--- a/lib/loader/fs-resolver.js
+++ b/lib/loader/fs-resolver.js
@@ -31,6 +31,16 @@ function getFile(filePath, deferred, secondPath) {
 }
 
 /**
+ * Returns true if `p` is an absolute path.
+ * TODO: Replace with path.isAbsolute() once node 0.10 support is no longer needed.
+ * @param  {string}  p path
+ * @return {boolean}
+ */
+function isAbsolute(p) {
+  return path.normalize(p + '/') === path.normalize(path.resolve(p) + '/');
+}
+
+/**
  * Returns true if `patha` is a sibling or aunt of `pathb`.
  * @return {boolean}
  */
@@ -90,7 +100,7 @@ FSResolver.prototype = {
       if (base) {
         local = path.relative(base, local);
       }
-      if (root) {
+      if (root && !isAbsolute(local)) {
         local = path.join(root, local);
       }
 


lib/loader/fs-resolver.js
```diff
@@ -90,7 +99,7 @@ FSResolver.prototype = {
       if (base) {
         local = path.relative(base, local);
       }
-      if (root) {
+      if (root && !isAbsolute(local)) {
```
garlicnation: How about [path.isAbsolute](https://nodejs.org/api/path.html#path_path_isabsolute_path)?

nomego: Well I didn't want to introduce a dependency on node 0.12 since current Ubuntu 15.04 ships with 0.10. That would have to include the npm package https://www.npmjs.com/package/path-is-absolute which felt too invasive for this.

garlicnation: I'm OK merging as-is, then. Could you add a comment with a TODO to remove it once 0.10 support is no longer needed?

nomego:
Commit has been updated but tests timed out.

addyosmani:
Changes LGTM. The timing out tests (currently) appear to be an issue across all of the PRs against the repo and I don't think they're related to what you've changed.

lib/loader/fs-resolver.js
```diff
@@ -31,6 +31,16 @@ function getFile(filePath, deferred, secondPath) {
...
 
 /**
+ * Returns true if `p` is an absolute path.
+ * TODO: Replace with path.isAbsolute() once node 0.10 support is no longer needed.
+ * @param  {string}  p path
+ * @return {boolean}
+ */
+function isAbsolute(p) {
+  return path.normalize(p + '/') === path.normalize(path.resolve(p) + '/');
+}
```
sindresorhus: Just use the polyfill for now: https://github.com/sindresorhus/path-is-absolute The above will definitely break in some cases.

garlicnation:
@nomego I had to revert this, since it was causing mocha test failures. I have since reconfigured CI to run linting/mocha, and would love to keep working with you on supporting sublimelinter.

PR #172, opened 2015-09-29
startsWith doesn't exist

garlicnation:


azakus:
ugh, how'd that typo get in there?
LGTM

garlicnation:
It was experimental API that nothing used(until today when I updated the catalog build a bit).

PR #173, opened 2015-09-29
without an href, the pseudoElements were being dropped by matchesDocu…

garlicnation:
…mentFolder


frankiefu:
I was just going to ping you about the issue I saw with `@pseudoElements` and then you just fixed it.  :-)
LGTM :+1:

PR #176, opened 2015-09-30
redirectResolver

garlicnation:


lib/loader/redirect-resolver.js
```diff
@@ -0,0 +1,91 @@
...
+// jshint node:true
+'use strict';
+
+var fs = require('fs');
+var path = require('path');
+var url = require('url');
+
+var FSResolver = require('./fs-resolver');
+
+function ProtocolRedirect(protocol, hostname, path, redirectPath){
```
azakus: Maybe use an object as an argument map like FSResolver does.

azakus:
LGTM with nit

PR #180, opened 2015-10-05
Support absolute paths in FSResolver

nomego:
Since #168 was merged, I can't amend that commit anymore.
This is an alternative implementation with the https://github.com/sindresorhus/path-is-absolute ponyfill/polyfill.

This support is needed for PolymerLabs/polylint#22


nomego:
Actually looking at the tests, this patch tries to circumvent the fs absolute path things.
Didn't realize this was by design, fixing with a check in the polylint tool instead.

PR #182, opened 2015-10-07
observer support

garlicnation:


lib/analyzer.js
```diff
@@ -364,7 +364,7 @@ Analyzer.prototype._processScript = function _processScript(script, href) {
       line += err.lineNumber;
       col += err.column;
       var message = "Error parsing script in " + href + " at " + line + ":" + col;
-      message += "\n" + err.description;
+      message += "\n" + err.stack;
```
azakus: This change seems unrelated to the observer support. Not saying it needs to be removed, just double checking that this was intended.

azakus:
LGTM with nit.

PR #184, opened 2015-10-09
ordering

garlicnation:


azakus:
I assume this just changes the behavior order to process from "deepest" to "shallowest"?
LGTM then.

PR #191, opened 2015-11-10
add string resolver

garlicnation:


lib/loader/string-resolver.js
```diff
@@ -0,0 +1,48 @@
...
+StringResolver.prototype = {
+
+  /**
+   * @param {string}    uri      The absolute URI being requested.
+   * @param {!Deferred} deferred The deferred promise that should be resolved if
+   *     this resolver handles the URI.
+   * @return {boolean} Whether the URI is handled by this resolver.
+   */
+  accept: function(uri, deferred) {
+    if (!this.url.test) {
```
rictic: Can this.url be a regexp? You've annotated it as just a string.

lib/loader/string-resolver.js
```diff
@@ -0,0 +1,48 @@
...
+ * The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+
+// jshint node:true
+'use strict';
+
+/**
+ * A resolver that resolves to null any uri matching config.
```
rictic: "resolves to null"? Do you mean "resolves to config"?

lib/loader/string-resolver.js
```diff
@@ -0,0 +1,48 @@
...
+   * @param {!Deferred} deferred The deferred promise that should be resolved if
+   *     this resolver handles the URI.
+   * @return {boolean} Whether the URI is handled by this resolver.
+   */
+  accept: function(uri, deferred) {
+    if (!this.url.test) {
+      if (uri.search(this.url) == -1) {
+        return false;
+      }
+    } else if (!this.url.test(uri)) return false;
```
rictic: This control flow is impressively short, but hard to reason about. This is longer and dumber, but more obviously right IMO:

```javascript
if (this.url.test) { 
  // this.url is a regex
  if (!this.url.test(uri) {
    return false;
  }
} else {
  // this.url is a string
  if (uri.search(this.url) == -1) {
    return false;
  }
}

File: lib/analyzer.js [hunk 4]
```diff
@@ -204,6 +205,9 @@ Analyzer.analyze = function analyze(href, options) {
                         require('./loader/fs-resolver') :
                         require('./loader/xhr-resolver');
   loader.addResolver(new PrimaryResolver(options));
+  if (options.content) {
```
rictic: It would be helpful to document `options.content` on the annotation for LoadOptions above

rictic:
LGTM!

PR #202, opened 2015-11-22
Remove duplicated function

rictic:
This should be a noop, since second assignment wins.

Discovered during typescriptification

### DIFF
--- a/lib/analyzer.js
+++ b/lib/analyzer.js
@@ -710,21 +710,6 @@ Analyzer.prototype.getLoadedAst = function getLoadedAst(href, loaded) {
  * @param  {Object} predicate A dom5 predicate.
  * @return {Object}
  */
-Analyzer.prototype.nodeWalkDocuments = function nodeWalkDocuments(predicate) {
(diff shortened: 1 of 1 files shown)


lib/analyzer.js
```diff
@@ -710,21 +710,6 @@ Analyzer.prototype.getLoadedAst = function getLoadedAst(href, loaded) {
...
-  }
-  return null;
-};
-
-/**
- * Calls `dom5.nodeWalkAll` on each document that `Anayzler` has laoded.
- * @param  {Object} predicate A dom5 predicate.
- * @return {Object}
- */
 Analyzer.prototype.nodeWalkAllDocuments = function nodeWalkDocuments(predicate) {
```
garlicnation: The correct fix is to rename this function to `nodeWalkAllDocuments`. `nodeWalk` and `nodeWalkAll` are distinct.

rictic: Ouch, I totally missed that these are actually named differently...

PR #207, opened 2016-01-12
support binaryexpresson with +

garlicnation:
Fixes https://github.com/PolymerLabs/polylint/issues/63

### DIFF
--- a/lib/ast-utils/ast-value.js
+++ b/lib/ast-utils/ast-value.js
@@ -104,6 +104,16 @@ function objectExpressionToValue(obj) {
 }
 
 /**
+ * BinaryExpressions are of the form "literal" + "literal"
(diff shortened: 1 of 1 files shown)


azakus:
LGTM

lib/ast-utils/ast-value.js
```diff
@@ -136,7 +146,10 @@ function expressionToValue(valueExpression) {
       return identifierToValue(valueExpression);
     case 'MemberExpression':
       return memberExpressionToValue(valueExpression);
+    case 'BinaryExpression':
+      return binaryExpressionToValue(valueExpression);
     default:
+      console.log(valueExpression)
```
cdata: Lingering console log?

PR #209, opened 2016-01-23
Convert hydrolysis to typescript

rictic:
- typechecks cleanly with no warnings or errors and only a couple of places where I had to use `any` to opt out of type checking
- added babel to the build as well so that we can use all ES2015 features + [async/await](https://jakearchibald.com/2014/es7-async-functions/)
- made the minimal reasonable changes. e.g. converted class-patterned objects to ES2015 classes, left the rest alone
  - That said, there was one function whose promise logic was so convoluted that I needed to convert it to async/await just so I could get it to type check.
- Added type annotations for a few node modules and estree. Will upstream those, there's a package manager for type annotations. `npm run init` will install the ones we use.
- Introduces one potentially backward-incompat change for node dependents. It seems like the semantics for exporting a default symbol in ES2015 are different than `module.exports = ...`
  - so `new require('hydrolysis/loader/fs-resolver')` should become `new require('hydrolysis/loader/fs-resolver').FSResolver`
  - `require('hydrolysis').FSResolver` and friends are unchanged, it's only when code is requiring stuff from `hydrolysis/...`, which strikes me as unlikely to break many people , as we re-export most symbols from the main `index.js`

Open questions:
  - still looking for a good documentation generator for typescript. I poked around with typedoc briefly but it didn't work very well for me

I'm excluding hydrolysis.js from this PR so that the diff isn't swamped, `npm build` still works fine, as does `npm test`.

/cc @justinfagnani @azakus


rictic:
Hm, travis isn't happy, looks like we may need to have npm build run before npm install.

azakus:
LGTM as a first pass. Cleaning some stuff up to ES6+ syntax can come later. Great work!

rictic:
So I checked in on the travis run now that it's cooled down. I'd made an oversight in supporting node <4 as a consumer of the full library, not just the compiled hydrolysis.js. Long story short, rather than the pipeline being `tsc | browserify+babel` it's `tsc | babel | browserify`. I moved the actual source to `src/` and told babel to put the old-node-friendly ES5 into `lib/`.

Now passes clean in node v0.10 and v0.12.

Keeping index.ts as a typescript file was now more trouble than it was worth, so it's not just index.js. I also needed to add another import of babel-polyfill for folks that are coming in to the loaders directly without going through index (which they probably shouldn't do, but the tests do, so..).

azakus:
`typings` isn't listed as a `devDependency`

rictic:
Oops, fixed

.babelrc
```diff
@@ -0,0 +1,3 @@
+{
```
justinfagnani: I've heard feedback from projects where I've included a .babelrc that it interferes with dependents that also use babel. Babel is somehow pulling in all the .babelrc's even in node_modules. Are you using a require hook, or can you include the config in a build.

custom_typings/dom5.d.ts
```diff
@@ -0,0 +1,56 @@
+declare module 'dom5' {
```
justinfagnani: Add a todo to upstream this to dom5? Or an issue in dom5?

custom_typings/estree.d.ts
```diff
@@ -0,0 +1,411 @@
+declare module 'estree' {
```
justinfagnani: Is this not a good typescript definition for estree: https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/estree

index.js
```diff
@@ -7,23 +7,23 @@
...
-  jsdoc:             require('./lib/ast-utils/jsdoc'),
-  Loader:            require('./lib/loader/file-loader'),
-  NoopResolver:      require('./lib/loader/noop-resolver'),
-  RedirectResolver:  require('./lib/loader/redirect-resolver'),
-  XHRResolver:       require('./lib/loader/xhr-resolver'),
-  StringResolver:    require('./lib/loader/string-resolver'),
-  _jsParse:          require('./lib/ast-utils/js-parse'),
-  _importParse:      require('./lib/ast-utils/import-parse'),
-};
+exports.Analyzer = require('./lib/analyzer').Analyzer;
```
justinfagnani: Why isn't this a .ts file?

src/analyzer.ts
```diff
@@ -0,0 +1,766 @@
...
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+// jshint node: true
+'use strict';
+
+import * as dom5 from 'dom5';
+import * as url from 'url';
+import * as estree from 'estree';
+
```
justinfagnani: nit: extra newline?

src/analyzer.ts
```diff
@@ -0,0 +1,766 @@
...
+ */
+interface LocError extends Error{
+  location: {line: number; column: number};
+  ownerDocument: string;
+}
+
+/**
+ * A database of Polymer metadata defined in HTML
+ *
+ * @param  {boolean} attachAST  If true, attach a parse5 compliant AST
```
justinfagnani: Can the parameter docs be moved to the constructor?

justinfagnani: Also, is there any hope of getting tsc to generate the jsdoc/jscompiler type annotations?

src/ast-utils/ast-value.ts
```diff
@@ -10,21 +10,23 @@
 // jshint node: true
 'use strict';
 
+import * as ESTree from 'estree';
```
justinfagnani: style nit: lowercase library names?

src/ast-utils/ast-value.ts
```diff
@@ -10,21 +10,23 @@
...
+import {LiteralValue} from './descriptors';
+
 // useful tool to visualize AST: http://esprima.org/demo/parse.html
 
 /**
  * converts literal: {"type": "Literal", "value": 5,  "raw": "5" }
  * to string
  */
-function literalToValue(literal) {
+function literalToValue(literal:ESTree.Literal) {
```
justinfagnani: Should this have return type string, or is the comment above wrong?

src/ast-utils/ast-value.ts
```diff
@@ -106,9 +108,12 @@ function objectExpressionToValue(obj) {
...
  * BinaryExpressions are of the form "literal" + "literal"
  */
-function binaryExpressionToValue(member) {
+function binaryExpressionToValue(member: ESTree.BinaryExpression):(number|string) {
   if (member.operator == "+") {
-    return expressionToValue(member.left) + expressionToValue(member.right);
+    // We need to cast to `any` here because, while it's usually not the right
+    // thing to do to use '+' on two values of a mix of types because it's
+    // unpredictable, that is what the original code we're evaluating does.
+    return <any>expressionToValue(member.left) + expressionToValue(member.right);
```
justinfagnani: can you not cast to a union of `number|string`?

src/ast-utils/ast-value.ts
```diff
@@ -116,45 +121,38 @@ function binaryExpressionToValue(member) {
 /**
  * MemberExpression references a variable with name
  */
-function memberExpressionToValue(member) {
+function memberExpressionToValue(member:ESTree.MemberExpression) {
```
justinfagnani: return type string?

src/ast-utils/ast-value.ts
```diff
@@ -35,43 +37,43 @@ function unaryToValue(unary) {
...
-function identifierToValue(identifier) {
+function identifierToValue(identifier:ESTree.Identifier):string {
   return identifier.name;
 }
 
 /**
  * Function is a block statement.
  */
-function functionDeclarationToValue(fn) {
+function functionDeclarationToValue(fn:ESTree.FunctionDeclaration) {
```
justinfagnani: TO me, it's unclear enough what a "value" might be that return types would help, even if it's `any`. Union types, in the case the types can be primitives or AST nodes, would be helpful too.

src/ast-utils/ast-value.ts
```diff
@@ -116,45 +121,38 @@ function binaryExpressionToValue(member) {
...
- * valueExpression example:
- * { type: "Literal",
+ * Tries to get the value of an expression. Returns undefined on failure.
  */
-function expressionToValue(valueExpression) {
+export function expressionToValue(valueExpression:ESTree.Node):LiteralValue {
   switch(valueExpression.type) {
     case 'Literal':
-      return literalToValue(valueExpression);
+      return literalToValue(<ESTree.Literal>valueExpression);
```
justinfagnani: I thought TS has a way to infer the types here. Can you add the value of Node.type to the interfaces for the node classes so that the value check on .type acts as a type guard?

src/ast-utils/behavior-finder.ts
```diff
@@ -0,0 +1,233 @@
...
+        const n = <estree.VariableDeclaration>node;
+        return n.declarations.length > 0 ? n.declarations[0].init : null;
+    }
+  }
+
+  /**
+   * checks whether an expression is a simple array containing only member
+   * expressions or identifiers.
+   */
+  function isSimpleBehaviorArray(expression: estree.Node) {
```
justinfagnani: return type boolean

src/ast-utils/behavior-finder.ts
```diff
@@ -0,0 +1,233 @@
...
+
+  var currentBehavior: BehaviorDescriptor = null;
+  var propertyHandlers: PropertyHandlers = null;
+
+  /**
+   * merges behavior with preexisting behavior with the same name.
+   * here to support multiple @polymerBehavior tags referring
+   * to same behavior. See iron-multi-selectable for example.
+   */
+  function mergeBehavior(newBehavior:BehaviorDescriptor) {
```
justinfagnani: return type?

src/ast-utils/behavior-finder.ts
```diff
@@ -0,0 +1,233 @@
...
+    var symbol = getName();
+    // Quickly filter down to potential candidates.
+    if (!comment || comment.indexOf('@polymerBehavior') === -1) {
+      if (symbol !== templatizer) {
+        return;
+      }
+    }
+
+
+    currentBehavior = {
```
justinfagnani: I'd love to see classes here, but I suppose that's for another CL

src/ast-utils/behavior-finder.ts
```diff
@@ -0,0 +1,233 @@
...
+import * as esutil from './esutil';
+import * as jsdoc from './jsdoc';
+import * as analyzeProperties from './analyze-properties';
+import * as astValue from './ast-value';
+import * as estree from 'estree';
+import {Visitor} from './fluent-traverse';
+import {declarationPropertyHandlers, PropertyHandlers} from './declaration-property-handlers';
+import {BehaviorDescriptor, LiteralValue} from './descriptors';
+
+export function behaviorFinder() {
```
justinfagnani: What's the return type here? Is there an interface for it, or can behaviorFinder be converted to a class?

src/ast-utils/declaration-property-handlers.ts
```diff
@@ -21,48 +24,47 @@ var analyzeProperties = require('./analyze-properties');
...
     },
-    observers: function(node) {
+    observers: function(node: estree.Node) {
       if (node.type != 'ArrayExpression') {
         return;
       }
-      for (var i=0; i<node.elements.length; i++) {
-        var v = astValue.expressionToValue(node.elements[i]);
+      const arrNode = <estree.ArrayExpression>node;
+      for (let element of arrNode.elements) {
```
justinfagnani: I see here you converted the loop to let/of. I've wanted to see that in several places above. Any reason why you did it here? Maybe a followup CL can change all var to let...

src/ast-utils/descriptors.ts
```diff
@@ -0,0 +1,93 @@
+import * as estree from 'estree';
```
justinfagnani: 'use strict'? Unless we can use TS 1.8 which emits it automatically.

src/ast-utils/descriptors.ts
```diff
@@ -0,0 +1,93 @@
...
+import * as jsdoc from './jsdoc';
+import * as dom5 from 'dom5';
```
justinfagnani: do we have an import ordering style? I'm used to alphabetical, but in ES6, what would be alphabetized? The module name/path?

src/ast-utils/docs.ts
```diff
@@ -11,12 +11,12 @@
 
 // jshint node:true
 
-var jsdoc = require('./jsdoc');
-
-var dom5 = require('dom5');
+import * as jsdoc from './jsdoc'
+import * as dom5 from 'dom5'
+import {FeatureDescriptor, FunctionDescriptor, PropertyDescriptor, Descriptor, ElementDescriptor, BehaviorsByName, EventDescriptor, BehaviorDescriptor} from './descriptors'
```
justinfagnani: > 80 cols. Imports support line breaks, I hope.

src/ast-utils/docs.ts
```diff
@@ -53,7 +53,7 @@ var HANDLED_TAGS = [
  * @param {Object} descriptor The descriptor node to process.
  * @return {Object} The descriptor that was given.
  */
-function annotate(descriptor) {
+export function annotate(descriptor: Descriptor) {
```
justinfagnani: return type

src/ast-utils/docs.ts
```diff
@@ -69,7 +69,7 @@ function annotate(descriptor) {
 /**
  * Annotates @event, @hero, & @demo tags
  */
-function annotateElementHeader(descriptor) {
+export function annotateElementHeader(descriptor: ElementDescriptor) {
```
justinfagnani: return type? In this case there's one path that returns and the rest don't... weird.

src/ast-utils/docs.ts
```diff
@@ -90,19 +90,16 @@ function annotateElementHeader(descriptor) {
...
     });
   }
 }
 
-function matchByName(propa, propb) {
-  return propa.name == propb.name;
-}
-
-function copyProperties(from, to, behaviorsByName) {
+function copyProperties(from:ElementDescriptor, to:ElementDescriptor, behaviorsByName:BehaviorsByName) {
```
justinfagnani: line length

src/ast-utils/docs.ts
```diff
@@ -90,19 +90,16 @@ function annotateElementHeader(descriptor) {
...
   if (from.properties) {
     from.properties.forEach(function(fromProp){
-      for (var toProp, i = 0; i < to.properties.length; i++) {
+      for (var toProp:PropertyDescriptor, i = 0; i < to.properties.length; i++) {
```
justinfagnani: let/of?

justinfagnani: Just kidding... we can do this later

src/ast-utils/docs.ts
```diff
@@ -220,7 +219,7 @@ function annotateElement(descriptor, behaviorsByName) {
  * @param {Object} descriptor behavior descriptor
  * @return {Object} descriptor passed in as param
  */
-function annotateBehavior(descriptor, behaviorsByName) {
+export function annotateBehavior(descriptor:BehaviorDescriptor) {
```
justinfagnani: return type

src/ast-utils/docs.ts
```diff
@@ -172,7 +171,7 @@ function mixinBehaviors(descriptor, behaviorsByName) {
  * @param {Object} descriptor The element descriptor.
  * @return {Object} The descriptor that was given.
  */
-function annotateElement(descriptor, behaviorsByName) {
+export function annotateElement(descriptor: ElementDescriptor, behaviorsByName: BehaviorsByName) {
```
justinfagnani: return type

src/ast-utils/docs.ts
```diff
@@ -259,8 +258,9 @@ function _annotateEvent(descriptor) {
  * @param {boolean} ignoreConfiguration If true, `configuration` is not set.
  * @return {Object} The descriptior that was given.
  */
-function annotateProperty(descriptor, ignoreConfiguration) {
-  annotate(descriptor);
+function annotateProperty(descriptor:PropertyDescriptor, ignoreConfiguration:boolean) {
```
justinfagnani: return type

src/ast-utils/feature-finder.ts
```diff
@@ -0,0 +1,59 @@
...
+
+export function featureFinder() {
+  /** The features we've found. */
+  var features : FeatureDescriptor[] = [];
+
+  function _extractDesc(feature:FeatureDescriptor, node:estree.CallExpression, parent: estree.Node) {
+    feature.desc = esutil.getAttachedComment(parent);
+  }
+
+  function _extractProperties(feature:FeatureDescriptor, node:estree.CallExpression, parent: estree.Node) {
```
justinfagnani: line length

src/ast-utils/feature-finder.ts
```diff
@@ -0,0 +1,59 @@
...
+import {Visitor} from './fluent-traverse';
+import * as estree from 'estree';
+
+var numFeatures = 0;
+
+export function featureFinder() {
+  /** The features we've found. */
+  var features : FeatureDescriptor[] = [];
+
+  function _extractDesc(feature:FeatureDescriptor, node:estree.CallExpression, parent: estree.Node) {
```
justinfagnani: line length

src/ast-utils/feature-finder.ts
```diff
@@ -0,0 +1,59 @@
...
+import * as esutil from './esutil';
+import {FeatureDescriptor} from './descriptors';
+import {Visitor} from './fluent-traverse';
+import * as estree from 'estree';
+
+var numFeatures = 0;
+
+export function featureFinder() {
+  /** The features we've found. */
+  var features : FeatureDescriptor[] = [];
```
justinfagnani: what's our style with colons separating type annotations? I actually prefer this style, as you know :)

src/ast-utils/feature-finder.ts
```diff
@@ -0,0 +1,59 @@
...
+    const objExpr = <estree.ObjectExpression>featureNode;
+    if (!objExpr.properties) return;
+
+    feature.properties = objExpr.properties.map(esutil.toPropertyDescriptor);
+  }
+
+  var visitors: Visitor = {
+
+    enterCallExpression: function enterCallExpression(node, parent) {
+      if (!esutil.matchesCallExpression(<estree.MemberExpression>node.callee, ['Polymer', 'Base', '_addFeature'])) {
```
justinfagnani: line length

package.json
```diff
@@ -16,25 +16,33 @@
     "lib/"
   ],
   "scripts": {
-    "build": "browserify -d -r ./index.js:hydrolysis -o hydrolysis.js",
-    "build:watch": "watch 'npm run build' ./lib",
-    "release": "browserify -r ./index.js:hydrolysis -o hydrolysis.js",
+    "init": "typings install",
+    "clean": "rm -rf node_modules/ typings/ lib/ && find src | grep \\.js$ | xargs rm",
+    "build": "tsc && babel -q -d lib src && browserify -d -r ./index.js:hydrolysis -o hydrolysis.js",
```
justinfagnani: I've usually seen lib/ used for either the source or for checked in dependencies. I'd rather have build output go to build/ or dist/

package.json
```diff
@@ -16,25 +16,33 @@
     "lib/"
   ],
   "scripts": {
-    "build": "browserify -d -r ./index.js:hydrolysis -o hydrolysis.js",
-    "build:watch": "watch 'npm run build' ./lib",
-    "release": "browserify -r ./index.js:hydrolysis -o hydrolysis.js",
+    "init": "typings install",
+    "clean": "rm -rf node_modules/ typings/ lib/ && find src | grep \\.js$ | xargs rm",
```
justinfagnani: This won't work on windows. What do you think about adding gulp?

package.json
```diff
@@ -16,25 +16,33 @@
...
+    "build": "tsc && babel -q -d lib src && browserify -d -r ./index.js:hydrolysis -o hydrolysis.js",
+    "build:watch": "watch 'npm run build' ./src",
+    "release": "browserify -r ./index.js:hydrolysis -o hydrolysis.js -t babelify",
     "apidocs": "node_modules/jsdoc-to-markdown/bin/cli.js {index.js,lib/{analyzer,loader/*}.js} > API.md",
-    "test": "jshint index.js lib/ && npm run build && wct && mocha test/test.js"
+    "test": "npm run build && wct && mocha test/test.js",
```
justinfagnani: Will the plain `tsc` invocation fail on warnings? I'd like the tests to be strict here.

justinfagnani:
I made a bunch of small comments, but basically this looks good. My only real questions are about the build and test setup.

.babelrc
rictic: Interesting. We only call babel in one place in package.json, so this is easy to remove.

custom_typings/dom5.d.ts
rictic: Done.

custom_typings/estree.d.ts
rictic: I swear I looked for such a thing, I don't know how I missed this.

Annoyingly, now that I've written my own I can see some ways in which theirs are worse.

Added a TODO to merge and rationalize.

index.js
rictic: Because this file is super simple, and it's not in the src/ dir, so the choice was either just make it a js file or special case it in the build rules. I chose the simpler option

src/analyzer.ts
rictic: Done.

src/analyzer.ts
rictic: Done, and yeah, I believe there's folks working on going from typescript -> jscompiler annotations

src/ast-utils/ast-value.ts
rictic: Done, here and elsewhere.

src/ast-utils/ast-value.ts
rictic: `expressionToValue` returns `string | number | boolean | RegExp`, and generally speaking you shouldn't `+` such things together, but we actually do want to do that in this case. The easiest way is to just cast the left value to `any` which appears to just make the local type inference give up completely.

src/ast-utils/ast-value.ts
rictic: Added return types to every function in this file.

src/ast-utils/ast-value.ts
rictic: Done.

src/ast-utils/ast-value.ts
rictic: Done.

src/ast-utils/ast-value.ts
rictic: AIUI being able to use the `type` property as a type guard based on the string literal value isn't quite supported yet.

src/ast-utils/behavior-finder.ts
rictic: Done.

src/ast-utils/behavior-finder.ts
rictic: Done.

src/ast-utils/behavior-finder.ts
rictic: :+1:

src/ast-utils/behavior-finder.ts
rictic: Added a simple interface. It can be converted into a class.

src/ast-utils/declaration-property-handlers.ts
rictic: :+1: to migrating all of them eventually. I *think* there was a reason why I did it here? I can't seem to find one though..

src/ast-utils/descriptors.ts
rictic: Done. Also added copyright.

src/ast-utils/docs.ts
rictic: Done.

src/ast-utils/docs.ts
rictic: Done

src/ast-utils/docs.ts
rictic: Did 80 cols in this whole file

src/ast-utils/docs.ts
rictic: :+1: to the change, and to punting.

src/ast-utils/docs.ts
rictic: Done

src/ast-utils/docs.ts
rictic: Done

src/ast-utils/docs.ts
rictic: Done

src/ast-utils/behavior-finder.ts
rictic: Removed the interface, as it's only used in one place. Just letting the type inference work for now, class later.

src/ast-utils/feature-finder.ts
rictic: Done

src/ast-utils/feature-finder.ts
rictic: Done

src/ast-utils/feature-finder.ts
rictic: At the moment we're going with clang-format's

```typescript
let x: Type;
```

src/ast-utils/feature-finder.ts
rictic: Done

package.json
rictic: My goal here was to keep back-compat with the lib directory's layout as it was previously

package.json
rictic: Not at all thrilled unless it also got us faster builds (which admittedly should be possible, save probably for the final browserify step).

package.json
rictic: Yep! And `tsc` is configured to be extra pedantic.

rictic:
All comments addressed, PTAL!

justinfagnani:
LGTM++

src/ast-utils/docs.ts
```diff
@@ -135,11 +138,14 @@ function copyProperties(from, to, behaviorsByName) {
   if (!from.behaviors) {
```
garlicnation: Line 133 captures a bug fixed in: https://github.com/Polymer/hydrolysis/pull/211

garlicnation:
LGTM!

This is a huge boon for hydrolysis.

rictic:
Cool, should I make another commit with a build version of hydrolysis.js before merging?

rictic:
Chatted with Justin, sounds like the thing to do is to update hydrolysis.js here, so that's what I've done.

Thanks for the reviews everyone!

PR #211, opened 2016-02-02
stop abusing the prototype chain. Simplify event serialization

garlicnation:


justinfagnani:
I assume the reason here is that your serializer can deal with references easier than prototypes?

LGTM

src/ast-utils/docs.ts
```diff
@@ -130,8 +130,10 @@ function copyProperties(
       if (fromEvent.__fromBehavior) {
         return;
       }
-      var newEvent = Object.create(fromEvent);
-      newEvent.__fromBehavior = from.is;
+      var newEvent = {__fromBehavior: from.is};
+      Object.keys(fromEvent).forEach(function(eventField){
```
azakus: This can just be `Object.assign(newEvent, fromEvent)`

justinfagnani: Not without a polyfill

justinfagnani: @garlicnation are you including an es6 polyfill library?

azakus: Babel won't transpile it?

justinfagnani: It can't, you could do:

```
let x = Object.assign;
// things that can't be statically determined
x(a, b);
```

It needs to be added to `Object`

azakus: Sigh, nvm then.

azakus:
LGTM2

PR #213, opened 2016-02-04
Autoinit typings

garlicnation:
Init will run on compile if necessary.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
   "scripts": {
     "init": "typings install",
     "clean": "rm -rf node_modules/ typings/ lib/ && find src | grep \\.js$ | xargs rm",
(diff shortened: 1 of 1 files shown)


garlicnation:
@rictic Is there a way to produce hydrolysis.js in a stable way? 

I'm seeing diffs like this:
```
   "_args": [
     [
       "doctrine@^0.7.0",
-      "/usr/local/google/home/rictic/open/hydrolysis"
+      "/Users/ajo/workspace/hydrolysis"
     ]
```

rictic:
As far as I can tell those paths are from `browserify`, as it's not in the output of tsc + babel.

I tried https://github.com/substack/bundle-collapser but that didn't seem to help the problem.

rictic:
This change LGTM.

:+1:

PR #214, opened 2016-02-04
node users need lib/ directory

garlicnation:


rictic:
Which users need lib? npm users? bower users?

garlicnation:
npm users need lib. I would like to make releases tags of master, not leaves that have been built. This CL enables releasing by running `git pull` and `mversion`

justinfagnani:
It doesn't seem like modifying .gitignore is the right way to do this. You can include other files in a release, no? Or release from a branch that includes the files?

leodido:
+1

rictic:
`npm publish` goes off of working directory, not git state, so this change is unnecessary for publishing versions up to npm. We would need to check things in like this if we want to support people doing an npm install directly from git/github.

rictic:
Closing this PR. It seems like this isn't the direction that we're going with our built node modules.

PR #216, opened 2016-02-06
Fix type error in loader/string-resolver.ts

justinfagnani:


rictic:
As discussed offline, this works fine in tsc >=1.8. Has to do with function overloading and type guards.

PR #217, opened 2016-02-09
allow users of hydrolysis-analyzer.html to specify the loader

garlicnation:


hydrolysis-analyzer.html
```diff
@@ -65,6 +65,10 @@
         notify:   true,
       },
 
+      permissive: {
```
justinfagnani: what about a `resolver` property that defaults to `'xhr'`?

garlicnation: fsResolver is never correct for this usecase. I'd rather not let someone specify it.

justinfagnani: It is in Electron, where Designer runs :)

garlicnation:
@justinfagnani PTAL

justinfagnani:
What does "permissive loader" mean here. The PR title and change don't connect for me.

garlicnation:
xhr loader, fs loader, and a permissive version of the loader that doesn't throw when it fails to load

justinfagnani:
LGTM

fwiw, it's the word "permissive" in the title that confused me. I looked for something to do with a permissive loader in the change. This change just let's the user specify the loader in general, and thus the very general change makes sense.

In the future could resolver accept an loader instance rather than a string?

garlicnation:
It could accept an instance, but that would require an API change to hydrolysis.Analyzer.analyze

justinfagnani:
LGTM

fwiw, it's the word "permissive" in the title that confused me. I looked for something to do with a permissive loader in the change. This change just let's the user specify the loader in general, and thus the very general change makes sense.

In the future could resolver accept an loader instance rather than a string?

PR #225, opened 2016-02-25
Properly manipulate script tags with external scripts

nomego:
Previous handling of external script files breaks the DOM
structure and creates invalid text nodes according to dom5.

This manipulates existing script element to maintain DOM
structure and creates a text node with dom5 method.

Fixes Polymer/hydrolysis#185


garlicnation:
LGTM

PR #228, opened 2016-03-07
Remove an unused variable

ankon:


ankon:
Oops. Forgot to pull, and worked on an old working copy.

PR #204, opened 2015-12-02
Implement ES6 classes parsing

jgautheron:
This commit implements ES6 classes parsing.

It currently supports:
- properties
- methods
- behaviors

:warning: A few limitations remain due to the nature of the parsing.
The parser expects the keywords (`is`, `properties`...) to be defined using an assignment.

Works :+1: 

```javascript
class Foo {
  beforeRegister() {
    this.is = 'foo';
    this.properties = {
      cover: String,
    };
  }
}
```

Not supported :-1:

```javascript
class Foo {
    get is() {
      return 'foo';
    }
}
```

Two WCT tests don't pass, I'm out of time right now to go further, any help or direction would be appreciated.

```
chrome 47 (32/0/2)
firefox 42 (32/0/2)
safari 9.0.1 (31/0/3)    
```

PR #229, opened 2016-03-08
Support for ES6 classes

leodido:
Following from #204.

Initial work. Feedbacks, suggestions very welcome.

cc @garlicnation

### DIFF
--- a/custom_typings/estraverse.d.ts
+++ b/custom_typings/estraverse.d.ts
@@ -10,6 +10,7 @@ declare module 'estraverse' {
     break?: ()=>void;
     remove?: ()=>void;
     skip?: ()=>void;
+    keys?: {};
   }
   export enum VisitorOption {
       Skip, Break, Remove
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "init": "typings install",
     "clean": "rm -rf typings/ lib/ && find src | grep \\.js$ | xargs rm",
     "compile": "if [ ! -f typings/main.d.ts ]; then npm run init; fi; tsc && babel --presets es2015 -q -d lib src",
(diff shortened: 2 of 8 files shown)


RoXuS:
`enterMethodDefinition` is not setted, I propose this solution to have methods with their annotations.
https://github.com/RoXuS/hydrolysis/blob/es6/src/ast-utils/element-finder.ts#L74

I see two other problems :
- The properties are setted but the jsdoc is not parsed.
- The behaviors doesn't work.

garlicnation:
The biggest thing this needs before any more work is tests. Try to come up with some sample cases and get a few tests failing so that we can cleanly work towards a working implementation.

RoXuS:
Ok I have openned a new PR with test see #230

justinfagnani:
Can't the tests be added to this PR?

RoXuS:
This PR does not work with ES6 classes.
I have lot of js error when I try to see the <iron-component-page> of a composent with this branch.

It is not just the tests.

leodido:
Some considerations follow.

1. Missing support for **factoryImpl** method

    Only used when creating and element with its constructor, not used when element is created by markup or via document.createElement, called after the element initialisation phase (local DOM created, default values set, ...)

2. Missing support for **listeners**

3. Missing support for **extends**

4. Not sure about the use of getters for **observers**.

    Is it better to declare them with a getter ? Or is it better to declare them inside **beforeRegister** with **this.observers** ?
    Same holds for **listeners** and **extends**.

5. Also a **setter** for **behaviors** could be required (used by Polymer to flatten the list/hierarchy of behaviors)

I have created a gist containing a [skeleton ES2015 class for Polymer elements](https://gist.github.com/leodido/8a5aa2b1b3a84e190798).

garlicnation:
@leodido I merged this outside of the PR

PR #236, opened 2016-03-29
make es5 and es6 classes play nicely together.

garlicnation:
Add some escape routes for walking the tree without an element.
ES6 classes now play well with ES5 elements.

ES6 support caused any es5-defined elements to throw during parsing.

### DIFF
--- a/src/ast-utils/element-finder.ts
+++ b/src/ast-utils/element-finder.ts
@@ -40,8 +40,8 @@ export function elementFinder() {
       this.classDetected = true;
       element = {
         type: 'element',
-        desc: esutil.getAttachedComment(parent),
-        events: esutil.getEventComments(parent).map(function(event) {
+        desc: esutil.getAttachedComment(node),
+        events: esutil.getEventComments(node).map(function(event) {
           return { desc: event };
         }),
         properties: [],
@@ -58,9 +58,13 @@ export function elementFinder() {
         element = null;
         propertyHandlers = null;
       }
+      this.classDetected = false;
     },
 
     enterAssignmentExpression: function enterAssignmentExpression(node, parent) {
+      if (!element) {
+        return;
+      }
       const left = <estree.MemberExpression>node.left;
       if (left && left.object && left.object.type !== 'ThisExpression') {
         return;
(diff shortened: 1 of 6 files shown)


test/static/es6-support.js
```diff
@@ -46,3 +47,82 @@ class A {
     this.data = 'Hello World';
   }
 }
+
+/**
+ * I am a description of test-element.
+ *
+ * @hero /path/to/hero.png
```
justinfagnani: what are these annotations?

justinfagnani:
Can you add a little more description about what this PR is accomplishing?

test/static/es6-support.js
garlicnation: @hero gives an icon for use in catalog or documentation

src/ast-utils/element-finder.ts
```diff
@@ -40,8 +40,8 @@ export function elementFinder() {
       this.classDetected = true;
       element = {
         type: 'element',
-        desc: esutil.getAttachedComment(parent),
-        events: esutil.getEventComments(parent).map(function(event) {
+        desc: esutil.getAttachedComment(node),
```
justinfagnani: is this related to class support, or is it fixing something else?

test/static/es6-support.js
justinfagnani: is it used in the test?

src/ast-utils/element-finder.ts
garlicnation: Class support - the parent node of the class declaration is the Program, so all events were getting associated with every es6 class in the file.

garlicnation: The test case I added tickled this.

test/static/es6-support.js
justinfagnani: I'm just confused why this element is being added. This style of element was already supported, right?

garlicnation: No, it's cargo-cult copied from another test - I like having it in to make sure it doesn't throw or cause strange behavior, but I can reduce this case significantly.

justinfagnani: got it, thanks

justinfagnani:
LGTM

PR #247, opened 2016-06-18
fs-resolver: don't add root if path is already absolute

fiws:


fiws:
seems like that behaviour is wanted...

PR #249, opened 2016-07-14
Get most of the tests running on node

justinfagnani:


justinfagnani:
All tests running and passing on Travis and Sauce now.

.travis.yml
```diff
@@ -1,10 +1,26 @@
 language: node_js
 node_js:
-  - "0.10"
-  - "0.12"
   - "4"
```
garlicnation: 5?

justinfagnani: I'd like to move our support policy to latest LTS and latest stable. It'll probably still work on 5, but the extra latency in testing isn't worth it, IMO.

.travis.yml
```diff
@@ -1,10 +1,26 @@
...
+    sources:
+    - google-chrome
+    packages:
+    - google-chrome-stable
+  sauce_connect: true
 script:
-  - npm run testNode
+- mocha
+- xvfb-run wct
+- if [ "${TRAVIS_PULL_REQUEST}" = "false" ]; then wct -s 'default'; fi
```
justinfagnani: Anyone know how to only run the Sauce tests for one of the node versions? No need to run them twice.

garlicnation: Something like: `node --version | grep v6`?

garlicnation: Even better: TRAVIS_NODE_VERSION

https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables

garlicnation:
LGTM

.travis.yml
justinfagnani: docs don't say on what it's set to, probably `node --version` I guess

justinfagnani: This appears to work:

```
- if node --version | grep v6; then xvfb-run wct; fi
- if [ "${TRAVIS_PULL_REQUEST}" = "false" ] && node --version | grep v6; then wct -s 'default'; fi
```

.travis.yml
rictic: +1

PR #250, opened 2016-07-17
2.0 work in progress

justinfagnani:


src/css/css-document.ts
```diff
@@ -0,0 +1,21 @@
...
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+
+import {Node, NodeVisitor} from 'shady-css-parser';
+
+import {Document} from '../parser/document';
+
+export class CssDocument extends Document<Node, NodeVisitor> {
+  type = 'css';
```
rictic: This results in a constructor being generated that uses `...args` as a parameter, which isn't supported in node v4.

rictic:
It looks like the tests don't run, e.g. missing `../lib/analyzed-document` in `test/analyzer_test.js`

src/analyzer.ts
```diff
@@ -8,762 +8,153 @@
...
+    if (!this._loader.canLoad(url)) {
+      throw new Error(`Can't load URL: ${url}`);
     }
-    return results;
-  };
+    // Use an immediately executed async function to create the final Promise
+    // synchronously so we can store it in this._documents before any other
+    // async operations to avoid any race conditions.
+    let promise = (async () => {
+      let content = await this._loader.load(url);
```
rictic: `this._loader.load(url)` is called before `this._documents.set(url, promise)` is.

I confirmed this with the following script:

```javascript
console.log('first');
const p = (async () => {
  console.log('second');
  await Promise.resolve();
  console.log('fourth');
})();
console.log('third');
```

I think to be confident that this is correct we should start the function with `await Promise.resolve()`

src/parser/parser.ts
```diff
@@ -0,0 +1,13 @@
...
+ * Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+
+export interface Parser<T> {
+  parse(contents: string, url: string): T;
```
rictic: You know, if this returned `Promise<T>` we could use something like [webworker-threads](https://www.npmjs.com/package/webworker-threads) to do parallel parsing in a thread pool. It's not unlikely that parsing will be the slowest part of most analysis runs, so it may be worth the effort, or at least not shutting the door on the possibility.

src/analyzer.ts
```diff
@@ -8,762 +8,153 @@
...
+import {Document} from './parser/document';
+import {Parser} from './parser/parser';
+import {
+  Descriptor,
+  DocumentDescriptor,
+  ImportDescriptor,
+} from './ast/ast';
+import {UrlLoader} from './url-loader/url-loader';
+
+export interface AnalyzerInit {
```
rictic: Style note: I find that, when there's only one such interface in a file, it's clearer to call it just `Options`. No need to stutter the name since this name is scoped to the current module, and `Options` is the more common name for this kind of object in the js world than `Init`.

src/javascript/javascript-document.ts
```diff
@@ -0,0 +1,45 @@
...
+        let callback = visitor[callbackName];
+        if (callback) {
+          return callback(node, parent) || undefined;
+        }
+      }
+    }
+
+    traverse(<Node><any>this.ast, {
+      enter(node, parent) {
+        return applyFinders(`enter ${node.type}`, node, parent);
```
rictic: The spaces are unnecessary and will make it harder to define a visitor won't it? estree types are PascalCase

src/html/html-parser.ts
```diff
@@ -0,0 +1,80 @@
...
+
+const isInlineJSScript = p.AND(
+  p.hasTagName('script'),
+  p.NOT(p.hasAttr('src')),
+  p.OR(
+    p.NOT(p.hasAttr('type')),
+    p.hasAttrValue('type', 'text/javascript'),
+    p.hasAttrValue('type', 'application/javascript')
+  )
+);
```
rictic: I think it would be encouraging to the community if we at least failed with a clear error message when encountering an inline es2015 module

rictic: Oh wait, nm, we don't need to care about whether a given script is in module mode or not at parse time. Only time it would show up might be in inferring whether a toplevel identifier is global.

rictic:
I'd recommend pulling in the tools-common tslint.config and integrate it into your editor sooner rather than later (the atom plugin is high quality).

rictic:
I've confirmed that it compiles cleanly with typescript 2.0 beta. I know you don't want to do too much at once, but it seems like a very transparent upgrade, no need to turn on all the new stricter checks immediately.

One nice part is that we could get rid of typings and the init phase entirely. Just `npm install --save-dev @types/parse5 @types/acorn @types/node`

src/ast/descriptor.ts
```diff
@@ -0,0 +1,8 @@
+import * as jsdoc from '../ast-utils/jsdoc';
+
+export type LiteralValue = string|number|boolean|RegExp;
+
+export interface Descriptor {
```
rictic: For incrementality we're gonna need to know the path/url of this descriptor. We'll also need to be able to walk to the the entities that depend on this entity (e.g. if this entity is a behavior then we'd need to be able to walk to all the upstream entities that mix it in to refresh them based on its updated state).

src/html/html-entity-finder.ts
```diff
@@ -0,0 +1,17 @@
...
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+
+import {ASTNode} from 'parse5';
+import {Descriptor} from '../ast/ast';
+import {HtmlDocument, HtmlVisitor} from './html-document';
+import {EntityFinder} from '../entity/entity-finder';
+
+export interface HtmlEntityFinder
```
rictic: Arg, name clash, HTML entities are totally a thing.

src/html/html-parser.ts
```diff
@@ -0,0 +1,80 @@
...
+ * This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+
+import * as dom5 from 'dom5';
+import * as parse5 from 'parse5';
+import {ASTNode} from 'parse5';
```
rictic: Does parse5 support a parse -> modify -> serialize workflow? I remember that many parsers did not support modification and serialization back when writing polyup, but I don't recall how much I evaluated parse5.

gulpfile.js
```diff
@@ -0,0 +1,143 @@
...
+ * This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+
+'use strict';
+
+const depcheck_lib = require('depcheck');
```
rictic: Try to use tools-common

src/url-loader/package-url-resolver.ts
```diff
@@ -0,0 +1,65 @@
...
+  componentDir: string;
+  hostname: string;
+
+  constructor(options?: PackageUrlResolverOptions) {
+    options = options || {};
+    this.componentDir = options.componentDir || 'bower_components/';
+    this.hostname = options.hostname || null;
+  }
+
+  canResolve(url: string): boolean {
```
rictic: Document what resolution already needs to have taken place on this url. Also we want tests.

src/analyzer.ts
```diff
@@ -8,762 +8,161 @@
...
+  urlLoader?: UrlLoader;
+  parsers?: Map<string, Parser<any>>;
+  entityFinders?: Map<string, EntityFinder<any, any, any>[]>;
 }
 
 /**
  * A database of Polymer metadata defined in HTML
```
rictic: Update this comment

src/ast/import-descriptor.ts
```diff
@@ -0,0 +1,19 @@
...
+ * Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+ 
+export class ImportDescriptor {
+  type: string;
```
rictic: Consider giving some examples as part of the type, then noting that we're falling back to `string` for extensibility. i.e.

```typescript
type: 'html-import' | 'script-import' | string;
```

src/ast/import-descriptor.ts
```diff
@@ -0,0 +1,19 @@
...
+  url: string;
```
rictic: Document what resolution must have happened to this URL at this point, and what it is relative to.

src/analyzer.ts
```diff
@@ -8,762 +8,161 @@
...
+      let visitPromise: Promise<void>;
+      visitPromise = (async () => {
+        let visitors: any = [];
+        // Collect visitors and return the batch Promise
+        let visit = (visitor: any) => {
+          visitors.push(visitor);
+          return visitPromise;
+        };
+        finderPromises = finders.map((f) => f.findEntities(document, visit));
+        document.visit(visitors);
```
rictic: What happens if a visitor registers a visitor?

test/analyzer_test.js
```diff
@@ -0,0 +1,178 @@
...
+    test('loads and parses a JavaScript document', () => {
+      return analyzer.load('/static/js-elements.js')
+        .then((doc) => {
+          assert.instanceOf(doc, JavaScriptDocument);
+          assert.equal(doc.url, '/static/js-elements.js');
+        });
+    });
+
+    test('returns a Promise that rejects for non-existant files', () => {
+      return analyzer.load('/static/not-found')
```
rictic: Recommend using a promise inverter function. e.g.

```javascript
function invert(promise) {
  return new Promise((resolve, reject) => {
    promise.then(
        () => reject(new Error('expected promise to reject but it resolved'), 
        resolve
    );
  );
}
```

src/css/css-document.ts
justinfagnani: you mean because of not specifying a ctor?

src/html/html-entity-finder.ts
justinfagnani: yeah... I want an answer to the Entity/Descriptor question too. Let's go to the design doc and try to come with names for the resolved types too.

src/javascript/javascript-document.ts
justinfagnani: This is a bug... don't have proper test coverage here

src/ast/import-descriptor.ts
justinfagnani: Let's move this over to the url-loader repo. I'll pull this package out.

src/javascript/javascript-document.ts
justinfagnani: Fixed

src/analyzer.ts
justinfagnani: will do. I used `Init` because it's the initialization options...

src/analyzer.ts
justinfagnani: done

src/ast/import-descriptor.ts
rictic: ImportDescriptor is still part of hydrolysis, yeah?

src/url-loader/package-url-resolver.ts
rictic: This should move over to the other repo.

src/entity/find-entities.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    // Let `findEntities` continue after calls to visit().then()
+    currentDoneCallback();
+  };
+
+  // The callback passed to `findEntities()`
+  function visit(visitor: any) {
+    visitors.push(visitor);
+    if (!runner) {
+      runner = Promise.resolve().then(runVisitors);
```
rictic: This is a painfully complex system, and I predict that most readers will be unable to reason successfully about what is executed and when.

What about a basic loop of:

```typescript
let finderPromises = finders.map((f) => f.findEntities(document, visit));

// visitors may register more visitors, so wait until we're done.
while(visitors) {
  const visitorsThisRun = visitors.slice();
  visitors.length = 0;
  await batchVisit(visitorsThisRun);
}

let nestedEntities = await Promise.all(finderPromises);
// ...
```

src/entity/find-entities.ts
```diff
@@ -0,0 +1,93 @@
...
+    } catch (error) {
+      visitError = error;
+    }
+
+    // Let `findEntities` continue after calls to visit().then()
+    currentDoneCallback();
+  };
+
+  // The callback passed to `findEntities()`
+  function visit(visitor: any) {
```
rictic: We should throw if visit is called after findEntities returns.

src/javascript/javascript-document.ts
```diff
@@ -25,19 +25,18 @@ export class JavaScriptDocument extends Document<Program, Visitor> {
      */
     function applyFinders(callbackName: string, node: Node, parent: Node) {
       for (let visitor of visitors) {
-        let callback = visitor[callbackName];
-        if (callback) {
-          return callback(node, parent) || undefined;
+        if (callbackName in visitor) {
+          return visitor[callbackName](node, parent) || undefined;
```
rictic: Why `|| undefined` here?

src/entity/find-entities.ts
justinfagnani: This speaks to why I built the full complex scheduler here. As long as `visit()` returns a Promise, I think the expectation will be that `findEntities()` can have arbitrary async operations. It's not much of a change to go from this:

```javascript
findEntities(document, visit) {
  let entities = [];
  await visit((node) => process(node));
  return entities;
}
```

which would be valid, to

```javascript
findEntities(document, visit) {
  let entities = [];
  await setup();
  await visit((node) => process(node));
  return entities;
}
```

which would be invalid.

I wanted to try to avoid hard-to-remember rules like "visit() can only be called once", or "visit() must be called synchronously"

rictic:
I'd like to be able to write the tests in typescript. It's a bit tricky though if we're moving all compiled output to the src directory. Putting a nested tsconfig inside the test directory has its own issues though, as importing anything from lib/ doesn't quite work.

I think the ticket will be to generate a .d.t.s file for hydrolysis and then use that npm package that lets us import it by name from within this package.

src/entity/find-entities.ts
justinfagnani: I'm pretty sympathetic to this point, and considered it. The reason I went down this path is because I felt the complexity could be localized here, and documented and tested well, while keeping `findEntities()` implementations simpler. Once this is written and burned in, it will need much less attention than the many upcoming `findEntities()` implementations, possibly written by many different authors, and dependent on different libraries. Especially because `findEntities()` will be part of the eventual user-exposed plugin API, it should have a simple contract. This approach preserves `visit()` as a pretty straightforward async call, even if it's complex behind the scenes.

There are other options that also preserve simplicity in `findEntities()`: `visit()` could be synchronous, `visit()` could be async but not batch, or `findEntities()` could be two phase: it returns a set of visitors, and then has some kind of finish callback.

I still think for the moment that the behavior that this implementation gives `findEntities()` is close to ideal, but I could be convinced otherwise.

gulpfile.js
```diff
@@ -0,0 +1,143 @@
...
+
+function task(name, deps, impl) {
+  if (gulp.hasTask(name)) {
+    throw new Error(
+        `A task with the name ${JSON.stringify(name)} already exists!`);
+  }
+  gulp.task(name, deps, impl);
+}
+
+// module.exports.init = function() {
```
garlicnation: Kill these lines?

src/analyzer.ts
```diff
@@ -8,762 +8,139 @@
...
-  _content: {[path:string]: string} = {};
-
-  /**
-   * @param  {boolean} attachAST  If true, attach a parse5 compliant AST
-   * @param  {FileLoader=} loader An optional `FileLoader` used to load external
-   *                              resources
-   */
-  constructor(attachAST:boolean, loader:FileLoader) {
-    this.loader = loader;
+  private _parsers: Map<string, Parser<any>> = new Map<string, Parser<any>>([
```
garlicnation: I'd rather see a more constrained interface here instead of <any>

src/analyzer.ts
```diff
@@ -8,762 +8,139 @@
...
-  };
+   * @param {string} url the location of the file to analyze
+   * @return {Promise<DocumentDescriptor>}
+   */
+  async analyze(url: string): Promise<DocumentDescriptor> {
+    if (this._documentDescriptors.has(url)) {
+      return this._documentDescriptors.get(url);
+    }
+    let promise = (async () => {
+      let document = await this.load(url);
```
garlicnation: The outer function is async. Can you await this.load at the top level and avoid the nesting here?

src/analyzer.ts
```diff
@@ -8,762 +8,139 @@
...
-   * @param {Object.<string,boolean>=} loaded An object keyed by already loaded documents.
-   * @return {Promise.<DocumentAST>}
+   * Loads and parses a single file, deduplicating any requrests for the same
+   * URL.
    */
-  getLoadedAst(href:string, loaded:{[href:string]:boolean}) {
-    if (!loaded) {
-      loaded = {};
+  async load(url: string): Promise<Document<any, any>> {
+    // TODO(justinfagnani): normalize url
```
garlicnation: For normalization to be effective, the entire document tree needs to get resolved against the same Base.

src/ast/import-descriptor.ts
justinfagnani: Replied to the wrong comment...

src/analyzer.ts
rictic: Can you clarify this comment?

src/analyzer.ts
rictic: We want to set the promise just for the loading and analyzing into `this._documentDescriptors`, so no.

src/css/css-document.ts
justinfagnani: Done

src/analyzer.ts
justinfagnani: Done

src/analyzer.ts
justinfagnani: Fixed

justinfagnani: DOne

src/ast/descriptor.ts
justinfagnani: Punt for now, we'll be forced to figure this out when the first clients come online.

src/html/html-parser.ts
justinfagnani: Yes

gulpfile.js
justinfagnani: Next PR will use tools-common

src/url-loader/package-url-resolver.ts
justinfagnani: Punt on moving to other repo until it's published to NPM

src/analyzer.ts
justinfagnani: Fixed with new scheduler. Added tests

src/ast/import-descriptor.ts
justinfagnani: Done now

src/ast/import-descriptor.ts
justinfagnani: Done

test/analyzer_test.js
justinfagnani: done

gulpfile.js
justinfagnani: done

src/analyzer.ts
rictic: There's not a common interface between a CSS document node, a JS document node, an HTML document node, produced by three different parsers.

src/analyzer.ts
justinfagnani: ack

.travis.yml
```diff
@@ -4,23 +4,17 @@ node_js:
   - "6"
 sudo: required
 before_script:
-- npm install -g web-component-tester mocha bower
+- npm install -g mocha bower
```
rictic: Don't need mocha here either

.travis.yml
```diff
@@ -4,23 +4,17 @@ node_js:
...
 - bower install
 - npm run build
```
rictic: can remove npm run build if you change the script to `npm test`

.travis.yml
justinfagnani: done

.travis.yml
justinfagnani: done

PR #252, opened 2016-07-23
Use the Typescript 2.0 beta

rictic:
* use the `@types/` npm namespace rather than the `typings` tool, eliminating the need for any init steps beyond `npm install`
* express the file globs using the new `include` syntax in tsconfig.json
* update our estree type definitions to be way more useful, letting us do stuff like: 
```typescript
    import * as estree from 'estree';

    let node: estree.Node;

    if (node.type === 'Identifier') {
      // The type of `node` inside this block is narrowed to estree.Identifier.
      console.log(`Node is identifier with name: ${node.name}`);
    } else if (node.type === 'CallEpression' /* Note the typo */) {
      // This is impossible, so the line below is an error. node has been narrowed
      // here to the type `never`.
      console.log(`Node is a CallExpression: ${node.type}`);
    }
```
* compiles clean


tsconfig.json
```diff
@@ -9,45 +9,12 @@
...
         "suppressImplicitAnyIndexErrors": true
     },
-    "filesGlob": [
+    "include": [
         "src/analyzer.ts",
         "src/**/*.ts",
-        "typings/main.d.ts",
         "custom_typings/**.d.ts"
     ],
     "atom": {
```
justinfagnani: I don't think we need this anymore

custom_typings/estree.d.ts
```diff
@@ -1,12 +1,17 @@
 // TODO(rictic): Upstream/merge with https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/estree/estree.d.ts
 
 declare module 'estree' {
-  export interface Node {
-    type: string;
+  interface BaseNode {
     leadingComments?: Comment[];
```
justinfagnani: can you try adding back the `type` field?

justinfagnani:
LGTM

custom_typings/estree.d.ts
rictic: That works, but it would paper over mistakes in BaseNode leaf interfaces. It's better that it be missing so that the type of `Node.type` is always a union of string literal types.

Added a comment describing this.

tsconfig.json
rictic: Removed.

PR #254, opened 2016-07-25
Update package.json for upcoming polymer-analyzer rename

justinfagnani:


package.json
```diff
@@ -1,19 +1,16 @@
...
     "type": "git",
-    "url": "https://github.com/Polymer/hydrolysis.git"
+    "url": "https://github.com/Polymer/polymer-analyzer.git"
   },
   "main": "index.js",
   "files": [
-    "hydrolysis-analyzer.html",
-    "hydrolysis.html",
-    "hydrolysis.js",
     "index.js",
```
rictic: This is now `lib/index.js` so it can be removed.

rictic: Should add custom_typings/ to this array though, so downstream installers get all the definitions they need.

rictic:
LGTM once the above two comments are addressed.

package.json
justinfagnani: done and done

PR #257, opened 2016-07-26
Fix our exported type definitions.

rictic:
With these changes, hydrolysis works properly in a downstream typescript
project when `npm link`ed in.

### DIFF
--- a/custom_typings/main.d.ts
+++ b/custom_typings/main.d.ts
@@ -1,1 +1,7 @@
+/// <reference path="./doctrine.d.ts" />
+/// <reference path="./dom5.d.ts" />
+/// <reference path="./escodegen.d.ts" />
+/// <reference path="./espree.d.ts" />
+/// <reference path="./estraverse.d.ts" />
 /// <reference path="./estree.d.ts" />
+/// <reference path="./shady-css-parser.d.ts" />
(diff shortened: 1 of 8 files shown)


custom_typings/main.d.ts
```diff
@@ -1 +1,7 @@
+/// <reference path="./doctrine.d.ts" />
```
justinfagnani: why do we have a main.d.ts file anyway? Can't we just include `custom_typings`?

src/index.ts
```diff
@@ -8,6 +8,8 @@
  * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
  */
 
+/// <reference path="../custom_typings/main.d.ts" />
```
justinfagnani: hmm... seems a bit unfortunate to expose our custom_typings. We should work on upstreaming.

custom_typings/main.d.ts
justinfagnani: nvm. I see

justinfagnani:
LGTM

src/index.ts
rictic: Agreed, I've figured out DefinitelyTyped enough to get my PRs accepted. I'm putting together a PR for estree right now.

PR #258, opened 2016-07-26
Clang format all the things

rictic:
So I tried out vscode to check whether it's typescript in general or atom in particular that's slow with the improved estree typings. Looks like it's atom in particular, as vscode doesn't lose a beat.

I also discovered that it's very easy to get it to run clang-format on save, so I decided I'd take a look to see how it would look if we adopted clang-format for the repo.

### DIFF
--- a/custom_typings/escodegen.d.ts
+++ b/custom_typings/escodegen.d.ts
@@ -3,8 +3,11 @@ declare module 'escodegen' {
   interface GenerateOpts {
     comment?: boolean;
     format?: {
-      indent?:
-          {style?: string; base?: number; adjustMultilineComment: boolean;}
+      indent?: {
+        style?: string;  //
+        base?: number;
+        adjustMultilineComment: boolean;
+      }
     }
   }
   export function generate(ast: Node, opts?: GenerateOpts): string;
--- a/custom_typings/escodegen.d.ts
+++ b/custom_typings/escodegen.d.ts
@@ -3,13 +3,9 @@ declare module 'escodegen' {
   interface GenerateOpts {
     comment?: boolean;
     format?: {
-      indent?: {
-        style?: string;
-        base?: number;
-        adjustMultilineComment: boolean;
-      }
+      indent?:
+          {style?: string; base?: number; adjustMultilineComment: boolean;}
     }
   }
-  export function generate(ast:Node, opts?: GenerateOpts):string;
-
+  export function generate(ast: Node, opts?: GenerateOpts): string;
 }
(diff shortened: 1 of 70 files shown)


FredKSchott:
personal opinion: Not familiar with clang-format, but I like the change overall. 1 thing I'm not a fan of: squashing all object literals into single lines. Is there a way to turn that off?

rictic:
You can hint that you'd like it not to do that by ending a line with `//`. We may also be able to give it a style that we like better.

I generally like the changes, but not all of clang-format's opinions are my own. The thing that I like about an automatic formatter is it gives us a uniform style while saving us from having to think about it.

src/analyzer.ts
```diff
@@ -81,7 +87,7 @@ export class Analyzer {
     if (this._documentDescriptors.has(url)) {
       return this._documentDescriptors.get(url);
     }
-    let promise = (async () => {
+    let promise = (async() => {
```
justinfagnani: this is unfortunate

src/analyzer.ts
```diff
@@ -1,11 +1,15 @@
 /**
  * @license
  * Copyright (c) 2015 The Polymer Project Authors. All rights reserved.
- * This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
- * The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
- * The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+ * This code may only be used under the BSD style license found at
```
justinfagnani: can we exclude @licence comments from formatting?

custom_typings/escodegen.d.ts
```diff
@@ -3,13 +3,9 @@ declare module 'escodegen' {
   interface GenerateOpts {
     comment?: boolean;
     format?: {
-      indent?: {
-        style?: string;
-        base?: number;
-        adjustMultilineComment: boolean;
-      }
+      indent?:
+          {style?: string; base?: number; adjustMultilineComment: boolean;}
```
justinfagnani: are there any controls over literals besides the `//` hint?

src/analyzer.ts
rictic: AFAICT, no: http://clang.llvm.org/docs/ClangFormatStyleOptions.html

src/analyzer.ts
rictic: yep.

custom_typings/escodegen.d.ts
rictic: Not afaict. I've reduced the wrapping some, but haven't found any tricks for object literals yet.

rictic:
Got verbal LGTM from Justin. Merging

PR #263, opened 2016-07-28
Preserve document ordering of entities

justinfagnani:


src/javascript/javascript-document.ts
```diff
@@ -45,4 +45,14 @@ export class JavaScriptDocument extends Document<Program, Visitor> {
       fallback: 'iteration',
     });
   }
+
+  forEachNode(callback: (node: Node) => void) {
+    traverse(<Node><any>this.ast, {
```
rictic: Why do you need to do the `<Node><any>` dance, a Program is a Node..

src/entity/find-entities.ts
```diff
@@ -86,5 +86,37 @@ export async function findEntities(
   }
 
   // Flatten the nested list
-  return Array.prototype.concat.apply([], nestedEntities);
+  let unorderedEntities = Array.prototype.concat.apply([], nestedEntities);
+  let orderedEntities = orderEntities(document, unorderedEntities);
```
rictic: WDYT about leaving out the `orderedEntities` local and just returning instead?

src/css/css-document.ts
```diff
@@ -22,4 +22,8 @@ export class CssDocument extends Document<Node, NodeVisitor> {
   visit(visitors: NodeVisitor[]) {
     throw new Error('Not implemented');
   }
+
+  forEachNode(callback: (node: Node) => void) {
```
rictic:
```javascript
/**
 * _Must_ call the callback with every node, and must do so in document order.
 */
```

src/entity/find-entities.ts
```diff
@@ -86,5 +86,37 @@ export async function findEntities(
...
+
+  return orderedEntities;
+}
+
+function orderEntities(
+  document: Document<any, any>,
+  unorderedEntities: Descriptor[]): Descriptor[] {
+
+  let entitiesByNode = new Map<any, Descriptor[]>();
+  let orderedEntities: Descriptor[] = [];
```
rictic: This type should be `Descriptor[][]` shouldn't it? Is there something wrong with the type checker, or something wrong with my brain?

rictic:
Looks reasonable to me. You mentioned adding more tests, and that seems wise to me too.

src/javascript/javascript-document.ts
justinfagnani: I forget, let me check

justinfagnani: removed

src/entity/find-entities.ts
justinfagnani: sure

src/css/css-document.ts
justinfagnani: added to comment

src/entity/find-entities.ts
justinfagnani: You're right. I don't know what's going on with the type checking here. I've tried several combinations of what should be invalid types and tsc isn't complaining.

justinfagnani:
I want to move on this before adding more test cases so we can unlock bundling...

PR #266, opened 2016-07-28
Initial typing of a serialized analysis

rictic:
* Stubbed out implementation of the `Context` object, which I've named `Analysis` (it's what you get from asking an analyzer to analyze!
* Wrote out and documented interfaces for the metadata format, which I've called `SerializedAnalysis`.
  * Wrote out an example for paper-button here: https://gist.github.com/rictic/b6fd4f19b6f265753d5d11cdb8a41940

### DIFF
--- a/src/analysis.ts
+++ b/src/analysis.ts
@@ -1,0 +1,13 @@
+import {DocumentDescriptor} from './ast/ast';
+import {SerializedAnalysis} from './serialized-analysis';
+
+export class Analysis {
+  private descriptors_: DocumentDescriptor[];
+  constructor(descriptors: DocumentDescriptor[]) {
+    this.descriptors_ = descriptors;
+  }
+
+  serialize(): SerializedAnalysis {
+    throw new Error('not implemented');
+  }
+}
--- a/src/analysis.ts
+++ b/src/analysis.ts
@@ -16,9 +16,9 @@ import {DocumentDescriptor} from './ast/ast';
 import {SerializedAnalysis} from './serialized-analysis';
 
 export class Analysis {
-  private descriptors_: DocumentDescriptor[];
+  private _descriptors: DocumentDescriptor[];
   constructor(descriptors: DocumentDescriptor[]) {
-    this.descriptors_ = descriptors;
+    this._descriptors = descriptors;
   }
 
   serialize(): SerializedAnalysis {
--- a/src/analysis.ts
+++ b/src/analysis.ts
@@ -13,7 +13,7 @@
  */
 
 import {DocumentDescriptor} from './ast/ast';
-import {SerializedAnalysis} from './serialized-analysis';
+import {AnalyzedPackage} from './serialized-analysis';
 
 export class Analysis {
   private _descriptors: DocumentDescriptor[];
@@ -21,7 +21,7 @@ export class Analysis {
     this._descriptors = descriptors;
   }
 
-  serialize(): SerializedAnalysis {
+  serialize(): AnalyzedPackage {
     throw new Error('not implemented');
   }
 }
--- a/src/analysis.ts
+++ b/src/analysis.ts
@@ -1,3 +1,17 @@
+/**
+ * @license
+ * Copyright (c) 2015 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at
+ * http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
 import {DocumentDescriptor} from './ast/ast';
 import {SerializedAnalysis} from './serialized-analysis';
 
--- a/src/analyzer.ts
+++ b/src/analyzer.ts
@@ -93,8 +93,7 @@ export class Analyzer {
   }
 
   async resolve(): Promise<Analysis> {
-    const descriptorPromises = Array.from(this._documentDescriptors.values());
-    return new Analysis(Array.from(await Promise.all(descriptorPromises)));
+    return new Analysis(await Promise.all(this._documentDescriptors.values()));
   }
 
   /**
--- a/src/analyzer.ts
+++ b/src/analyzer.ts
@@ -15,6 +15,7 @@
 import * as path from 'path';
 import * as urlLib from 'url';
 
+import {Analysis} from './analysis';
 import {Descriptor, DocumentDescriptor, ImportDescriptor, InlineDocumentDescriptor} from './ast/ast';
 import {CssParser} from './css/css-parser';
 import {EntityFinder} from './entity/entity-finder';
@@ -89,6 +90,11 @@ export class Analyzer {
     })();
     this._documentDescriptors.set(url, promise);
     return promise;
+  }
+
+  async resolve(): Promise<Analysis> {
+    const descriptorPromises = Array.from(this._documentDescriptors.values());
+    return new Analysis(Array.from(await Promise.all(descriptorPromises)));
   }
 
   /**
--- a/src/serialized-analysis.ts
+++ b/src/serialized-analysis.ts
@@ -1,0 +1,425 @@
+export interface SerializedAnalysis { packages: Package[]; }
+
+export interface Package {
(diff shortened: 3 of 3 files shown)


src/analysis.ts
```diff
@@ -0,0 +1,13 @@
+import {DocumentDescriptor} from './ast/ast';
+import {SerializedAnalysis} from './serialized-analysis';
+
+export class Analysis {
```
justinfagnani: I was thinking `AnalysisResult`

src/analysis.ts
```diff
@@ -0,0 +1,13 @@
+import {DocumentDescriptor} from './ast/ast';
```
justinfagnani: license

src/analysis.ts
```diff
@@ -0,0 +1,13 @@
...
+import {SerializedAnalysis} from './serialized-analysis';
+
+export class Analysis {
+  private descriptors_: DocumentDescriptor[];
```
justinfagnani: Use leading underscore

src/analysis.ts
```diff
@@ -0,0 +1,13 @@
...
+  constructor(descriptors: DocumentDescriptor[]) {
+    this.descriptors_ = descriptors;
+  }
+
+  serialize(): SerializedAnalysis {
```
justinfagnani: Ultimately, I think this should be a function that takes an Analysis, not a method.

src/analyzer.ts
```diff
@@ -91,6 +92,11 @@ export class Analyzer {
     return promise;
   }
 
+  async resolve(): Promise<Analysis> {
+    const descriptorPromises = Array.from(this._documentDescriptors.values());
+    return new Analysis(Array.from(await Promise.all(descriptorPromises)));
```
justinfagnani: `await Promise.all(...)` is already an `Array`

src/analyzer.ts
```diff
@@ -91,6 +92,11 @@ export class Analyzer {
     return promise;
   }
 
+  async resolve(): Promise<Analysis> {
+    const descriptorPromises = Array.from(this._documentDescriptors.values());
```
justinfagnani: `Promise.all()` takes an iterable, no need for `Array.from()`

src/serialized-analysis.ts
```diff
@@ -0,0 +1,425 @@
+export interface SerializedAnalysis { packages: Package[]; }
+
+export interface Package {
```
justinfagnani: I don't think we need `Package`. An `elements.json` file should already be in a package, next to a `bower.json` and/or `package.json`. No need to duplicate that info here.

src/serialized-analysis.ts
```diff
@@ -0,0 +1,425 @@
...
+   *
+   * e.g. `paper-input.html` or `app-toolbar/app-toolbar.html` (given that
+   * app-toolbar lives in the app-layout package).
+   */
+  path: string;
+
+  /** The tagname that the element registers itself as. e.g. `paper-input` */
+  tagname: string;
+
+  /** A markdown description for the element. */
```
justinfagnani: We need to consider what to do if we read the documentation from an HTML comment...

src/serialized-analysis.ts
```diff
@@ -0,0 +1,425 @@
...
+  description: string;
+
+  /**
+   * Paths, relative to `this.path` to demo pages for the element.
```
justinfagnani: Not sure they should be relative to `this.path`: what about `/elements/foo-element.html` and `demos/foo-element-demo.html`? I think for simplicity all paths should be relative to this file.

src/serialized-analysis.ts
```diff
@@ -0,0 +1,425 @@
...
+  attributes?: Attribute[];
+
+  /** The properties that this element has. */
+  properties?: Property[];
+
+  /** The events that this element fires. */
+  events?: Event[];
+
+  /** The shadow dom content slots that this element accepts. */
+  'slots': Slot[];
```
justinfagnani: remove quotes

src/serialized-analysis.ts
```diff
@@ -0,0 +1,425 @@
...
+
+  /** Information useful for styling the element and its children. */
+  styling: {
+
+    /** CSS Classes that produce built-in custom styling for the element. */
+    classes: {
```
justinfagnani: I'm not sure this should just be classes. An element could have a `:host([height=5])` rule. Maybe just selectors?

src/serialized-analysis.ts
```diff
@@ -0,0 +1,425 @@
...
+
+  /**
+   * The type that the attribute will be serialized/deserialized as.
+   *
+   * e.g. `string`, `number`, `boolean`, `RegExp`, `Array`, `Object`.
+   */
+  type?: string;
+
+  /** The default value of the attribute, if any. */
+  defaultValue?: string;
```
justinfagnani: We need some way to indicate if an attribute is associated with a property... TBD?

src/serialized-analysis.ts
```diff
@@ -0,0 +1,425 @@
...
+    // display: block or inline or whatever.
+  };
+
+  /**
+   * An extension point for framework-specific metadata, as well as any
+   * metadata not yet standardized here such as what polyfills are needed,
+   * behaviors and mixins used, the framework that the element was written in,
+   * tags/categories, links to specs that the element implements, etc.
+   */
+  metadata: Object;
```
justinfagnani: How about we define a super-interface with `metadata` and `location`?

src/serialized-analysis.ts
```diff
@@ -0,0 +1,425 @@
...
+  /**
+   * An extension point for framework-specific metadata, as well as any
+   * metadata not yet standardized here such as associated events and
+   * observers.
+   */
+  metadata?: Object;
+}
+
+export interface Event {
+  /** The name of the event. */
```
justinfagnani: We also need some way to associate a change event with a property or attribute...

src/analysis.ts
rictic: Done.

src/analysis.ts
rictic: Seems unnecessarily wordy. The result of the analyzer is naturally an analysis.

OTOH `breakfast food` is an actual phrase people use, so, I guess there's precedent.

src/analyzer.ts
rictic: Oh, neat!

src/analyzer.ts
rictic: I can't imagine what I was thinking when I wrote that.

src/serialized-analysis.ts
rictic: Most of the elements we maintain are documented in HTML comments, but they still use markdown. e.g. https://github.com/PolymerElements/paper-button/blob/master/paper-button.html

src/serialized-analysis.ts
rictic: `elements.json`, I like it.

`version` was requested from the catalog folks, which prompted the concept of a package.

I'm ok with not duplicating info from bower.json and package.json and dropping those fields, I kinda threw them in on a lark. Do we ever want to represent multiple packages' worth of elements in one metadata file though? I suspect that we do for the catalog use case, where it would be nice to have one file that contained all of the metadata. For an `elements.json` it would make sense to serialize just one package, sure.

src/serialized-analysis.ts
rictic: I'm not sure I understand, can you rephrase?

src/serialized-analysis.ts
rictic: Done. clang-format gets a little confused and doesn't do the best thing, but leaving the quotes in confuses the json schema generator I'm using even more.

(spoiler: I've got json schema generation working. Also vscode understands json schema, and it's pretty great)

src/serialized-analysis.ts
rictic: Added a comment.

Is that needed though? What's the use case there?

src/analysis.ts
rictic: Acknowledged, but why's that?

src/analysis.ts
rictic: Done

src/serialized-analysis.ts
rictic: Added TBD comment

src/serialized-analysis.ts
rictic: Done

src/serialized-analysis.ts
justinfagnani: rather that "this file" (by which I meant the elements.json file) we can say package directory

justinfagnani:
LGTM

PR #270, opened 2016-07-30
Port behavior-finder to new API

justinfagnani:


package.json
```diff
@@ -56,6 +56,7 @@
     "doctrine": "^0.7.0",
     "dom5": "^1.1.0",
     "escodegen": "^1.7.0",
+    "espree": "^3.1.7",
```
justinfagnani: oh... I can remove the direct dependency on acorn

src/ast/element-descriptor.ts
```diff
@@ -29,3 +43,33 @@ export interface ElementDescriptor extends Descriptor {
 
   abstract?: boolean;
 }
+
+/**
+ * The metadata for a single polymer element
+ */
+export class ElementDescriptor implements Descriptor {
```
rictic: I'd kinda prefer to have two different classes. One for elements, one for behaviors that extend a common class (inheriting most of their properties from same).

That way we could do:


```typescript
if (elementOrBehavior.type === 'element') {
  // type checker knows it's an element here
}
```

test/analyzer_test.js
```diff
@@ -170,4 +170,30 @@ suite('Analyzer', () => {
...
+          assert.equal(element1.behaviors.length, 2);
+          assert.equal(element1.behaviors[0], 'Behavior1');
+          assert.equal(element1.behaviors[1], 'Behavior2');
+          assert.equal(element1.is, 'test-seed');
+
+          assert.equal(element1.observers.length, 2);
+          assert.equal(element1.properties.length, 4);
+
+          // TODO(justinfagnani): fix events
+          // assert.equal(elements[0].events.length, 1);
```
rictic: consider moving this into a skipped test, just so we don't forget it

test/polymer/behavior-finder_test.js
```diff
@@ -0,0 +1,98 @@
...
+  });
+
+  test('Supports behaviors at local assignments', () => {
+    assert(behaviors.has('SimpleBehavior'));
+    assert.equal(behaviors.get('SimpleBehavior').properties[0].name, 'simple');
+  });
+
+  test('Supports behaviors with renamed paths', () => {
+    assert(behaviors.has('AwesomeBehavior'));
+    // console.log(analyzer.behaviorsByName['AwesomeBehavior'].properties);
```
rictic: unneeded

test/polymer/behavior-finder_test.js
```diff
@@ -0,0 +1,98 @@
...
+    document = parser.parse(file, '/static/js-behaviors.js');
+    let finder = new BehaviorFinder();
+    let visit = (visitor) => document.visit([visitor]);
+
+    return finder.findEntities(document, visit)
+      .then((entities) => {
+        behaviors = new Map();
+        behaviorsList = entities.filter((e) => e instanceof BehaviorDescriptor);
+        for (let behavior of behaviorsList) {
+          behaviors.set(behavior.is, behavior);
```
rictic: assert that behaviors doesn't already have an element at behavior.is

test/polymer/behavior-finder_test.js
```diff
@@ -0,0 +1,98 @@
...
+  test('Supports behaviors at local assignments', () => {
+    assert(behaviors.has('SimpleBehavior'));
+    assert.equal(behaviors.get('SimpleBehavior').properties[0].name, 'simple');
+  });
+
+  test('Supports behaviors with renamed paths', () => {
+    assert(behaviors.has('AwesomeBehavior'));
+    // console.log(analyzer.behaviorsByName['AwesomeBehavior'].properties);
+    var found = false;
+    behaviors.get('AwesomeBehavior').properties.forEach((prop) => {
```
rictic: Consider:

```typescript
assert(behaviors.get('AwesomeBehavior').properties.some(prop => prop.name === 'custom'));
```

test/polymer/behavior-finder_test.js
```diff
@@ -0,0 +1,98 @@
...
+        // console.log(prop);
+        defaultValue = prop.default;
+      }
+    });
+    assert.equal(defaultValue, 1);
+  });
+
+  test('Supports chained behaviors', function() {
+    assert(behaviors.has('CustomBehaviorList'));
+    assert.equal(behaviors.get('CustomBehaviorList').behaviors[0], 'SimpleBehavior');
```
rictic: IMO it's way easier to read:

```typescript
assert.deepEquals(behaviors.get('CustomBehaviorList').behaviors, ['SimpleBehavior', 'CustomNamedBehavior', 'Really.Really.Deep.Behavior']);
```

The error messages are better too (e.g. way easier to read if the results are in a different order)

src/ast/element-descriptor.ts
justinfagnani: I agree, but I'd rather stay as-is for a very short time, as I think there's a big refactor coming here next.

I want to move Polymer-specific things into the `polymer/` folder and pull out a plain element descriptor interface to live here. I also want to limit the number of references that the descriptors have into the AST... it shouldn't normally be necessary.

test/polymer/behavior-finder_test.js
justinfagnani: I'm not sure that's so needed in the test. We know the input and check for the existence of all the behaviors defined there.

package.json
justinfagnani: done

test/analyzer_test.js
justinfagnani: done

test/polymer/behavior-finder_test.js
justinfagnani: done

src/ast/element-descriptor.ts
rictic: Yeah that makes sense. For that we'll likely  want to move some of the code I've been writing as part of the serializer up into the element finder, to convert the polymer-specific stuff into general custom element metadata.

test/polymer/behavior-finder_test.js
rictic: fair enough

test/polymer/behavior-finder_test.js
justinfagnani: done

test/polymer/behavior-finder_test.js
justinfagnani: yeah... I was keeping the tests mostly the same, there wasn't one for all the behaviors existing. Good idea.

justinfagnani:
PTAL

rictic:
Needs to be resolved against the typescriptified tests changes

justinfagnani:
Rebased and `push -f`'ed

rictic:
LGTM

PR #275, opened 2016-08-04
Extend elements.json generation with jsdoc and source locations

rictic:
- extract description for element and properties from jsdoc
  - convert constructor names into type names for primitive types
    (e.g. in a polymer properties block `type: String` means the
    `string` type, not `String`
  - thread line and column for source location through to elements.json
  - properly correct source location for inner documents


src/analyzer.ts
```diff
@@ -104,10 +104,13 @@ export class Analyzer {
...
    */
-  private async _analyzeSource(type: string, contents: string, url: string):
-      Promise<DocumentDescriptor> {
+  private async _analyzeSource(
+      type: string, contents: string, url: string,
+      locationOffset?: LocationOffset): Promise<DocumentDescriptor> {
     let document = this.parse(type, contents, url);
-    return this._analyzeDocument(document);
+    const documentDescriptor = await this._analyzeDocument(document);
+    documentDescriptor.locationOffset = locationOffset;
```
justinfagnani: what about passing this as an optional parameter to `_analyzeDocument()`?

src/generate-elements.ts
```diff
@@ -86,14 +90,18 @@ function serializeElementDescriptor(
...
     events: propChangeEvents,
     metadata: {},
+    sourceLocation:
+        correctSourceLocation(elementDescriptor.sourceLocation, locationOffset)
   };
 }
 
-function serializePropertyDescriptor(p: PropertyDescriptor): Property {
+function serializePropertyDescriptor(
+    p: PropertyDescriptor, locationOffset?: LocationOffset): Property {
```
justinfagnani: naming: `p` -> `property`?

src/generate-elements.ts
```diff
@@ -266,3 +292,18 @@ function propertyToAttributeName(propertyName: string): string|null {
...
       /([A-Z])/g, (_: string, c1: string) => `-${c1.toLowerCase()}`);
 }
+
+function correctSourceLocation(
+    sourceLocation: SourceLocation,
+    locationOffset?: LocationOffset): SourceLocation|undefined {
+  if (!sourceLocation) {
+    return undefined;
+  }
+  locationOffset = locationOffset || {line: 0, col: 0};
```
justinfagnani: How about:

```javascript
if (!locationOffset) return sourceLocation;
return sourceLocation && {
  line: ...,
  column: ...,
};
```

src/generate-elements.ts
```diff
@@ -169,7 +180,8 @@ class ElementGatherer implements AnalysisVisitor {
       }
     } else {
       this._elementPaths.set(element, pathToElement);
-      const elem = serializeElementDescriptor(element, pathToElement);
+      const elem =
```
justinfagnani: having both 'element' and 'elem' identifiers in the same scope could be confusing? How are they different? Can that be in the name?

src/html/html-script-finder.ts
```diff
@@ -44,13 +45,34 @@ export class HtmlScriptFinder implements HtmlEntityFinder {
           entities.push(
               new ImportDescriptor<ASTNode>('html-script', importUrl, node));
         } else {
+          let locationOffset: LocationOffset|undefined = undefined;
```
justinfagnani: This function is getting long, IMO. Can you pull this logic out?

justinfagnani: `= undefined` isn't necessary, it's the default value.

src/html/html-script-finder.ts
```diff
@@ -44,13 +45,34 @@ export class HtmlScriptFinder implements HtmlEntityFinder {
...
         }
       }
     });
 
     return entities;
   }
 }
+
+function isLocationInfo(loc: LocationInfo|
+                        ElementLocationInfo): loc is LocationInfo {
```
justinfagnani: weird alignment... clang does this?

justinfagnani:
It looks like most changes are covered by high-level tests, but I would love to see some more coverage by simple unit tests for things like correctSourceLocation, objectKeyToString, the locationOffset in HtmlScriptFinder, etc. They're border-line boring tests, but should be simple and easier to reason about and fix than looking at offsets from a test file.

src/analyzer.ts
rictic: Done

src/generate-elements.ts
rictic: Done

src/generate-elements.ts
rictic: Nice, done.

src/generate-elements.ts
rictic: Good point. Expanded `element` to `elementDescriptor` and `elem` to `element` here to clarify that ambiguity.

src/html/html-script-finder.ts
rictic: Extracted as `getLocationOffsetOfStartOfTextContent`, and removed unnecessary assignment.

src/html/html-script-finder.ts
rictic: Agreed that it's weird, but yep.

rictic:
All comments addressed, PTAL

src/html/html-script-finder.ts
```diff
@@ -75,4 +60,24 @@ export class HtmlScriptFinder implements HtmlEntityFinder {
 function isLocationInfo(loc: LocationInfo|
                         ElementLocationInfo): loc is LocationInfo {
   return 'line' in loc;
+}
+
+function getLocationOffsetOfStartOfTextContent(node: ASTNode) {
```
justinfagnani: you know... this would be a decent dom5 addition

justinfagnani:
LGTM

PR #282, opened 2016-08-07
Resolve behaviors and fold their properties into elements.json

rictic:
Also:

* also moves SourceLocation correction for inline documents into the Analyzer and entirely out of the resolution phase
* correctly points to the right file when a property isn't defined in the same file as the element
* separate out the concerns of statically analyzing simple expressions and getting names of identifiers

### DIFF
--- a/src/analysis.ts
+++ b/src/analysis.ts
@@ -17,7 +17,7 @@ import * as jsonschema from 'jsonschema';
 import * as path from 'path';
 import * as util from 'util';
 
-import {Descriptor, DocumentDescriptor, ElementDescriptor, ImportDescriptor, InlineDocumentDescriptor, LocationOffset, PropertyDescriptor} from './ast/ast';
+import {Descriptor, DocumentDescriptor, ElementDescriptor, ImportDescriptor, InlineDocumentDescriptor, PropertyDescriptor} from './ast/ast';
 import {Elements} from './elements-format';
 import {JsonDocument} from './json/json-document';
 import {trimLeft} from './utils';
@@ -43,7 +43,6 @@ export class Analysis {
   private _elementsByTagName = new Map<string, ElementDescriptor>();
   private _elementsByPackageDir = new Map<string, ElementDescriptor[]>();
   elementPaths = new Map<ElementDescriptor, string>();
-  locationOffsets = new Map<ElementDescriptor, LocationOffset>();
 
   constructor(descriptors: DocumentDescriptor[]) {
     this._descriptors = descriptors;
@@ -78,7 +77,6 @@ export class Analysis {
                            .substring(longestMatchingPackageDir.length),
                        '/'));
     }
-    this.locationOffsets = elementsGatherer.locationOffsets;
     this.elementPaths = elementsGatherer.elementPaths;
   }
 
@@ -127,15 +125,12 @@ class PackageGatherer implements AnalysisVisitor {
 
 class ElementGatherer implements AnalysisVisitor {
   elementDescriptors: ElementDescriptor[] = [];
-  locationOffsets = new Map<ElementDescriptor, LocationOffset>();
   elementPaths = new Map<ElementDescriptor, string>();
   visitElement(elementDescriptor: ElementDescriptor, path: Descriptor[]): void {
     let pathToElement: string|null = null;
-    let locationOffset: LocationOffset|undefined;
     for (const descriptor of path) {
       if (descriptor instanceof DocumentDescriptor) {
         pathToElement = descriptor.document.url;
-        locationOffset = descriptor.locationOffset;
       }
     }
     if (!pathToElement) {
@@ -152,7 +147,6 @@ class ElementGatherer implements AnalysisVisitor {
     }
 
     this.elementPaths.set(elementDescriptor, pathToElement);
-    this.locationOffsets.set(elementDescriptor, locationOffset);
     this.elementDescriptors.push(elementDescriptor);
   }
 }
--- a/src/analysis.ts
+++ b/src/analysis.ts
@@ -20,7 +20,6 @@ import * as util from 'util';
 import {BehaviorDescriptor, Descriptor, DocumentDescriptor, ElementDescriptor, ImportDescriptor, InlineDocumentDescriptor, PropertyDescriptor} from './ast/ast';
 import {Elements} from './elements-format';
 import {JsonDocument} from './json/json-document';
-import {trimLeft} from './utils';
 
 const validator = new jsonschema.Validator();
 const schema = JSON.parse(
@@ -62,7 +61,7 @@ export class Analysis {
       let elementPath = elementsGatherer.elementPaths.get(element);
       const matchingPackageDirs =
           Array.from(packageGatherer.packageDirs)
-              .filter(dir => trimLeft(elementPath, '/').startsWith(dir));
+              .filter(dir => elementPath.startsWith(dir));
       const longestMatchingPackageDir =
           matchingPackageDirs.sort((a, b) => b.length - a.length)[0] || '';
 
@@ -71,12 +70,6 @@ export class Analysis {
       elementsInPackage.push(element);
       this._elementsByPackageDir.set(
           longestMatchingPackageDir, elementsInPackage);
-      // We want element paths to be relative to the package directory.
-      elementsGatherer.elementPaths.set(
-          element, trimLeft(
-                       trimLeft(elementPath, '/')
-                           .substring(longestMatchingPackageDir.length),
-                       '/'));
     }
 
     for (const behavior of elementsGatherer.behaviorDescriptors) {
@@ -130,7 +123,7 @@ class PackageGatherer implements AnalysisVisitor {
          dd.document.url.endsWith('bower.json'))) {
       const dirname = path.dirname(dd.document.url);
       if (!this.packageDirs.has(dirname)) {
-        this.packageDirs.add(trimLeft(dirname, '/'));
+        this.packageDirs.add(dirname);
       }
     }
   }
--- a/src/analysis.ts
+++ b/src/analysis.ts
@@ -17,7 +17,7 @@ import * as jsonschema from 'jsonschema';
 import * as path from 'path';
 import * as util from 'util';
 
-import {Descriptor, DocumentDescriptor, ElementDescriptor, ImportDescriptor, InlineDocumentDescriptor, PropertyDescriptor} from './ast/ast';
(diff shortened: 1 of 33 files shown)


src/analysis.ts
```diff
@@ -36,10 +39,55 @@ export class ValidationError extends Error {
...
+    for (const element of elements) {
+      if (element.is) {
+        this._elementsByTagName.set(element.is);
+      }
+      let elementPath = elementsGatherer.elementPaths.get(element);
+      const matchingPackageDirs =
+          Array.from(packageGatherer.packageDirs)
+              .filter(dir => trimLeft(elementPath, '/').startsWith(dir));
+      const longestMatchingPackageDir =
+          matchingPackageDirs.sort((a, b) => b.length - a.length)[0] || '';
```
justinfagnani: sorting the whole list is an inefficient way to get the min or max. `reduce()`?

justinfagnani: actually, matching then getting the longest could be combined into one loop and it might be easier to follow

src/analysis.ts
```diff
@@ -36,10 +39,55 @@ export class ValidationError extends Error {
...
+      packageGatherer, elementsGatherer
+    ]);
+
+    const elements = elementsGatherer.elementDescriptors;
+
+    for (const element of elements) {
+      if (element.is) {
+        this._elementsByTagName.set(element.is);
+      }
+      let elementPath = elementsGatherer.elementPaths.get(element);
```
justinfagnani: can you add some comments about what this look is doing? I'm finding the code not super self-documenting

src/analysis.ts
```diff
@@ -36,10 +39,55 @@ export class ValidationError extends Error {
 
 
 export class Analysis {
-  descriptors: DocumentDescriptor[];
+  private _descriptors: DocumentDescriptor[];
+  private _elementsByTagName = new Map<string, ElementDescriptor>();
+  private _elementsByPackageDir = new Map<string, ElementDescriptor[]>();
+  elementPaths = new Map<ElementDescriptor, string>();
+  locationOffsets = new Map<ElementDescriptor, LocationOffset>();
```
justinfagnani: should this type of location info just be folded into the descriptors?

src/analysis.ts
```diff
@@ -36,10 +39,55 @@ export class ValidationError extends Error {
...
+          matchingPackageDirs.sort((a, b) => b.length - a.length)[0] || '';
+
+      const elementsInPackage =
+          this._elementsByPackageDir.get(longestMatchingPackageDir) || [];
+      elementsInPackage.push(element);
+      this._elementsByPackageDir.set(
+          longestMatchingPackageDir, elementsInPackage);
+      // We want element paths to be relative to the package directory.
+      elementsGatherer.elementPaths.set(
+          element, trimLeft(
```
justinfagnani: why trim operations on paths? is there a `path` operation that does what you want?

src/analysis.ts
```diff
@@ -61,3 +109,138 @@ export class Analysis {
     }
   }
 }
+
+
+class PackageGatherer implements AnalysisVisitor {
+  packageDirs = new Set<string>();
+  visitDocumentDescriptor(dd: DocumentDescriptor): void {
+    if (dd.document instanceof JsonDocument &&
+        (dd.document.url.endsWith('package.json') ||
```
justinfagnani: use `path.basename()`

src/analysis.ts
```diff
@@ -61,3 +109,138 @@ export class Analysis {
...
+         dd.document.url.endsWith('bower.json'))) {
+      const dirname = path.dirname(dd.document.url);
+      if (!this.packageDirs.has(dirname)) {
+        this.packageDirs.add(trimLeft(dirname, '/'));
```
justinfagnani: use `path.resolve()`?

src/analysis.ts
```diff
@@ -61,3 +109,138 @@ export class Analysis {
...
+      }
+    }
+  }
+}
+
+class ElementGatherer implements AnalysisVisitor {
+  elementDescriptors: ElementDescriptor[] = [];
+  locationOffsets = new Map<ElementDescriptor, LocationOffset>();
+  elementPaths = new Map<ElementDescriptor, string>();
+  visitElement(elementDescriptor: ElementDescriptor, path: Descriptor[]): void {
```
justinfagnani: style nit: newline between fields and methods

justinfagnani: `path` is shadowing the module here. Consider renaming the module to something like `pathLib`

src/analysis.ts
```diff
@@ -61,3 +109,138 @@ export class Analysis {
...
+    let pathToElement: string|null = null;
+    let locationOffset: LocationOffset|undefined;
+    for (const descriptor of path) {
```
justinfagnani: what's this loop doing? comment?

src/analysis.ts
```diff
@@ -61,3 +109,138 @@ export class Analysis {
...
+         dd.document.url.endsWith('bower.json'))) {
+      const dirname = path.dirname(dd.document.url);
+      if (!this.packageDirs.has(dirname)) {
+        this.packageDirs.add(trimLeft(dirname, '/'));
+      }
+    }
+  }
+}
+
+class ElementGatherer implements AnalysisVisitor {
```
justinfagnani: comment?

src/analysis.ts
```diff
@@ -61,3 +109,138 @@ export class Analysis {
...
+      return;
+    }
+
+    this.elementPaths.set(elementDescriptor, pathToElement);
+    this.locationOffsets.set(elementDescriptor, locationOffset);
+    this.elementDescriptors.push(elementDescriptor);
+  }
+}
+
+abstract class AnalysisVisitor {
```
justinfagnani: should this be an interface?

src/analysis.ts
```diff
@@ -61,3 +109,138 @@ export class Analysis {
...
+  visitDocumentDescriptor?(dd: DocumentDescriptor, path: Descriptor[]): void;
+  visitInlineDocumentDescriptor?
+      (dd: InlineDocumentDescriptor<any>, path: Descriptor[]): void;
+  visitElement?(element: ElementDescriptor, path: Descriptor[]): void;
+  visitImportDescriptor?
+      (importDesc: ImportDescriptor<any>, path: Descriptor[]): void;
+  done?(): void;
+}
+
+class AnalysisWalker {
```
justinfagnani: comment? Based on name alone I wonder what this difference is between this and Analysis Visitor

src/analysis.ts
```diff
@@ -61,3 +109,138 @@ export class Analysis {
...
+  private _documents: DocumentDescriptor[];
+  private _path: Descriptor[] = [];
+  constructor(descriptors: DocumentDescriptor[]) {
```
justinfagnani: newline before constructor

src/analysis.ts
justinfagnani: also, should it be exported?

src/analysis.ts
```diff
@@ -36,10 +38,60 @@ export class ValidationError extends Error {
...
+
+  getElement(tag: string): ElementDescriptor|undefined {
+    return this._elementsByTagName.get(tag);
+  }
+
+  getElementsForPackage(dirName: string): ElementDescriptor[]|undefined {
+    return this._elementsByPackageDir.get(dirName);
+  }
+
+  getBehavior(identifier: string): BehaviorDescriptor|undefined {
```
justinfagnani: since behaviors can be referenced by name, or an expression, make sure you clarify what this method expects

src/analysis.ts
rictic: Yes, that was the goal even as I wrote this, I just needed to control the complexity..

src/analysis.ts
rictic: Yes, path.resolve, eventually

src/analysis.ts
rictic: Switched to a O(n) max function

src/analysis.ts
rictic: `Index the elements that we found by their tag names and package names.` wdyt?

src/analysis.ts
rictic: done

src/analysis.ts
rictic: Removed string hackery

src/analysis.ts
rictic: changed the local to `elementPath` instead, added newline

src/analysis.ts
rictic: Commented and clarified

src/analysis.ts
rictic: Done.

src/analysis.ts
rictic: I wanted to make it easy to add functionality to the base class if needed. Once it's baked I'll export it.

src/analysis.ts
rictic: Added comment

src/analysis.ts
rictic: Done

src/analysis.ts
rictic: clarified

rictic:
All comments addressed.

(also just leaving this here as a preview of what's to come: https://github.com/Polymer/polymer-analyzer/compare/master...better-completions )

src/javascript/ast-value.ts
```diff
@@ -83,66 +86,57 @@ function returnStatementToValue(ret: estree.ReturnStatement): LiteralValue {
...
-      continue;
+function objectExpressionToValue(obj: estree.ObjectExpression): LiteralValue {
+  let value: LiteralObj = {};
+  for (const prop of obj.properties) {
+    if (prop.key.type !== 'Literal') {
+      return undefined;
     }
-    if (i !== 0) {
-      value += ', ';
+    const k = literalToValue(prop.key).toString();
```
FredKSchott: it would be nice to name these variables

FredKSchott:
Unrelated to this PR: Why did we move `test/` into `src/`?

FredKSchott:
@rictic LGTM!

rictic:
We moved `test/` into `src/` to have a clean way of compiling our tests. I don't love it, but it was the best solution of ~3 things I tried.

justinfagnani:
LGTM as well

src/javascript/ast-value.ts
rictic: Done!

rictic:
Rebased and merging in!

PR #283, opened 2016-08-12
The mega redesign PR

rictic:
After this the public API of the project should be more or less set.

The main known issues:
* things aren't in their correct files. There was so much upheaval in this PR, and github's move detection isn't the best. IMO we should do a followup PR to get things into more reasonable files. (e.g. the only reference to the term `descriptor` in the codebase now is in filenames).
* documentation is not up to snuff. I'll be sending follow on commits into this PR to add documentation. yelling about the worst spots appreciated!
* three tests that are currently set to skip:
  * extracting polymer elements as es6 classes is mostly disabled. the code was over-sensitive and started firing for vanilla elements. we need to rewrite that anyways, as it's not robust
  * two tests in the elements.json generation, because Analysis handled package directories and the new Document-centric api doesn't, so that needs some thinking
* After a lot of floundering I completely stopped trying to optimize Document and just focused on getting something that's correct. At the moment every resolved feature lookup is O(N) on the total number of features, and the constant isn't very small. Recursive tree walk with lots of set construction and for-of iterable walking. Without telemetry though it's hard to say whether that's a problem.


src/analyzer.ts
```diff
@@ -54,24 +52,34 @@ export interface Options {
...
+      ]
     ],
-    ['js', [new PolymerElementFinder(), new BehaviorFinder()]],
   ]);
 
   private _loader: UrlLoader;
-  private _resolver: UrlResolver;
-  private _documents = new Map<string, Promise<Document<any, any>>>();
-  private _documentDescriptors = new Map<string, Promise<DocumentDescriptor>>();
+  private _resolver: UrlResolver|undefined;
```
justinfagnani: is `o: T|undefined` any different from `o?: T`? I thought not. Which should we prefer?

src/analyzer.ts
```diff
@@ -82,102 +90,117 @@ export class Analyzer {
 
   /**
    * Loads, parses and analyzes a document and its transitive dependencies.
-   *
-   * @param {string} url the location of the file to analyze
-   * @return {Promise<DocumentDescriptor>}
    */
-  async analyze(url: string): Promise<DocumentDescriptor> {
-    return this._analyzeResolved(this._resolveUrl(url));
+  async analyze(url: string): Promise<Document|null> {
```
justinfagnani: when does this return null?

src/analyzer.ts
```diff
@@ -82,102 +90,117 @@ export class Analyzer {
...
-    if (isExternalUrl) {
-      return null;
-    }
-    let cachedResult = this._documentDescriptors.get(resolvedUrl);
+  /**
+   * Like `analyze`, but for use cases like editors where the caller believes
+   * the file has changed since the last time it was analyzed, and you may
+   * even have unsaved in-memory contents to use for the given file.
+   */
+  async analyzeChangedFile(url: string, contents?: string):
```
justinfagnani: are there any assumptions/constraints relevant to calling this method? I'm think things like, the file should have been previously loaded by `analyze()` already. I wonder if we can reason about this more simply if just add a `contents` parameter to `analyze()` and merge the two.

src/analyzer.ts
```diff
@@ -82,102 +90,117 @@ export class Analyzer {
...
       return this._analyzeDocument(document);
     })();
-    this._documentDescriptors.set(resolvedUrl, promise);
+    this._scannedDocuments.set(resolvedUrl, promise);
     return promise;
   }
 
-  async resolve(): Promise<Analysis> {
-    return new Analysis(await Promise.all(this._documentDescriptors.values()));
+  private async _resolve(url: string): Promise<ScannedDocument|null> {
```
justinfagnani: also wondering when this is null. We should decide/comment what's an error case vs null.

justinfagnani: can this method be removed?

src/analyzer.ts
justinfagnani: looks like `makeRootDocument` never returns null. Changing the return type

src/analyzer.ts
justinfagnani: removed and added a `contents` parameter to `analyze()` that triggers cache clearing.

src/ast/document-descriptor.ts
```diff
@@ -38,3 +46,206 @@ export class DocumentDescriptor implements Descriptor {
...
+  kinds: Iterable<string>;
+  identifiers: Set<string>;
+
+  private _rootDocument: Document;
+  // Should be able to emulate this once every Feature has a SourceLocation by
+  // just filtering out by feature.sourceLocation.file === this.url;
+  private _localFeatures = new Set<Feature>();
+
+  static makeRootDocument(scannedDocument: ScannedDocument): Document {
+    const result = new Document(scannedDocument);
```
justinfagnani: This needs some comments. It's not immediately clear why a document is added as a feature of itself.

src/ast/document-descriptor.ts
```diff
@@ -38,3 +46,206 @@ export class DocumentDescriptor implements Descriptor {
...
+
+export class Document implements Feature {
+  url: string;
+  parsedDocument: ParsedDocument<any, any>;
+  isInline: boolean;
+
+  kinds: Iterable<string>;
+  identifiers: Set<string>;
+
+  private _rootDocument: Document;
```
justinfagnani: what is a rootDocument?

src/ast/document-descriptor.ts
```diff
@@ -38,3 +46,206 @@ export class DocumentDescriptor implements Descriptor {
     return this.document.url;
   }
 }
+
+export interface Feature {
+  kinds: Iterable<string>;
```
justinfagnani: This might be better as a Set so that it's clear that there's uniqueness and so that it has a `has` method.

src/ast/document-descriptor.ts
rictic: You need to know the document that we began the analysis with in order to resolve globals that were declared before the current document. Ideally every document declares its dependencies but nothing forces this.

src/ast/document-descriptor.ts
rictic: Not at computer to add comment but this is to resolve import cycles that include the root document.

src/ast/document-descriptor.ts
rictic: +1

src/analyzer.ts
```diff
@@ -82,102 +90,117 @@ export class Analyzer {
...
-      maybeLocationOffset?: LocationOffset): Promise<DocumentDescriptor> {
+      document: ParsedDocument<any, any>, maybeLocationOffset?: LocationOffset,
+      maybeAttachedComment?: string): Promise<ScannedDocument> {
     const locationOffset =
         maybeLocationOffset || {line: 0, col: 0, filename: document.url};
-    let entities = await this.getEntities(document);
+    let entities = await this._getEntities(document);
     for (const entity of entities) {
-      if (entity instanceof ElementDescriptor) {
+      if (entity instanceof ScannedElement) {
```
justinfagnani: why do only ScannedElements get locationOffsets?

src/ast/element-descriptor.ts
```diff
@@ -12,72 +12,36 @@
...
-  scriptElement?: dom5.Node;
-
-  abstract?: boolean;
-  sourceLocation?: SourceLocation;
-
-  constructor(options: Options) {
-    Object.assign(this, options);
-  }
 
   applyLocationOffset(locationOffset?: LocationOffset) {
```
justinfagnani: can we refactor this so that `sourceLocation` is set to the correct value at scan time?

src/analyzer.ts
rictic: Agree. I'd like every scanned feature to require a source location, and to handle location offsets

src/ast/document-descriptor.ts
```diff
@@ -38,3 +46,206 @@ export class DocumentDescriptor implements Descriptor {
...
+  getByKind(kind: string): Set<Feature> {
+    return this._getByKind(kind, new Set());
+  }
+
+  getById(kind: 'element', tagName: string): Set<Element>;
+  getById(kind: 'polymer-element', tagName: string): Set<Document>;
+  getById(kind: 'behavior', className: string): Set<Behavior>;
+  getById(kind: 'dom-module', idAttr: string): Set<DomModule>;
+  getById(kind: 'document', url: string): Set<Document>;
+  getById(kind: string, identifier: string): Set<Feature>;
```
justinfagnani: what's an identifier?

src/ast/document-descriptor.ts
```diff
@@ -38,3 +46,206 @@ export class DocumentDescriptor implements Descriptor {
...
+
+  private _isResolved = false;
+  private _resolve(base: ScannedDocument) {
+    if (this._isResolved) {
+      return;
+    }
+    this._isResolved = true;
+    const dependenciesByUrl = new Map(
+        base.dependencies.map((sd) => <[string, ScannedDocument]>[sd.url, sd]));
+    let i = 0;  // DELETE ME
```
justinfagnani: delete it :)

src/ast/document-descriptor.ts
justinfagnani: what happens when we have multiple root documents? A single dependency can be referenced from multiple entry points. Imagine we add a version of `analyze()` that takes an array for convenience, and we pass the entry points: `analyzer.analyze(['a.html', 'b.html'])` and `c.html` is referenced somewhere from `a` and `b`. What's the root of `c.html`?

src/ast/document-descriptor.ts
```diff
@@ -38,3 +46,206 @@ export class DocumentDescriptor implements Descriptor {
...
+  }
+
+  private constructor(base: ScannedDocument, rootDocument?: Document) {
+    if (rootDocument == null) {
+      this._rootDocument = this;
+    } else {
+      if (!base.isInline) {
+        const existingInstance = rootDocument.getOnlyAtId('document', base.url);
+        if (existingInstance) {
+          return existingInstance;
```
justinfagnani: I'd like to avoid returning interesting objects from constructors unless there's a really compelling reason to do so. We already have a factory method for `makeRootDocument`, we should make this a factory too.

src/ast/document-descriptor.ts
justinfagnani: Thinking about this some more, I think we're going to need to either not support analyzing undeclared dependencies, or implementing a more sophisticated dependency tracking where we warn if every import of a document doesn't satisfy its undeclared dependencies before importing it.

I'm inclined to remove rootDocument for now and get undeclared dependencies working later.

rictic: I think it's better to match semantics first and warn second. It is very common to depend on implicit deps.

src/analyzer.ts
FredKSchott: +1 for `o?: T`

FredKSchott:
rebase needed to resolve conflicts :(

src/ast/document-descriptor.ts
FredKSchott: +1

src/ast/document-descriptor.ts
```diff
@@ -118,7 +118,7 @@ export class Document implements Feature {
     for (const scannedFeature of base.entities) {
       i += 1;
       if (scannedFeature instanceof ScannedImport) {
-        const scannedDoc = dependenciesByUrl.get(scannedFeature.url)!;
```
FredKSchott: I've never seen this before, is this valid TypeScript syntax? what is this trailing exclamation mark doing?

FredKSchott: Ah, just realized your comment above (https://github.com/Polymer/polymer-analyzer/commit/c3b133992aeccb7077a10c58fc206aa1fceb2645) answers this. TIL!

FredKSchott:
LGTM with the new `analyzeRoot` interface

justinfagnani:
Ok, I think this is in ok shape to merge, acknowledging that there are still some big changes to make regarding roots and file updates. Merge pending...

rictic:
\o/

PR #290, opened 2016-08-18
Source ranges are better than locations

rictic:
* Also gives the responsibility of producing SourceRanges from Nodes to the Document

Commits are pretty logical, but the end point also looks pretty skimmable.

### DIFF
--- a/src/ast/element-descriptor.ts
+++ b/src/ast/element-descriptor.ts
@@ -15,16 +15,14 @@
 import * as estree from 'estree';
 
 import {SourceRange} from '../ast/ast';
-import {SourceLocation} from '../elements-format';
 import * as jsdoc from '../javascript/jsdoc';
 
-import {Document, Event, Feature, LocationOffset, Property, Resolvable, ScannedEvent, ScannedFeature, ScannedProperty, correctSourceLocation, correctSourceRange} from './ast';
+import {Document, Event, Feature, LocationOffset, Property, Resolvable, ScannedEvent, ScannedFeature, ScannedProperty, correctSourceRange} from './ast';
 
 export {Visitor} from '../javascript/estree-visitor';
 
 export interface ScannedAttribute {
   name: string;
-  sourceLocation: SourceLocation;
   sourceRange: SourceRange;
   description?: string;
   type?: string;
@@ -40,7 +38,6 @@ export class ScannedElement implements ScannedFeature, Resolvable {
   description = '';
   demos: {desc?: string; path: string}[] = [];
   events: ScannedEvent[] = [];
-  sourceLocation: SourceLocation;
   node: estree.Node;
   sourceRange: SourceRange;
 
@@ -50,17 +47,11 @@ export class ScannedElement implements ScannedFeature, Resolvable {
     if (!locationOffset) {
       return;
     }
-    this.sourceLocation =
-        correctSourceLocation(this.sourceLocation, locationOffset);
     this.sourceRange = correctSourceRange(this.sourceRange, locationOffset);
     for (const prop of this.properties) {
-      prop.sourceLocation =
-          correctSourceLocation(prop.sourceLocation, locationOffset);
       prop.sourceRange = correctSourceRange(prop.sourceRange, locationOffset);
     }
     for (const attribute of this.attributes) {
-      attribute.sourceLocation =
-          correctSourceLocation(attribute.sourceLocation, locationOffset);
       attribute.sourceRange =
           correctSourceRange(attribute.sourceRange, locationOffset);
     }
@@ -94,7 +85,6 @@ export class Element implements Feature {
   demos: {desc?: string; path: string}[] = [];
   events: Event[] = [];
   sourceRange: SourceRange;
-  sourceLocation: SourceLocation;
   jsdoc?: jsdoc.Annotation;
   kinds: Set<string> = new Set(['element']);
   get identifiers(): Set<string> {
--- a/src/ast/element-descriptor.ts
+++ b/src/ast/element-descriptor.ts
@@ -14,16 +14,18 @@
 
 import * as estree from 'estree';
 
+import {SourceRange} from '../ast/ast';
 import {SourceLocation} from '../elements-format';
 import * as jsdoc from '../javascript/jsdoc';
 
-import {Document, Event, Feature, LocationOffset, Property, Resolvable, ScannedEvent, ScannedFeature, ScannedProperty, correctSourceLocation} from './ast';
+import {Document, Event, Feature, LocationOffset, Property, Resolvable, ScannedEvent, ScannedFeature, ScannedProperty, correctSourceLocation, correctSourceRange} from './ast';
 
 export {Visitor} from '../javascript/estree-visitor';
 
 export interface ScannedAttribute {
   name: string;
   sourceLocation: SourceLocation;
+  sourceRange: SourceRange;
   description?: string;
   type?: string;
 }
@@ -40,6 +42,7 @@ export class ScannedElement implements ScannedFeature, Resolvable {
   events: ScannedEvent[] = [];
   sourceLocation: SourceLocation;
   node: estree.Node;
+  sourceRange: SourceRange;
 
   jsdoc?: jsdoc.Annotation;
 
@@ -49,13 +52,20 @@ export class ScannedElement implements ScannedFeature, Resolvable {
     }
     this.sourceLocation =
         correctSourceLocation(this.sourceLocation, locationOffset);
+    this.sourceRange = correctSourceRange(this.sourceRange, locationOffset);
     for (const prop of this.properties) {
       prop.sourceLocation =
           correctSourceLocation(prop.sourceLocation, locationOffset);
+      prop.sourceRange = correctSourceRange(prop.sourceRange, locationOffset);
     }
     for (const attribute of this.attributes) {
       attribute.sourceLocation =
           correctSourceLocation(attribute.sourceLocation, locationOffset);
+      attribute.sourceRange =
+          correctSourceRange(attribute.sourceRange, locationOffset);
+    }
+    for (const event of this.events) {
+      event.sourceRange = correctSourceRange(event.sourceRange, locationOffset);
     }
   }
 
@@ -83,6 +93,7 @@ export class Element implements Feature {
   description = '';
   demos: {desc?: string; path: string}[] = [];
   events: Event[] = [];
+  sourceRange: SourceRange;
   sourceLocation: SourceLocation;
   jsdoc?: jsdoc.Annotation;
   kinds: Set<string> = new Set(['element']);
--- a/src/ast/inline-document-descriptor.ts
+++ b/src/ast/inline-document-descriptor.ts
@@ -13,11 +13,10 @@
  */
 
 import * as dom5 from 'dom5';
+import * as parse5 from 'parse5';
 import {ASTNode} from 'parse5';
-import {ElementLocationInfo, LocationInfo} from 'parse5';
 import * as util from 'util';
 
-import {SourceLocation} from '../elements-format';
 import * as jsdoc from '../javascript/jsdoc';
 
 import {ScannedFeature} from './descriptor';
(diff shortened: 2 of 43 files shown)


rictic:
FYI I've got a followup PR that removes `node` on ScannedFeatures, ordering instead by the `sourceRange.start`. Seems simpler, and should be faster and reduce memory usage.

FredKSchott:
+1 for adding short jsdoc-style descriptions to some of these functions like `applyLocationOffset` and `correctSourceRange`. With `correctSourceRange` specifically I was having trouble understanding what exactly "correction" meant in the context of a source range.

src/ast/inline-document-descriptor.ts
```diff
@@ -57,33 +57,39 @@ export class InlineParsedDocument<N> implements ScannedFeature {
...
 }
 
-export function correctSourceLocation(
-    sourceLocation: SourceLocation,
-    locationOffset?: LocationOffset): SourceLocation|undefined {
-  if (!locationOffset || !sourceLocation) {
-    return sourceLocation;
+export function correctSourceRange(
+    sourceRange: SourceRange, locationOffset?: LocationOffset): SourceRange|
+    undefined {
```
FredKSchott: did we decide to change to `SourceRange?`?

src/ast/inline-document-descriptor.ts
```diff
@@ -57,33 +57,39 @@ export class InlineParsedDocument<N> implements ScannedFeature {
...
   }
 }
 
-export function correctSourceLocation(
-    sourceLocation: SourceLocation,
-    locationOffset?: LocationOffset): SourceLocation|undefined {
-  if (!locationOffset || !sourceLocation) {
-    return sourceLocation;
+export function correctSourceRange(
+    sourceRange: SourceRange, locationOffset?: LocationOffset): SourceRange|
```
FredKSchott: if sourceRange can be undefined (you check for this below) this should be changed to `sourceRange?`?

src/elements-format.ts
```diff
@@ -15,9 +15,9 @@
 /**
  * The base interface, holding properties common to many nodes.
  */
-export interface Node {
+export interface Feature {
```
FredKSchott: Why this change?

src/generate-elements.ts
```diff
@@ -143,25 +142,21 @@ function serializeAttribute(
   return attribute;
 }
 
-function resolveSourceLocationPath(
+function resolveSourceRangePath(
     elementPath: string,
-    sourceLocation: SourceLocation|undefined): SourceLocation|undefined {
-  if (!sourceLocation) {
+    sourceRange: ResolvedSourceRange|undefined): (SourceRange | undefined) {
```
FredKSchott: `ResolvedSourceRange?`, `SourceRange?`

src/generate-elements.ts
```diff
@@ -143,25 +142,21 @@ function serializeAttribute(
...
+  if (!sourceRange) {
     return undefined;
```
FredKSchott: nit: `return undefined` -> `return;`

src/elements-format.ts
rictic: `Node` was too general, and Feature is how the rest of the code refers to this kind of thing

src/ast/inline-document-descriptor.ts
```diff
@@ -57,33 +57,39 @@ export class InlineParsedDocument<N> implements ScannedFeature {
...
+    undefined {
+  if (!locationOffset || !sourceRange) {
+    return sourceRange;
   }
-  const result: SourceLocation = {
-    line: sourceLocation.line + locationOffset.line,
-    // The location offset column only matters for the first line.
-    column: sourceLocation.column +
-        (sourceLocation.line === 0 ? locationOffset.col : 0),
+  return <SourceRange>{
```
FredKSchott: I don't think you want to cast here, I'm not 100% sure but I think this might silence errors where the object doesn't actually conform to SourceRange

src/html/html-document.ts
```diff
@@ -43,4 +45,34 @@ export class ParsedHtmlDocument extends ParsedDocument<ASTNode, HtmlVisitor> {
...
+  sourceRangeForNode(node: ASTNode): SourceRange {
+    if (!node || !node.__location) {
+      return;
+    }
+    const location = node.__location;
+    if (isElementLocationInfo(location)) {
+      return {
+        file: this.url,
+        start:
+            {line: location.startTag.line - 1, column: location.startTag.col},
```
FredKSchott: why -1 on line but not column?

src/html/html-document.ts
```diff
@@ -43,4 +45,34 @@ export class ParsedHtmlDocument extends ParsedDocument<ASTNode, HtmlVisitor> {
...
+        line: location.line - 1,
+        column: location.col + (location.endOffset - location.startOffset)
+      }
+    };
+  }
+}
+
+function isElementLocationInfo(location: parse5.LocationInfo|
+                               parse5.ElementLocationInfo):
+    location is parse5.ElementLocationInfo {
```
FredKSchott: woahhhhh I've never seen this before. Groovy

src/test/ast/inline-document-descriptor_test.ts
```diff
@@ -14,41 +14,40 @@
...
+            correctPosition({line: 1, column: 1}, {line: 1, col: 1}));
       });
 
   test('does not modify its input', function() {
-    const input: SourceLocation = {line: 5, column: 5};
+    const input: Position = {line: 5, column: 5};
     const offset: LocationOffset = {line: 5, col: 5};
-    const expected: SourceLocation = {line: 10, column: 5};
-    assert.deepEqual(expected, correctSourceLocation(input, offset));
+    const expected: Position = {line: 10, column: 5};
```
FredKSchott: wouldn't this be `column:10`?

src/ast/element-descriptor.ts
```diff
@@ -38,24 +38,25 @@ export class ScannedElement implements ScannedFeature, Resolvable {
...
 
   jsdoc?: jsdoc.Annotation;
 
   applyLocationOffset(locationOffset?: LocationOffset) {
     if (!locationOffset) {
       return;
     }
-    this.sourceLocation =
-        correctSourceLocation(this.sourceLocation, locationOffset);
+    this.sourceRange = correctSourceRange(this.sourceRange, locationOffset);
```
justinfagnani: what would it take to push this upstream to be correct when the object is created? The parser needs to know the loationOffset probably...

src/ast/source-range.ts
```diff
@@ -0,0 +1,27 @@
...
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+export interface SourceRange {
+  /* The resolved path to the file. */
+  file: string;
```
justinfagnani: shouldn't the file be implicit? A Feature should belong to a document, so if a Feature it should be possible to get SourceRange's file from it's Feature. IOW, we need something like `.document` on Features.

src/elements-format.ts
```diff
@@ -31,16 +31,26 @@ export interface Node {
   metadata?: any;
 }
 
-export interface SourceLocation {
-  /** Line number, zero indexed. */
-  line: number;
-  /** Column number, zero indexed. */
-  column: number;
+export interface SourceRange {
```
justinfagnani: why is this here and in source-range.ts?

src/elements-format.ts
```diff
@@ -31,16 +31,26 @@ export interface Node {
...
   /**
-   * Path to file, relative to the package base. If not present, is the
-   * element's file.
+   * Path to the file containing the definition of the feature,
+   * relative to the parent feature, or the package if the feature has no parent
+   * (e.g. elements).
+   *
+   * If blank, the feature is defined in the same file as its parent.
```
justinfagnani: When is a Feature not in the same file as its parent?

src/generate-elements.ts
justinfagnani: I think you mean `sourceRange?: ResolvedSourceRange`?

src/html/html-document.ts
justinfagnani: 0 vs 1 based lines? Add a comment :)

src/polymer/feature-finder.ts
```diff
@@ -43,14 +45,15 @@ export function featureFinder() {
       return;
     }
 
-    for (const prop of featureNode.properties.map(
-             esutil.toScannedPolymerProperty)) {
+    const polymerProps = featureNode.properties.map(
+        p =>
```
justinfagnani: style nit: I prefer always using parens on arrow functions (it's also a "tip" int he style guide)

src/ast/inline-document-descriptor.ts
rictic: That only works on identifier names, not on type names, so AIUI this is the only syntax for annotating return values.

src/ast/inline-document-descriptor.ts
rictic: Good catch, done

src/generate-elements.ts
rictic: Done

src/generate-elements.ts
rictic: Done, here and elsewhere.

src/ast/inline-document-descriptor.ts
rictic: Good catch, done.

src/html/html-document.ts
rictic: Added a comment. I think this gets tweaked in a downstream commit.

src/html/html-document.ts
rictic: Yeah, super useful!

src/test/ast/inline-document-descriptor_test.ts
rictic: It's initially counterintuitive, but no! We only should correct the column offset on the *first line* of the inline document. Otherwise the column offsets are correct.

src/test/ast/inline-document-descriptor_test.ts
```diff
@@ -14,41 +14,40 @@
...
+    assert.deepEqual(expected, correctPosition(input, offset));
```
rictic: Ack, I always get that wrong. Fixed here and elsewhere in this file.

src/ast/element-descriptor.ts
rictic: +1. Added a TODO

src/ast/source-range.ts
rictic: Aren't we operating on a principle of avoiding pointers back up? I'd also add that all current uses of source ranges want string file urls, which suggests to me that this is the more useful format.

src/elements-format.ts
rictic: Here the file is optional, in source-range.ts it is not.

src/elements-format.ts
rictic: Inheritance mostly

src/polymer/feature-finder.ts
rictic: Done. Can we get tslint to check for that?

rictic:
All comments addressed, PTAL

FredKSchott:
LGTM! 💥🚀🎉! 
(but wait for a LGTM from @justinfagnani before merging :)

src/elements-format.ts
justinfagnani: I don't quite grok that. Is a "parent" the same as "super"? Wouldn't a parent be an enclosing Feature, like a Element in a Document?

rictic: This is a json format, so parent is the containing object. There's no `super` concept. 

This is what I mean:

```javascript
{
"elements": [{
  "tagName": "paper-button",
  "url": "paper-button.html",
  "properties": {
    "buttonState": {
      "raised": {
        /* no file, defined in paper-button.html */
        "start": {/*...*/},
        "end": {/*...*/}
      },
    "buttonState": {
      "sourceRange": {
        "file": "./bower_components/iron-button-behavior/iron-button-behavior.html",
        "start": {/*...*/},
        "end": {/*...*/}
      },
      /*...*/
    },
    /*...*/
  }

justinfagnani: I see.

src/ast/source-range.ts
justinfagnani: I think the way we're doing resolution could mean that we don't need that rule for resolved objects, just for the things being cached. But SourceRange applies to more than resolved things, so...

justinfagnani:
LGTM++

rictic:
Manually squashed and merged onto master.

PR #292, opened 2016-08-18
Add sourcemap support to build and mocha

usergenic:
Tired of figuring out which javascript line was which typescript line when npm test failures happened so I'm adding sourcemap support.

Note: because we use gulp-typescript to incrementally build (it wont build if js file is newer than ts), you need to blow out your lib/ folder to have the maps generate if you're pulling this to existing clone.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -21,6 +21,7 @@ const eslint_lib = require('gulp-eslint');
 const mocha = require('gulp-mocha');
 const newer = require('gulp-newer');
 const shell = require('gulp-shell');
+const sourcemaps = require('gulp-sourcemaps');
 const tslint_lib = require('gulp-tslint');
 const typescript = require('gulp-typescript');
 const mergeStream = require('merge-stream');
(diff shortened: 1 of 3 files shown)


gulpfile.js
```diff
@@ -21,6 +21,7 @@ const eslint_lib = require('gulp-eslint');
 const mocha = require('gulp-mocha');
 const newer = require('gulp-newer');
 const shell = require('gulp-shell');
+const sourcemaps = require('gulp-sourcemaps');
```
rictic: `npm install --save-dev gulp-sourcemaps`?

test/mocha.opts
```diff
@@ -1,2 +1,3 @@
 --ui tdd
+--require source-map-support/register
```
rictic: `npm install --save-dev source-map-support`?

rictic:
LGTM with the above comments (and once travis is happy)

test/mocha.opts
usergenic: Weird- I thought it was in my package.json devDependencies already.  Thanks!

PR #293, opened 2016-08-21
Measure and improve resolution performance.

rictic:
Makes Document generation linear (was quadratic) by adding some simple indexes on kind and kind+identifier on the root document.

Before this change Document.makeRootDocument on all PolymerElements sources took 1500ms, after this it takes ~30ms.

Also adds a resolution benchmark runnable with `npm run benchmark`

### DIFF
--- a/src/analyzer.ts
+++ b/src/analyzer.ts
@@ -81,7 +81,7 @@ export class Analyzer {
   private _parsedDocuments =
       new Map<string, Promise<ParsedDocument<any, any>>>();
   private _scannedDocuments = new Map<string, Promise<ScannedDocument>>();
-  private _telemetry = new TelemetryTracker();
+  private _telemetryTracker = new TelemetryTracker();
 
   constructor(options: Options) {
     this._loader = options.urlLoader;
@@ -117,27 +117,28 @@ export class Analyzer {
     }
 
     const scannedDocument = await this._analyzeResolved(resolvedUrl, contents);
-    const doneTiming = this._telemetry.start('Document.makeRootDocument', url);
+    const doneTiming =
+        this._telemetryTracker.start('Document.makeRootDocument', url);
     const document = Document.makeRootDocument(scannedDocument);
     doneTiming();
     return document;
   }
 
   async getTelemetryMeasurements(): Promise<Measurement[]> {
-    return this._telemetry.getMeasurements();
+    return this._telemetryTracker.getMeasurements();
   }
 
   private async _analyzeResolved(resolvedUrl: string, contents?: string):
       Promise<ScannedDocument> {
-    let cachedResult = this._scannedDocuments.get(resolvedUrl);
+    const cachedResult = this._scannedDocuments.get(resolvedUrl);
     if (cachedResult) {
       return cachedResult;
     }
-    let promise = (async() => {
+    const promise = (async() => {
       // Make sure we wait and return a Promise before doing any work, so that
       // the Promise is cached before anything else happens.
       await Promise.resolve();
-      let document = await this._loadResolved(resolvedUrl, contents);
+      const document = await this._loadResolved(resolvedUrl, contents);
       return this._analyzeDocument(document);
     })();
     this._scannedDocuments.set(resolvedUrl, promise);
@@ -151,8 +152,8 @@ export class Analyzer {
       type: string, contents: string, url: string,
       locationOffset?: LocationOffset,
       attachedComment?: string): Promise<ScannedDocument> {
-    let resolvedUrl = this._resolveUrl(url);
-    let document = this._parse(type, contents, resolvedUrl);
+    const resolvedUrl = this._resolveUrl(url);
+    const document = this._parse(type, contents, resolvedUrl);
     return await this._analyzeDocument(
         document, locationOffset, attachedComment);
   }
@@ -163,6 +164,8 @@ export class Analyzer {
   private async _analyzeDocument(
       document: ParsedDocument<any, any>, maybeLocationOffset?: LocationOffset,
       maybeAttachedComment?: string): Promise<ScannedDocument> {
+    // TODO(rictic): We shouldn't be calling _analyzeDocument with
+    // null/undefined.
     if (document == null) {
       return null;
     }
@@ -214,7 +217,7 @@ export class Analyzer {
           }
         });
 
-    let dependencies =
+    const dependencies =
         (await Promise.all(analyzeDependencies)).filter(s => !!s);
 
     return new ScannedDocument(
@@ -233,16 +236,16 @@ export class Analyzer {
     // Use an immediately executed async function to create the final Promise
     // synchronously so we can store it in this._documents before any other
     // async operations to avoid any race conditions.
-    let promise = (async() => {
+    const promise = (async() => {
       // Make sure we wait and return a Promise before doing any work, so that
       // the Promise can be cached.
       await Promise.resolve();
       const content = providedContents == null ?
           await this._loader.load(resolvedUrl) :
           providedContents;
-      let extension = path.extname(resolvedUrl).substring(1);
+      const extension = path.extname(resolvedUrl).substring(1);
 
-      const doneTiming = this._telemetry.start('parse', 'resolvedUrl');
+      const doneTiming = this._telemetryTracker.start('parse', 'resolvedUrl');
       const parsedDoc = this._parse(extension, content, resolvedUrl);
       doneTiming();
       return parsedDoc;
@@ -253,7 +256,7 @@ export class Analyzer {
 
   private _parse(type: string, contents: string, url: string):
       ParsedDocument<any, any> {
-    let parser = this._parsers.get(type);
+    const parser = this._parsers.get(type);
     if (parser == null) {
       throw new Error(`No parser for for file type ${type}`);
     }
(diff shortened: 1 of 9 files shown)


src/analyzer.ts
```diff
@@ -80,6 +81,7 @@ export class Analyzer {
   private _parsedDocuments =
       new Map<string, Promise<ParsedDocument<any, any>>>();
   private _scannedDocuments = new Map<string, Promise<ScannedDocument>>();
+  private _telemetry = new TelemetryTracker();
```
justinfagnani: style: variable names should be a noun-phrase

src/analyzer.ts
```diff
@@ -154,6 +163,9 @@ export class Analyzer {
   private async _analyzeDocument(
       document: ParsedDocument<any, any>, maybeLocationOffset?: LocationOffset,
       maybeAttachedComment?: string): Promise<ScannedDocument> {
+    if (document == null) {
```
justinfagnani: who's calling `_analyzeDocument` with `null`? This seems like it'd be more appropriate as an assert to force the check up the call stack.

src/analyzer.ts
```diff
@@ -187,15 +205,17 @@ export class Analyzer {
...
+                await this._analyzeResolved(scannedDependency.url);
             scannedDependency.scannedDocument = scannedDocument;
             return scannedDocument;
           } else {
             throw new Error(`Unexpected dependency type: ${scannedDependency}`);
           }
         });
 
-    let dependencies = await Promise.all(analyzeDependencies);
+    let dependencies =
```
justinfagnani: `const`?

src/analyzer.ts
```diff
@@ -217,11 +237,15 @@ export class Analyzer {
       // Make sure we wait and return a Promise before doing any work, so that
       // the Promise can be cached.
       await Promise.resolve();
-      let content = providedContents == null ?
+      const content = providedContents == null ?
           await this._loader.load(resolvedUrl) :
           providedContents;
       let extension = path.extname(resolvedUrl).substring(1);
```
justinfagnani: `const`?

src/analyzer.ts
rictic: We could do it upstream, but we need to return `null` somewhere around here in the case where there's a parse error. This is handled better in a downstream commit where we actually handle it before calling _analyzeDocument but it was expedient to do it here just so that parsing all polymer elements works at all.

src/ast/document-descriptor.ts
```diff
@@ -146,6 +148,9 @@ export class Document implements Feature {
   getByKind(kind: 'import'): Set<Import>;
   getByKind(kind: string): Set<Feature>;
   getByKind(kind: string): Set<Feature> {
+    if (this._featuresByKind) {
```
justinfagnani: when do we not have `_featuresByKind`? Can we just always have it?

justinfagnani: I would make these caches, so that they're populated as the queries come in. This will discourage callers from holding references to things, if the second query is always fast.

src/ast/document-descriptor.ts
```diff
@@ -234,4 +264,30 @@ export class Document implements Feature {
 
     return result;
   }
+
+  private _featuresByKind: Map<string, Set<Feature>> = null;
+  private _featuresByKindAndId: Map<string, Map<string, Set<Feature>>> = null;
+  private _initIndexes() {
```
justinfagnani: If the indices are caches, this could be named `_prepoulateCaches` or `_warmupCaches` or something.

src/analyzer.ts
justinfagnani: Even in the case of an error, I'd rather not return `null`. I'd prefer returning some kind of `Result` object that has the document or errors, or incorporating errors into the document classes themselves. In general I'd like to push `null` handling upstream until we gotten rid of as much of it as we can.

justinfagnani: We can do this later though. `TODO`?

src/perf/parse-all-benchmark.ts
```diff
@@ -0,0 +1,155 @@
...
+    return true;
+  } catch (_) {
+    return false;
+  }
+}
+
+const fakeFileContents =
+    filesToAnalyze.map(fn => `<link rel="import" href="${fn}">`).join('\n');
+
+function padLeft(str: string, num: number): string {
```
justinfagnani: dude, there're entire packages for this

src/analyzer.ts
```diff
@@ -169,9 +181,15 @@ export class Analyzer {
       firstEntity.applyHtmlComment(maybeAttachedComment);
     }
 
-    let scannedDependencies: ScannedFeature[] = entities.filter(
+    // HACK HACK HACK(rictic): remote this in upcoming PR that propagates errors
```
FredKSchott: Hack, hack, hack the pain away!

src/perf/parse-all-benchmark.ts
```diff
@@ -0,0 +1,155 @@
...
+  if (str.length < num) {
+    return padLeft(' ' + str, num);
+  }
+  return str;
+}
+
+const measure = async() => {
```
justinfagnani: why this rather than a function declaration?

src/perf/telemetry.ts
```diff
@@ -0,0 +1,62 @@
...
+export interface Measurement {
+  elapsedTime: number;
+  kind: string;
+  identifier: string;
+}
+
+export class TelemetryTracker {
+  private _measurements: Measurement[] = [];
+  private _promises: Promise<any>[] = [];
+  async getMeasurements(): Promise<Measurement[]> {
```
justinfagnani: add newline

FredKSchott:
Consider adding a test for `lib/perf`, even just to make sure that it or the `benchmark` task runs without an error.

src/perf/telemetry.ts
```diff
@@ -0,0 +1,62 @@
...
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import * as now from 'performance-now';
+
+export interface Measurement {
+  elapsedTime: number;
+  kind: string;
```
justinfagnani: what's the difference between a kind and an identifier?

src/perf/parse-all-benchmark.ts
```diff
@@ -0,0 +1,155 @@
...
+  try {
+    fs.statSync(fn);
+    return true;
+  } catch (_) {
+    return false;
+  }
+}
+
+const fakeFileContents =
+    filesToAnalyze.map(fn => `<link rel="import" href="${fn}">`).join('\n');
```
justinfagnani: style nit: always use parens on arrow function parameters

src/perf/telemetry.ts
rictic: "parse" is a kind "./bower_components/paper-button/paper-button.html" is an identifier

src/analyzer.ts
rictic: Added TODO

src/analyzer.ts
rictic: Done

src/analyzer.ts
rictic: done

src/analyzer.ts
rictic: done

src/perf/parse-all-benchmark.ts
rictic: no no that one's called left pad. mine's padleft. so.

src/perf/parse-all-benchmark.ts
rictic: hm, I thought clang-format was misformatting it as a declaration. looks like that's not the case though.

converted to declaration

src/perf/telemetry.ts
rictic: done

src/perf/telemetry.ts
justinfagnani: so... comments? :)

src/ast/document-descriptor.ts
rictic: Done!

Had to think about it a bit, and I documented the design constraints in the code with assertions (that should never trip in the current code, but which future mistaken modifications might)

src/ast/document-descriptor.ts
rictic: I think `index` is more correct than cache here.

src/perf/parse-all-benchmark.ts
rictic: Done

rictic:
Got travis to pass/fail based on the benchmark succeeding as well.

All comments addressed PTAL

src/analyzer.ts
justinfagnani: remote -> remove

src/ast/document-descriptor.ts
justinfagnani: Do you need to push changes? Still checking for the existence of _featuresByKind

rictic: See right below, if there's not a _featuresByKind we build it and recurse

justinfagnani:
LGTM

PR #295, opened 2016-08-25
Add a warnings type and expose it usefully

rictic:


src/analyzer.ts
```diff
@@ -203,16 +201,47 @@ export class Analyzer {
...
             };
-            const scannedDocument = await this._analyzeSource(
-                scannedDependency.type, scannedDependency.contents,
-                document.url, locationOffset,
-                scannedDependency.attachedComment);
-            scannedDependency.scannedDocument = scannedDocument;
-            scannedDependency.scannedDocument.isInline = true;
-            return scannedDocument;
+            try {
+              const scannedDocument = await this._analyzeSource(
```
justinfagnani: This function is getting to be too long. Can you split it up into focused subtasks?

justinfagnani: how about one method to analyze an `InlineParsedDocument` and another to analyze `ScannedImport`. The methods name themselves even.

src/analyzer.ts
```diff
@@ -203,16 +201,47 @@ export class Analyzer {
...
+                code: 'could-not-load',
+                message: `Unable to load import: ${error.message || error}`,
+                sourceRange: scannedDependency.sourceRange,
+                severity: Severity.ERROR
+              });
+              return null;
+            }
             scannedDependency.scannedDocument = scannedDocument;
             return scannedDocument;
           } else {
```
justinfagnani: This is dead code. Maybe rewrite the filter/map so you don't feel the need for it?

src/analyzer.ts
```diff
@@ -203,16 +201,47 @@ export class Analyzer {
...
-                scannedDependency.attachedComment);
-            scannedDependency.scannedDocument = scannedDocument;
-            scannedDependency.scannedDocument.isInline = true;
-            return scannedDocument;
+            try {
+              const scannedDocument = await this._analyzeSource(
+                  scannedDependency.type, scannedDependency.contents,
+                  document.url, locationOffset,
+                  scannedDependency.attachedComment);
+              scannedDependency.scannedDocument = scannedDocument;
```
justinfagnani: why is this method mutating these objects?

src/analyzer.ts
rictic: +1, done

src/analyzer.ts
rictic: Yeah it's super ugly. I didn't see another point in time where we know the mapping between the inline doc and its scanned document.

Setting whether a scanned document is inline could be done at a number of places though.

Thoughts?

src/analyzer.ts
rictic: The above return is in a catch clause, this isn't dead.

The code is awkward to read though, pulled import and inline handling code into methods

rictic: Also renamed `analyze` to `scan` where appropriate`

rictic: IMO we should also rename `entityFinder` to `scanner`

rictic:
Rebased onto master, only the most recent two commits are new to this PR

justinfagnani:
LGTM

PR #296, opened 2016-08-25
Remove node from ScannedFeature, order by SourceRange instead.

rictic:
Slightly faster (eyeballing it, saves maybe 2% on the total runtime of parse-all), but also cleaner architecturally IMO.


src/entity/find-entities.ts
```diff
@@ -88,43 +88,36 @@ export async function findEntities(
...
-  let orhphanedEntities = entitiesByNode.values();
-  for (let entitySet of orhphanedEntities) {
-    orderedEntities.push(entitySet);
+  if (range1 == null) {
+    // ent1 comes second
+    return 1;
   }
+  const position1 = range1.start;
+  const position2 = range2.start;
+  if (position1.line < position2.line) {
```
justinfagnani: `return position1.line - position2.line || position1.column - position2.column;` is less branchy at the arguable cost of a little readability :)

src/entity/find-entities.ts
```diff
@@ -88,43 +88,36 @@ export async function findEntities(
...
+  }
+  // Lines are equal, compare by column.
+  return position1.column - position2.column;
+}
 
-  return Array.prototype.concat.apply([], orderedEntities);
+function orderEntities(unorderedEntities: ScannedFeature[][]):
+    ScannedFeature[] {
+  const allEntities: ScannedFeature[] =
+      Array.prototype.concat.apply([], unorderedEntities);
```
justinfagnani: `Array.from(unorderedEntities)`

src/entity/find-entities.ts
```diff
@@ -88,43 +88,36 @@ export async function findEntities(
...
-      }
-      entities.push(entity);
-    }
+function compareEntitiesBySourceLocation(
+    ent1: ScannedFeature, ent2: ScannedFeature): number {
+  const range1 = ent1.sourceRange;
+  const range2 = ent2.sourceRange;
+  if (range2 == null) {
+    // ent1 comes first
+    return -1;
```
justinfagnani: I think you want to handle the both null case first so you return 0.

```
if (range1 == range2) return 0;
if (range1 == null) return -1;
if (range2 == null) return 1;
```

btw, this is a case where single-line if's are nice. Our style guide allows it :)

justinfagnani:
LGTM

src/entity/find-entities.ts
rictic: Array.from doesn't concat:

```javascript
console.log(Array.from([[1,2,3], [4,5,6]]));

// [ [ 1, 2, 3 ], [ 4, 5, 6 ] ]
```

src/entity/find-entities.ts
rictic: I like having comments for the intent of the -1, 0, and 1 return values in comparators because I can never remember.

src/entity/find-entities.ts
rictic: Oh god if I read that I would have to spend so long bouncing around in my head.

Also if someone's travels lead them to this code they're probably already having a bad day.

rictic:
I pushed back on all your comments (save the checking both null cases), so do you want to take one more look before I merge?

src/entity/find-entities.ts
justinfagnani: This is actually a fairly common pattern in comparators. I personally find it a bit _more_ readable, because of the fewer if statements.

justinfagnani:
No, I wouldn't have LGTM'ed if I cared that much :) Proceed!

PR #297, opened 2016-08-25
Give completions for paths

rictic:
Also attribute import warnings to the source range for the url specifically.

### DIFF
--- a/src/url-loader/url-loader.ts
+++ b/src/url-loader/url-loader.ts
@@ -2,16 +2,37 @@
 /**
  * An object that reads files.
  */
-export interface UrlLoader {
+export abstract class UrlLoader {
   /**
    * Returns `true` if this loader can load the given `url`.
    */
-  canLoad(url: string): boolean;
+  abstract canLoad(url: string): boolean;
 
   /**
    * Reads a file from `url`.
    *
    * This should only be called if `canLoad` returns `true` for `url`.
    */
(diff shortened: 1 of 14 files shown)


src/url-loader/url-loader.ts
```diff
@@ -2,16 +2,37 @@
...
+   */
+  offersCompletions() {
+    return false;
+  }
+
+  /**
+   * Given the partial path (e.g. that a user may be in the process of typing)
+   * of `pathSoFar`, return an array of paths that pathSoFar could complete to.
+   *
+   * `pathSoFar` is a resolved URL relative to the package root.
```
justinfagnani: what's a "package root" wrt the UrlLoader interface? Maybe say it should have been resolved by a UrlResolver?

rictic: Confession: I think there's a number of bugs that will come up the first time we use a different url loader. I'd like to chat with you when you get a chance because package relative paths are an assumption that's tacitly made in a bunch of places, and I'm not sure how to think about url resolvers in general.

rictic: Also, done

justinfagnani: Yep, we can chat

rictic:
After some offline discussion, we've agreed that url loader isn't quite the right place for path completing to live. Closing this PR for now.

PR #298, opened 2016-08-25
Rename entity to feature and entityFinder to scanner

rictic:
I think we've got a solid naming system here, let's use it.

Otherwise a noop.


justinfagnani:
LGTM

PR #299, opened 2016-08-28
Add JavaScriptImportScanner

justinfagnani:
This doesn't wire the scanner in yet, but just makes it pass a simple test.

### DIFF
--- a/src/analyzer.ts
+++ b/src/analyzer.ts
@@ -57,7 +57,7 @@ export class NoKnownParserError extends Error {};
 export class Analyzer {
   private _parsers = new Map<string, Parser<ParsedDocument<any, any>>>([
     ['html', new HtmlParser()],
-    ['js', new JavaScriptParser()],
+    ['js', new JavaScriptParser({sourceType: 'script'})],
     ['css', new CssParser()],
     ['json', new JsonParser()],
   ]);
--- a/src/javascript/javascript-import-scanner.ts
+++ b/src/javascript/javascript-import-scanner.ts
@@ -1,0 +1,47 @@
+/**
+ * @license
+ * Copyright (c) 2015 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at
+ * http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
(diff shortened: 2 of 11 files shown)


src/javascript/javascript-import-scanner.ts
```diff
@@ -0,0 +1,47 @@
...
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+// import * as estraverse from 'estraverse';
```
rictic: remove?

src/javascript/javascript-import-scanner.ts
```diff
@@ -0,0 +1,47 @@
...
+  async scan(
+      document: JavaScriptDocument,
+      visit: (visitor: Visitor) => Promise<void>): Promise<ScannedImport[]> {
+    const imports: ScannedImport[] = [];
+
+    await visit({
+      enterImportDeclaration(node: estree.ImportDeclaration, _: estree.Node) {
+        const source = node.source.value as string;
+        if (!isPathImport(source)) {
+          // TODO(justinfagnani): push a warning
```
rictic: consider skipping it? otherwise we'll resolve it as though it were a relative url anyways…

justinfagnani: Oh, oops, did mean to skip it. I'll add a test.

We do need a way for a scanner to report back errors/warnings though.

rictic: Throw a WarningContainingException?

rictic: Hm, findEntities would need to be taught about it, or maybe all of the scanners...

No, an `AddWarning` method on `Scanner` is better.

src/analyzer.ts
```diff
@@ -57,7 +57,7 @@ export class NoKnownParserError extends Error {};
 export class Analyzer {
   private _parsers = new Map<string, Parser<ParsedDocument<any, any>>>([
     ['html', new HtmlParser()],
-    ['js', new JavaScriptParser()],
+    ['js', new JavaScriptParser({sourceType: 'script'})],
     ['css', new CssParser()],
     ['json', new JsonParser()],
```
rictic: You know, file extension isn't the only way we need to grab parsers.

Like, if something has been html-imported, then it needs to be handled as html. If something has been grabbed with `<script src='foo.bar>` it needs to be treated as javascript. And if something has been imported as an ES6 module it needs to be handled as a JS module.

rictic:
This is nice to see.

LGTM once there's a test for non-relative es imports.

PR #300, opened 2016-09-05
Add an out of process service for getting analyzer info

rictic:
It supports communicating over node IPCs. In a follow-up PR I'll add a stdio interface.


src/analyzer.ts
```diff
@@ -261,7 +273,7 @@ export class Analyzer {
       warnings.push({
         code: 'could-not-load',
         message: `Unable to load import: ${error.message || error}`,
-        sourceRange: scannedImport.sourceRange,
+        sourceRange: scannedImport.urlSourceRange || scannedImport.sourceRange,
```
justinfagnani: why is this logic here and not in ScannedImport?

src/ast/import-descriptor.ts
```diff
@@ -35,16 +35,21 @@ export class ScannedImport implements ScannedFeature, Resolvable {
   scannedDocument: ScannedDocument;
 
   sourceRange: SourceRange;
+  urlSourceRange: SourceRange;
```
justinfagnani: documentation please :)

src/editor-service.ts
```diff
@@ -55,6 +49,14 @@ export enum Severity {
   WARNING,
   INFO
 }
+
+export interface Position {
```
justinfagnani: this might be better named as `SourcePosition`

src/editor-service.ts
```diff
@@ -63,14 +65,40 @@ export class WarningCarryingException extends Error {
...
 }
 
-export class EditorService {
+// Important note: all arguments to, and results returned from editor service
+//     methods MUST be serializable as JSON, as the editor service may be
+//     running out of process and communicating with JSON strings.
+//
+//     Fortunately, editor-service_test will test that the results are JSON
+//     serializable.
+export abstract class BaseEditor {
```
justinfagnani: what's the split between Base and EditorService for?

src/editor-service.ts
```diff
@@ -63,14 +65,40 @@ export class WarningCarryingException extends Error {
...
+  abstract async fileChanged(localPath: string, contents?: string):
+      Promise<void>;
+
+  abstract async getDocumentationFor(localPath: string, position: Position):
+      Promise<string|undefined>;
+
+  abstract async getDefinitionFor(localPath: string, position: Position):
+      Promise<SourceRange>;
```
justinfagnani: what if there's no definition found? Should the return type be `Promise<SourceRange|undefined>`, or is that a rejection? This should be documented.

src/remote-editor-service.ts
```diff
@@ -0,0 +1,268 @@
...
+import {BaseEditor, EditorService, Position, TypeaheadCompletion, Warning} from './editor-service';
+import {FSUrlLoader} from './url-loader/fs-url-loader';
+import {PackageUrlResolver} from './url-loader/package-url-resolver';
+
+interface RequestWrapper {
+  id: number;
+  value: Request;
+}
+interface ResponseWrapper {
+  id: number;
```
justinfagnani: docs. Is `id` the id of the associated Request?

src/remote-editor-service.ts
```diff
@@ -0,0 +1,268 @@
...
+}
+interface ClearCachesRequest {
+  kind: 'clearCaches';
+}
+
+/**
+ * Runs this file (remote-editor-service.js) in a new node process and
+ * exposes a promise based request API for communicating with it.
+ */
+class SelfChannel {
```
justinfagnani: `SelfChannel`? What's that?

src/remote-editor-service.ts
```diff
@@ -0,0 +1,268 @@
...
+      default:
+        // This assignment makes it a type error if we don't handle all possible
+        // values of `message.kind`.
+        const never: never = message;
+        throw new Error(`Got unknown kind of message: ${util.inspect(never)}`);
+    }
+  }
+}
+
+function makeDeferred<V>(): Deferred<V> {
```
justinfagnani: why don't you make this a class?

src/remote-editor-service.ts
```diff
@@ -0,0 +1,268 @@
...
+    return this._channel.request({kind: 'clearCaches'});
+  }
+
+  dispose(): void {
+    this._channel.dispose();
+  }
+}
+
+/**
+ * Runs out of process and handles
```
justinfagnani: incomplete

src/remote-editor-service.ts
```diff
@@ -0,0 +1,268 @@
...
+  }
+}
+
+/**
+ * Provides a similar interface to EditorServer, but implemented out of process.
+ *
+ * This class runs in-process and communicates via SelfChannel with
+ * EditorServer, which runs in the child process.
+ */
+export class RemoteEditorService extends BaseEditor {
```
justinfagnani: There are several meaningful classes in this file, it might be nice to put them in separate files in a folder.

src/editor-service.ts
justinfagnani: Should this be `BaseEditorService`?

justinfagnani:
I think I would reorganize the files here a bit, putting things in more files under a `src/editor-service` folder. First, I'd rename `BaseEditor` to `EditorService` in it's own file with just the definitions shared by the remote and local implementations. Then I'd rename `EditorService` to `LocalEditorService`, and put all of it's implementation in another file. `RemoteEditorService` should be split between the local part that implements `EditorService` and the server. An editor plugin would import only `RemoteEditorService`, and not the server.

src/remote-editor-service.ts
rictic: Can you expand on that? What questions do you have that aren't answered by the comment on this class?

src/analyzer.ts
rictic: Grabbing one source range or the other? Or generating the entire warning given an import error?

justinfagnani: preferring one range over the other. Why are they even substitutable?

src/remote-editor-service.ts
justinfagnani: It's just a very non-self-explaining name to me. What's "self" in this context?

src/analyzer.ts
rictic: If we want to point out a problem with the url, like so:

```
<link rel="import" href="./nonexistant.html">
                        ^------------------^
```

If we don't know where the URL is, it's at least a reasonable fallback to do:

```
<link rel="import" href="./nonexistant.html">
^-------------------------------------------^
```

src/editor-service.ts
rictic: Done

src/analyzer.ts
justinfagnani: I guess that's my question: when won't we have a URL, or some other kind of import identifier?

rictic: If we had strict null checks I wouldn't worry about it, but I haven't had a chance to fix the remaining issues so I fall back to doing paranoid checks like this.

src/remote-editor-service.ts
rictic: "Self" as in this file, it speaks the protocol defined in this file, solely for the purpose of this file (as a library) talking to itself (as a process). I may be too close to it to describe it well, any suggestions?

src/remote-editor-service.ts
rictic: That seems like overkill for a private type with three properties, yeah?

src/remote-editor-service.ts
rictic: Documented the request/response protocol.

src/editor-service.ts
rictic: Renamed as you suggested

rictic:
Renamed and rearranged files as suggested.

src/remote-editor-service.ts
justinfagnani: The class will probably be smaller than the interface + function, and it'll be in one place, making it easier to read. I bet we'll find uses for it elsewhere too.

rictic:
ping @justinfagnani

src/editor-service/editor-service.ts
```diff
@@ -0,0 +1,86 @@
...
+//     methods MUST be serializable as JSON, as the editor service may be
+//     running out of process and communicating with JSON strings.
+//
+//     Fortunately, editor-service_test will test that the results are JSON
+//     serializable.
+export abstract class EditorService {
+  abstract async fileChanged(localPath: string, contents?: string):
+      Promise<void>;
+
+  abstract async getDocumentationFor(
```
justinfagnani: If this were Smalltalk or ObjC, I think ending the method name in a proposition and including the argument names would work. Here the call site won't read quite as well, especially for methods that have different signatures than this like `getDocumentationFor`. I'd go with names like `getWarningsForDocument`, `getDocumentationAtPosition`, `getDefinitionAtPosition`, etc.

src/editor-service/editor-service.ts
```diff
@@ -0,0 +1,86 @@
...
+  abstract async getDefinitionFor(localPath: string, position: SourcePosition):
+      Promise<SourceRange>;
+
+  abstract async getTypeaheadCompletionsFor  // dang clang-format..
+      (localPath: string, position: SourcePosition):
+          Promise<TypeaheadCompletion|undefined>;
+
+  abstract async getWarningsFor(localPath: string): Promise<Warning[]>;
+
+  abstract async clearCaches(): Promise<void>;
```
justinfagnani: docs

src/editor-service/editor-service.ts
```diff
@@ -0,0 +1,86 @@
...
+  abstract async fileChanged(localPath: string, contents?: string):
+      Promise<void>;
+
+  abstract async getDocumentationFor(
+      localPath: string, position: SourcePosition): Promise<string|undefined>;
+
+  abstract async getDefinitionFor(localPath: string, position: SourcePosition):
+      Promise<SourceRange>;
+
+  abstract async getTypeaheadCompletionsFor  // dang clang-format..
```
justinfagnani: you know... that comment may keep that line break there even if clang-format becomes more sane

src/editor-service/remote-editor-service.ts
```diff
@@ -0,0 +1,136 @@
...
+    const deferred = makeDeferred<any>();
+    this._outstandingRequests.set(id, deferred);
+    await this._sendRequest(id, req);
+    return deferred.promise;
+  }
+
+  private _handleResponse(response: ResponseWrapper): void {
+    const deferred = this._outstandingRequests.get(response.id);
+    if (!deferred) {
+      return;
```
justinfagnani: log a warning?

src/editor-service/remote-editor-service.ts
```diff
@@ -0,0 +1,136 @@
...
+      case 'rejection':
+        return deferred.reject(response.value.rejection);
+      default:
+        const never: never = response.value;
+        throw new Error(`Got unknown kind of response: ${util.inspect(never)}`);
+    }
+  }
+
+  private async _sendRequest(id: number, value: Request): Promise<void> {
+    const request: RequestWrapper = {id, value: value};
```
justinfagnani: might as well use shorthand for `value` too

src/editor-service/remote-editor-service.ts
```diff
@@ -0,0 +1,136 @@
...
+    const serverJsFile = path.join(__dirname, 'editor-server.js');
+    this._child = child_process.fork(serverJsFile, [], {});
+    this._child.addListener(
+        'message', (m: ResponseWrapper) => this._handleResponse(m));
+  }
+
+  async request(req: Request): Promise<any> {
+    const id = this._idCounter++;
+    const deferred = makeDeferred<any>();
+    this._outstandingRequests.set(id, deferred);
```
justinfagnani: it'd be nice to start a timer and warn if the response takes mroe than X ms

src/editor-service/remote-editor-service.ts
```diff
@@ -0,0 +1,136 @@
...
+  }
+}
+
+interface Deferred<V> {
+  resolve: (resp: V) => void;
+  reject: (err: any) => void;
+  promise: Promise<V>;
+}
+
+function makeDeferred<V>(): Deferred<V> {
```
justinfagnani: I still think that with both an interface and a factory function, that this might as well be a class.

src/test/editor-service/editor-service_test.ts
```diff
@@ -416,12 +417,151 @@ suite('EditorService', function() {
...
+    });
+
+    test(`Warn on imports of files that aren't found.`, async function() {
+      const badImport = `<link rel="import" href="./does-not-exist.html">`;
+      await editorService.fileChanged(
+          indexFile, `${badImport}\n\n${indexContents}`);
+      const warnings = await editorService.getWarningsFor(indexFile);
+      assert.containSubset(warnings, <Warning[]>[{
+                             code: 'could-not-load',
+                             severity: Severity.ERROR,
```
justinfagnani: wow clang-format

src/editor-service/editor-server.ts
```diff
@@ -0,0 +1,167 @@
...
+export interface Rejection {
+  kind: 'rejection';
+  rejection: string;
+}
+
+
+export type Request = InitRequest | FileChangedRequest | GetWarningsRequest |
+    GetDocumentationRequest | GetDefinitionRequest |
+    GetTypeaheadCompletionsRequest | ClearCachesRequest;
+export interface InitRequest {
```
justinfagnani: COnsider moving these interfaces to a file that's imported by both editor-server and remote-editor-service

src/editor-service/editor-server.ts
```diff
@@ -0,0 +1,167 @@
...
+      default:
+        // This assignment makes it a type error if we don't handle all possible
+        // values of `message.kind`.
+        const never: never = message;
+        throw new Error(`Got unknown kind of message: ${util.inspect(never)}`);
+    }
+  }
+}
+
+if (!module.parent) {
```
justinfagnani: what is `module.parent`?

src/remote-editor-service.ts
rictic: Good point. 

**repeating to himself**
```
  classes are lightweight
  classes are lightweight
  classes are lightweight
```

src/editor-service/editor-service.ts
rictic: This is true. I updated clang-format and reformatted, and lo and behold, the new version doesn't mess this up!

src/editor-service/editor-service.ts
rictic: Documented all of the methods, made clearCaches internal. (not private because abstract methods can't be private)

src/editor-service/remote-editor-service.ts
rictic: Who's the intended user of such a warning?

src/editor-service/remote-editor-service.ts
rictic: Done

src/editor-service/remote-editor-service.ts
rictic: Good point, done

src/test/editor-service/editor-service_test.ts
rictic: Looks like latest version is smarter

src/editor-service/remote-editor-service.ts
rictic: +1, done

src/editor-service/editor-server.ts
rictic: +1, done

src/editor-service/editor-server.ts
rictic: It's some sort of reference to the importing node library. editor-server now doesn't export anything and should only ever be used as a binary. Removed.

rictic:
All comments addressed, and rebased to master. PTAL

src/editor-service/editor-service.ts
rictic: PTAL

justinfagnani:
LGTM!

PR #301, opened 2016-09-06
add multi url resolver

garlicnation:


usergenic:
LGTM.  Exactly what's needed.

src/test/url-loader/multi-url-resolver_test.ts
```diff
@@ -0,0 +1,119 @@
...
+      canResolve() {
+        this.canResolveCount++;
+        return resolution != null;
+      }
+      resolve() {
+        this.resolveCount++;
+        return resolution;
+      }
+    }
+    ();
```
justinfagnani: remove?

justinfagnani: oh I see, the `new`... parens around the class expression, or can you do this without subclassing? Just pass resolution to the constructor ofMockResolver

src/test/url-loader/multi-url-resolver_test.ts
```diff
@@ -0,0 +1,119 @@
...
+  });
+};
+
+const firstResolverTrue = () =>
```
justinfagnani: are these declarations buying anything? Consider inlining into tests

src/url-loader/multi-url-resolver.ts
```diff
@@ -0,0 +1,41 @@
...
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import {UrlResolver} from './url-resolver';
+
+/**
+ * Resolves a URL using multiple resolvers.
+ */
+export class MultiUrlResolver implements UrlResolver {
+  constructor(private resolvers: Array<UrlResolver>) {
```
justinfagnani: let's name all private fields with a leading underscore, so there's some indication in compiled code.

src/url-loader/multi-url-resolver.ts
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  resolve(url: string): string {
+    for (let i = 0; i < this.resolvers.length; i++) {
+      let resolver = this.resolvers[i];
+      if (resolver.canResolve(url)) {
+        return resolver.resolve(url);
+      }
+    }
+  }
```
justinfagnani: should this throw here?

usergenic: Fair point, our reference example package-url-resolver throws if its given a url to resolve that canResolve returned false for:

```
  canResolve(url: string): boolean {
    let urlObject = parseUrl(url);
    let pathname =
        pathlib.normalize(decodeURIComponent(urlObject.pathname || ''));
    return this._isValid(urlObject, pathname);
  }

  _isValid(urlObject: Url, pathname: string) {
    return (urlObject.hostname === this.hostname || !urlObject.hostname) &&
        !pathname.startsWith('../../');
  }

  resolve(url: string): string {
    let urlObject = parseUrl(url);
    let pathname =
        pathlib.normalize(decodeURIComponent(urlObject.pathname || ''));

    if (!this._isValid(urlObject, pathname)) {
      throw new Error(`Invalid URL ${url}`);
    }
```

I feel like we should add a comment to url-resolver.ts about the expected behavior of resolve() when provided an unresolveable url.

rictic: Hm, I guess it partly depends on what not returning from a resolver should mean. In a lot of situations no URLs will need to be resolved. Do we want to require that people use a final noop resolver in that case?

usergenic: I think its more a matter of the relationship between `canResolve()` and `resolve()`.  It seems like `canResolve()` returning `false` would indicate that `resolve()` should throw for same URL.  Don't know if we need an explicit noop if caller knows not to call resolve for the false cases.  That assumes that caller always calls `canResolve()` first though.  Not sure if that's the intended API.

justinfagnani: If the behavior we want out of callers is that they only call `resolve` when `canResolve()` is `true`, then throwing is the correct behavior. Otherwise, even if you default to the original url, you have inconsistent answers. We could default, but then I'd say that `canResolve()` should just return `true`.

usergenic: MultiUrlResolver is actually a UseTheFirstOneThatCanResolveUrlResolver; would we anticipate strategies like ApplyAnyThatCanResolveInSequenceMultiUrlResolver?

justinfagnani: @usergenic What would be the difference? AnyResolve would randomly pick one that can resolve? :P

usergenic: I guess it would be where the result of the prior resolve would be fed to the subsequent resolver.  Kind of like a filter.  Maybe if it were needed it could be called `CascadingMultiUrlResolver`.  I don't have an uncontrived use case in mind.

src/test/url-loader/multi-url-resolver_test.ts
garlicnation: done

src/test/url-loader/multi-url-resolver_test.ts
garlicnation: done

src/url-loader/multi-url-resolver.ts
garlicnation: done

src/url-loader/multi-url-resolver.ts
garlicnation: yes, done

justinfagnani:
LGTM!

PR #303, opened 2016-09-07
Fix gulp build to generate the .d.ts files in lib/.

usergenic:
Perhaps in a recent version of gulp-typescript the typings files were no longer being output.  Either that or they were never output since switching to gulp-typescript.  (was that my bad when I added the sourcemaps support?)  I don't know... However, this makes them output.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -74,12 +74,13 @@ task('build', ['compile', 'json-schema']);
 
 task('compile', function() {
   const srcs = gulp.src('src/**/*.ts').pipe(newer({dest: 'lib', ext: '.js'}));
+  const tsResult = srcs.pipe(sourcemaps.init()).pipe(typescript(tsProject));
+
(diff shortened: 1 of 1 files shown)


rictic:
Oh nice! I think this was just never hooked up.

LGTM!

Fodder for a followup: do you know if it's possible to fail the build on compile errors?

usergenic:
Well, I think it fully honors the `tsconfig.json` `compilerOptions` -- this fix just ensures that we use gulp-typescript's separate .dts stream.  It fails spectacularly, for example, if you un-comment `strictNullChecks: true` in the `tsconfig.json`

usergenic:
*oh* i see you said fail the `build` like prevent next step etc.  will look into.

rictic:
Feel free to merge this PR first though, because there are typescript errors at HEAD (due to typescript releasing RC1 with some new checks, will be fixed once #300 lands)

PR #304, opened 2016-09-08
Fix warnings found in latest typescript compiler version.

rictic:
Split out of the commit in #300 since that's not gonna land today.

### DIFF
--- a/src/test/elements-metadata_tests.ts
+++ b/src/test/elements-metadata_tests.ts
@@ -193,6 +193,6 @@ const paperButton: Elements = {
 // We want to keep these local values as is, because they serve as
 // documentation and compilation tests.
 // So there's this hack to prevent a warning about unused code.
-if (false !== true) {
+if (Math.random() > 1000) {
   console.log(JSON.stringify(paperButton, null, 2));
 }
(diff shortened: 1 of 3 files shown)


src/test/elements-metadata_tests.ts
```diff
@@ -193,6 +193,6 @@ const paperButton: Elements = {
 // We want to keep these local values as is, because they serve as
 // documentation and compilation tests.
 // So there's this hack to prevent a warning about unused code.
-if (false !== true) {
+if (Math.random() > 1000) {
```
usergenic: Shouldn't this be Math.random() < 1000? (if the same result expected as false !== true)

usergenic: I recognize this probably doesn't matter.

usergenic:
Apart from my pedantic nit about using Math.random() > 1000 for both true and false, this LGTM.

src/test/elements-metadata_tests.ts
rictic: Arg, the old code should have been `false === true`.

shamecube.gif

rictic:
Thanks for the review!

PR #306, opened 2016-09-14
Created a CssImportScanner to support deprecated, but still supported, polymer-specific <link rel="import" type="css"> imports.

usergenic:
- Created a new CssImportScanner and test suite.
 - Include the new scanner in Analyzer's html document scanners list.
 - Added a new Analyzer test.

Note: required for bundler to support css inlining.

### DIFF
--- a/src/polymer/css-import-scanner.ts
+++ b/src/polymer/css-import-scanner.ts
@@ -1,0 +1,47 @@
+/**
+ * @license
+ * Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at
+ * http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import * as dom5 from 'dom5';
+import {resolve as resolveUrl} from 'url';
+
+import {ScannedImport} from '../ast/ast';
+
+import {HtmlVisitor, ParsedHtmlDocument} from '../html/html-document';
+import {HtmlScanner} from '../html/html-scanner';
+
+const p = dom5.predicates;
+
+const isCssImportNode = p.AND(p.hasTagName('link'), (node) => {
+  const rel = dom5.getAttribute(node, 'rel') || '';
+  return rel.split(' ').indexOf('import') !== -1;
+}, p.hasAttrValue('type', 'css'));
+
+export class CssImportScanner implements HtmlScanner {
+  async scan(
+      document: ParsedHtmlDocument,
+      visit: (visitor: HtmlVisitor) => Promise<void>):
+      Promise<ScannedImport[]> {
+    const imports: ScannedImport[] = [];
+
+    await visit((node) => {
+      if (isCssImportNode(node)) {
+        const href = dom5.getAttribute(node, 'href');
+        const importUrl = resolveUrl(document.url, href);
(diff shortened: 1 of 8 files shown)


src/polymer/css-import-scanner.ts
```diff
@@ -0,0 +1,47 @@
...
+
+import {ScannedImport} from '../ast/ast';
+
+import {HtmlVisitor, ParsedHtmlDocument} from '../html/html-document';
+import {HtmlScanner} from '../html/html-scanner';
+
+const p = dom5.predicates;
+
+const isCssImportNode = p.AND(p.hasTagName('link'), (node) => {
+  const rel = dom5.getAttribute(node, 'rel') || '';
```
rictic: You know, this would make a nice predicate factory. Also useful for detecting `class`es. Something like:

```typescript
dom5.predicates.hasSpaceSeparatedAttributeValue('rel', 'import');
```

src/test/analyzer_test.ts
```diff
@@ -168,6 +168,21 @@ suite('Analyzer', () => {
           ['polymer.html', 'foo.js', 'foo.css']);
     });
 
+    test('polymer css import scanner', async() => {
+      let contents = `<html><head>
+          <link rel="import" type="css" href="bar.css">
```
rictic: Is this valid? I think we might only support that kind of import inside of a polymer template

src/test/polymer/css-import-scanner_test.ts
```diff
@@ -0,0 +1,54 @@
...
+
+     test('finds CSS Imports', async() => {
+       let contents = `<html><head>
+           <link rel="import" href="polymer.html">
+           <link rel="import" type="css" href="polymer.css">
+           <script src="foo.js"></script>
+           <link rel="stylesheet" href="foo.css"></link>
+         </head></html>`;
+       let ast = parse5.parse(contents);
+       let document = new ParsedHtmlDocument({
```
rictic: Recommend importing and using the HtmlParser class rather than doing our own parsing for the test. That way the test environment will more closely resemble the production environment.

src/test/analyzer_test.ts
usergenic: Oh, crumbs. 😨   It hadn't occurred to me that I had to be a bit *smarter* about this.  Putting thinking cap back on, looking up documentation for the feature...

According to https://www.polymer-project.org/1.0/docs/devguide/styling#external-stylesheets the link should live in the `<dom-module>` but outside the `<template>`.

I expect this is because `<template>` contents are parsed but not "active" or what-not.  On that note, do we intend that the analyzer flag or exclude certain features during scan when they are encountered inside <template> tags for this reason?

rictic: I think it's worth bringing up a little example element locally to try to figure out how Polymer handles these links. (I know that they're widely used by some of our very important customers, or I might not push on this as much, as it **is** deprecated).

As for `<template>` contents, I think the goal should be to try hard to do the Correct thing in Analyzer. I'm confident that we're going to end up treating Polymer template contents (i.e. template contents inside `<dom-module>`s and `<template is="dom-bind">` specially).

src/polymer/css-import-scanner.ts
usergenic: Challenge accepted.  https://github.com/PolymerLabs/dom5/pull/36

src/test/polymer/css-import-scanner_test.ts
usergenic: Okay.  I modified the `html-import-scanner_test.ts` I cargo'd style from to use HtmlParser too.

src/test/analyzer_test.ts
usergenic: It turns out that there's already magic in analyzer preventing the scanner from diving into generic `<template>`.  I haven't tried with `<template is="dom-bind">` but I did modify test to demonstrate the analzyer's behavior vis-a-vis `<dom-module>` and `<template>`.

PR #307, opened 2016-09-16
support lazy html imports

garlicnation:


garlicnation:
@justinfagnani About to push the manual lazy feature inclusion we discussed over VC earlier.

src/html/html-import-scanner.ts
```diff
@@ -22,9 +22,28 @@ import {HtmlScanner} from './html-scanner';
...
+    notCssLink,
+    p.parentMatches(
+      p.AND(
+        p.hasTagName('dom-module'),
+        p.NOT(p.hasTagName('template'))
+      )
+    )
+  );
 
 export class HtmlImportScanner implements HtmlScanner {
```
justinfagnani: Could you ad a comment that mentions lazy imports?

justinfagnani:
I need to get used to the fact that some comments don't go out immediately, and not forget to send them :)

justinfagnani:
This is LGTM, if you want to merge it first.

src/html/html-import-scanner.ts
garlicnation: Done.

src/html/html-import-scanner.ts
```diff
@@ -22,9 +22,28 @@ import {HtmlScanner} from './html-scanner';
...
+    linkTag,
+    p.hasSpaceSeparatedAttrValue('rel', 'import'),
+    notCssLink,
+    p.NOT(
+      p.parentMatches(p.hasTagName('template')))
+);
+
+const isLazyImportNode = p.AND(
+    p.hasTagName('link'),
+    p.hasSpaceSeparatedAttrValue('rel', 'lazy-import'),
```
garlicnation: Changed the logic - now it bails if it encounters that.

PR #309, opened 2016-09-17
Small cleanups

justinfagnani:
Started moving things to better files, a lot more to do still.


src/ast/ast.ts
```diff
@@ -32,11 +32,13 @@
  * Earlier stages have the longer names, like ParsedDocument and ScannedElement.
  */
 
-export * from './descriptor';
+export * from './scanned-feature';
```
justinfagnani: weird, I expected the formatter to sort exports... I'll fix this.

PR #310, opened 2016-09-18
Initial pass at printing warnings to the command line

rictic:
A basic class for formatting warnings and errors in a helpful way. Produces output like this, but with colors too:

```
class ClassDeclaration extends HTMLElement {}
      ~~~~~~~~~~~~~~~~

vanilla-elements.js(0,6) warning [dumb-element-name] - This is a dumb name for an element.
```

Should be useful for the linter of course, but should generally be useful for anything downstream of the analyzer that needs to report problems and possibly bail out early.

### DIFF
--- a/custom_typings/memory-streams.d.ts
+++ b/custom_typings/memory-streams.d.ts
@@ -1,0 +1,6 @@
+declare module 'memory-streams' {
+  import {Readable, Writable} from 'stream';
+  export const ReadableStream: {new (contents: string): Readable;};
+
+  export const WritableStream: {new (): Writable;};
+}
--- a/custom_typings/memory-streams.d.ts
+++ b/custom_typings/memory-streams.d.ts
@@ -1,6 +1,9 @@
 declare module 'memory-streams' {
   import {Readable, Writable} from 'stream';
-  export const ReadableStream: {new (contents: string): Readable;};
 
-  export const WritableStream: {new (): Writable;};
+  export class ReadableStream extends Readable {
+    constructor(contents: string);
+  }
+
+  export class WritableStream extends Writable {}
 }
--- a/src/analyzer.ts
+++ b/src/analyzer.ts
@@ -282,6 +282,13 @@ export class Analyzer {
     return scannedDocument;
   }
 
+  /**
+   * Loads the content at the provided resolved URL.
+   *
+   * Currently does no caching. If the provided contents are given then they
+   * are used instead of hitting the UrlLoader (e.g. when you have in-memory
+   * contents that should override disk).
+   */
   async load(resolvedUrl: string, providedContents?: string) {
     if (!this._loader.canLoad(resolvedUrl)) {
       throw new Error(`Can't load URL: ${resolvedUrl}`);
--- a/src/analyzer.ts
+++ b/src/analyzer.ts
@@ -120,7 +120,7 @@ export class Analyzer {
       this._parsedDocuments.delete(resolvedUrl);
     }
 
-    const scannedDocument = await this._scanResolved(resolvedUrl, contents);
+    const scannedDocument = await this._scan(resolvedUrl, contents);
     const doneTiming =
         this._telemetryTracker.start('Document.makeRootDocument', url);
     const document = Document.makeRootDocument(scannedDocument);
@@ -144,7 +144,7 @@ export class Analyzer {
     this._scannedDocuments.clear();
   }
 
-  private async _scanResolved(resolvedUrl: string, contents?: string):
+  private async _scan(resolvedUrl: string, contents?: string):
       Promise<ScannedDocument> {
     const cachedResult = this._scannedDocuments.get(resolvedUrl);
     if (cachedResult) {
@@ -154,7 +154,7 @@ export class Analyzer {
       // Make sure we wait and return a Promise before doing any work, so that
       // the Promise is cached before anything else happens.
       await Promise.resolve();
-      const document = await this._loadResolved(resolvedUrl, contents);
+      const document = await this._parse(resolvedUrl, contents);
       return this._scanDocument(document);
     })();
     this._scannedDocuments.set(resolvedUrl, promise);
@@ -169,7 +169,7 @@ export class Analyzer {
       locationOffset?: LocationOffset,
       attachedComment?: string): Promise<ScannedDocument> {
     const resolvedUrl = this._resolveUrl(url);
-    const document = this._parse(type, contents, resolvedUrl);
+    const document = this._parseContents(type, contents, resolvedUrl);
     return await this._scanDocument(document, locationOffset, attachedComment);
   }
 
@@ -262,8 +262,7 @@ export class Analyzer {
       // HACK(rictic): this isn't quite right either, we need to get
       //     the scanned dependency's url relative to the basedir don't
       //     we?
-      scannedDocument =
-          await this._scanResolved(this._resolveUrl(scannedImport.url));
+      scannedDocument = await this._scan(this._resolveUrl(scannedImport.url));
     } catch (error) {
       if (error instanceof NoKnownParserError) {
         // We probably don't want to fail when importing something
@@ -283,15 +282,21 @@ export class Analyzer {
     return scannedDocument;
   }
 
-  private async _loadResolved(resolvedUrl: string, providedContents?: string):
+  async load(resolvedUrl: string, providedContents?: string) {
+    if (!this._loader.canLoad(resolvedUrl)) {
+      throw new Error(`Can't load URL: ${resolvedUrl}`);
+    }
+    return providedContents == null ? await this._loader.load(resolvedUrl) :
+                                      providedContents;
+  }
+
+  private async _parse(resolvedUrl: string, providedContents?: string):
       Promise<ParsedDocument<any, any>> {
     const cachedResult = this._parsedDocuments.get(resolvedUrl);
     if (cachedResult) {
       return cachedResult;
     }
-    if (!this._loader.canLoad(resolvedUrl)) {
-      throw new Error(`Can't load URL: ${resolvedUrl}`);
-    }
+
     // Use an immediately executed async function to create the final Promise
     // synchronously so we can store it in this._documents before any other
     // async operations to avoid any race conditions.
@@ -299,13 +304,12 @@ export class Analyzer {
       // Make sure we wait and return a Promise before doing any work, so that
       // the Promise can be cached.
       await Promise.resolve();
-      const content = providedContents == null ?
-          await this._loader.load(resolvedUrl) :
-          providedContents;
+
+      const content = await this.load(resolvedUrl, providedContents);
       const extension = path.extname(resolvedUrl).substring(1);
 
       const doneTiming = this._telemetryTracker.start('parse', 'resolvedUrl');
-      const parsedDoc = this._parse(extension, content, resolvedUrl);
+      const parsedDoc = this._parseContents(extension, content, resolvedUrl);
       doneTiming();
       return parsedDoc;
     })();
@@ -313,7 +317,7 @@ export class Analyzer {
     return promise;
   }
 
-  private _parse(type: string, contents: string, url: string):
+  private _parseContents(type: string, contents: string, url: string):
       ParsedDocument<any, any> {
     const parser = this._parsers.get(type);
     if (parser == null) {
--- a/src/analyzer.ts
+++ b/src/analyzer.ts
@@ -17,7 +17,6 @@
 import * as path from 'path';
 
 import {CssParser} from './css/css-parser';
-import {Severity, Warning, WarningCarryingException} from './editor-service/editor-service';
 import {HtmlImportScanner} from './html/html-import-scanner';
 import {HtmlParser} from './html/html-parser';
 import {HtmlScriptScanner} from './html/html-script-scanner';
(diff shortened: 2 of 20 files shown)


custom_typings/memory-streams.d.ts
```diff
@@ -0,0 +1,6 @@
+declare module 'memory-streams' {
+  import {Readable, Writable} from 'stream';
+  export const ReadableStream: {new (contents: string): Readable;};
```
justinfagnani: why a property rather than a class?

src/polymer-lint.ts
```diff
@@ -0,0 +1,56 @@
...
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import {Analyzer} from './analyzer';
```
justinfagnani: I would like to either put this in a demo directory, or not check it in and contribute it to the linter.

src/test/warnings_test.ts
```diff
@@ -0,0 +1,85 @@
...
+    end: {column: 22, line: 0}
+  }
+};
+
+const staticTestDir = path.join(__dirname, 'static');
+
+suite('WarningPrinter', () => {
+  let output: NodeJS.WritableStream;
+  let printer: WarningPrinter;
+  setup(() => {
```
justinfagnani: nit: add newline above

src/test/warnings_test.ts
```diff
@@ -0,0 +1,85 @@
...
+    output = new memoryStreams.WritableStream();
+    printer =
+        new WarningPrinter(output, {basedir: staticTestDir, color: false});
+  });
+  test('can handle printing no warnings', () => {
```
justinfagnani: nit: add newline above

src/polymer-lint.ts
```diff
@@ -0,0 +1,56 @@
+/**
```
FredKSchott: Is this file meant to be checked in, or is it just included to show us how warnings can be interacted with?

src/warnings.ts
```diff
@@ -0,0 +1,156 @@
...
+};
+
+export class WarningPrinter {
+  _chalk: typeof chalk;
+  constructor(
+      private _outStream: NodeJS.WritableStream,
+      private _options?: PrinterOptions) {
+    this._options = Object.assign({}, defaultPrinterOptions, _options);
+    this._chalk =
+        new (chalk.constructor as any)({enabled: this._options.color});
```
FredKSchott: What's wrong here that we need to use `any`? Is this something that can be fixed in the @types declaration?

src/warnings.ts
```diff
@@ -0,0 +1,156 @@
...
+ */
+
+import * as chalk from 'chalk';
+import * as fs from 'fs';
+import * as path from 'path';
+
+import {SourceRange} from './ast/source-range';
+import {Severity, Warning} from './editor-service';
+
+export enum Verbosity {
```
FredKSchott: Can we make this a enum with string values so that non-typescript consumers can have a easier time?

justinfagnani: this file could really use some comments

src/warnings.ts
```diff
@@ -0,0 +1,156 @@
...
+        return this._chalk.green;
+      default:
+        const never: never = severity;
+        throw new Error(`Unknown severity value - ${never
+                        } - encountered while printing warning.`);
+    }
+  }
+  private _getTextOfLine(line: number, localPath: string) {
+    const contents =
+        fs.readFileSync(path.join(this._options.basedir, localPath), 'utf-8');
```
FredKSchott: What about projects with custom resolvers and loaders?

src/warnings.ts
```diff
@@ -0,0 +1,156 @@
...
+  Full,
+}
+
+export interface PrinterOptions {
+  basedir: string;
+  verbosity?: Verbosity;
+  color?: boolean;
+}
+export interface FilterOptions {
+  warningCodesToIgnore?: Set<string>;
```
justinfagnani: what are the valid strings?

src/warnings.ts
```diff
@@ -0,0 +1,156 @@
...
+  }
+  private _getTextOfLine(line: number, localPath: string) {
+    const contents =
+        fs.readFileSync(path.join(this._options.basedir, localPath), 'utf-8');
+    return contents.split('\n')[line];
+  }
+}
+
+function repeatCharacter(char: string, num: number) {
+  const chars: string[] = [];
```
FredKSchott: code golf: could also be `Array(num).join(char)` :)

src/warnings.ts
```diff
@@ -0,0 +1,156 @@
...
+  minimumSeverity: Severity;
+}
+
+const defaultFilterOptions: FilterOptions = {
+  warningCodesToIgnore: new Set(),
+  minimumSeverity: Severity.INFO
+};
+
+export class WarningFilterer {
+  constructor(private _options: FilterOptions) {
```
justinfagnani: aren't you initializing `_options` twice?

src/warnings.ts
```diff
@@ -0,0 +1,156 @@
...
+  warningCodesToIgnore?: Set<string>;
+  minimumSeverity: Severity;
+}
+
+const defaultFilterOptions: FilterOptions = {
+  warningCodesToIgnore: new Set(),
+  minimumSeverity: Severity.INFO
+};
+
+export class WarningFilterer {
```
justinfagnani: looks like this could be named `WarningFilter`

src/warnings.ts
```diff
@@ -0,0 +1,156 @@
...
+}
+
+const defaultPrinterOptions = {
+  verbosity: Verbosity.Full,
+  color: true
+};
+
+export class WarningPrinter {
+  _chalk: typeof chalk;
+  constructor(
```
justinfagnani: nit: add some newlines. they're free :)

src/warnings.ts
justinfagnani: :(

src/warnings.ts
```diff
@@ -0,0 +1,156 @@
...
+      private _outStream: NodeJS.WritableStream,
+      private _options?: PrinterOptions) {
+    this._options = Object.assign({}, defaultPrinterOptions, _options);
+    this._chalk =
+        new (chalk.constructor as any)({enabled: this._options.color});
+  }
+
+  printWarnings(warnings: Iterable<Warning>) {
```
justinfagnani: would there be any benefit to formulating this as a mapping function from `Iterable<Warning>` -> `Iterable<String>`?

```typescript
const printer = warningPrinter(options);
warnings.map(printer).forEach(w => outputStream.write(w));
```

src/warnings.ts
```diff
@@ -0,0 +1,156 @@
...
+    }
+  }
+  private _getTextOfLine(line: number, localPath: string) {
+    const contents =
+        fs.readFileSync(path.join(this._options.basedir, localPath), 'utf-8');
+    return contents.split('\n')[line];
+  }
+}
+
+function repeatCharacter(char: string, num: number) {
```
justinfagnani: move to a utilities module?

src/warnings.ts
rictic: Yes! Good catch. Upstreamed with https://github.com/DefinitelyTyped/DefinitelyTyped/pull/11411 and added a PR to remove the cast once it lands.

rictic: Upstreamed as https://github.com/DefinitelyTyped/DefinitelyTyped/pull/11411 and added TODO

src/warnings.ts
rictic: Counterpoint: YAGNI, move it there the first time we need to use it in another file?

src/polymer-lint.ts
rictic: Moved to `src/demo`

src/polymer-lint.ts
rictic: Moved to `src/demo`

src/warnings.ts
rictic: Warning paths must be resolved but switched to urlLoader. This made everything async, but I don't think that's a problem.

src/warnings.ts
rictic: Clever!

(`num + 1` as it turns out, but I didn't catch that either until the tests failed)

src/warnings.ts
rictic: Typescript doesn't support enums like that. We could make it a union of string literals though, like: 

```typescript
export type Verbosity = 'one-line' | 'full';
```

OTOH, this works in JS with the enum: 

```javascript
Verbosity.OneLine 
// => 0
Verbosity[0]
// => 'OneLine'
```

I'm kinda leaning to the former. What do you think @justinfagnani ?

src/warnings.ts
justinfagnani: Why isn't this reading documents from an Analyzer instance?

src/warnings.ts
FredKSchott: +1 to refactoring out only if needed

justinfagnani: Coutercouterpoint: you might not find it. We'll have to just remember that we have it.

src/warnings.ts
justinfagnani: even better: `char.repeat(num)`

custom_typings/memory-streams.d.ts
rictic: Oh nice. I thought Readable was an interface not a class is all.

src/warnings.ts
justinfagnani: countercountercounterpoint: just use String#repeat

src/test/warnings_test.ts
rictic: Done

src/warnings.ts
rictic: Done

src/warnings.ts
rictic: Done

src/warnings.ts
rictic: yes, once with the passed value and again with defaults mixed in

src/warnings.ts
rictic: Done

src/test/warnings_test.ts
rictic: Done

src/warnings.ts
rictic: Done

src/warnings.ts
rictic: Good point, done

src/warnings.ts
rictic: It's an open set and we don't have many examples in the codebase currently, but I added some docs that may help?

src/warnings.ts
rictic: `printWarnings` is a convenience method around `printWarning`. Now that it's async it's a bit more complicated to do an iterable -> iterable mapping.

src/warnings.ts
rictic: Done

rictic:
Rebased to master and all comments addressed. PTAL

src/warnings.ts
justinfagnani: strings

src/analyzer.ts
```diff
@@ -283,37 +282,42 @@ export class Analyzer {
     return scannedDocument;
   }
 
-  private async _loadResolved(resolvedUrl: string, providedContents?: string):
+  async load(resolvedUrl: string, providedContents?: string) {
```
justinfagnani: comments. Mention that this doesn't cache?

src/warnings.ts
FredKSchott: Oh nice, I'd still prefer string literals but I didn't know TS would create a JS `Verbosity` object /w `OneLine` property.

src/warnings.ts
```diff
@@ -0,0 +1,160 @@
...
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import * as chalk from 'chalk';
+
+import {Analyzer} from './analyzer';
+import {Severity, Warning} from './editor-service/editor-service';
```
justinfagnani: we should move these to this file and import them into editor-service

src/warnings.ts
FredKSchott: @justinfagnani wins code golf

justinfagnani:
Looks nearly good. I'd create a `warnings/` dir and move Warning and Severity from editor-service to a file in there. Also, consider creating warning-filter and warning-printer files.

src/warnings.ts
rictic: +1

src/analyzer.ts
rictic: Comments longer than implementation! Still worth. Done.

rictic:
All comments addressed.

Moving `Warning` and `Severity` into `warning` directory is great, way cleaner, and eliminates all reaching into the editor-service directory, which is ~Correct~.

PR #312, opened 2016-09-22
Add a stdin/stdout interface to editor-server.

rictic:
Useful for extending editors like Sublime Text that (for some reason) aren't written in Node and can't use the Node IPC interface. The protocol is the same, this just extends editor-service.ts to speak the same protocol also over stdin/stdout.

### DIFF
--- a/src/editor-service/editor-server.ts
+++ b/src/editor-service/editor-server.ts
@@ -12,6 +12,16 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
+/**
+ * This file is a binary not a library, and should be run via
+ * `node editor-server.js` or `child_process.fork`
+ *
+ * Communication with this server is done via the Remote Editor Protocol via
+ * stdin and stdout, as well as the node process.on('message') IPC process.
+ *
+ * See 'remote-editor-protocol.ts' for details on the communication protocol.
+ */
+
 import * as split from 'split';
 import * as util from 'util';
 
--- a/src/editor-service/editor-server.ts
+++ b/src/editor-service/editor-server.ts
@@ -12,6 +12,7 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
+import * as split from 'split';
 import * as util from 'util';
 
 import {FSUrlLoader} from '../url-loader/fs-url-loader';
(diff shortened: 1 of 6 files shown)


src/editor-service/editor-server.ts
```diff
@@ -21,74 +22,78 @@ import {LocalEditorService} from './local-editor-service';
...
-  server = new EditorServer(initRequest.value.basedir);
 
-  process.on('message', async(request: RequestWrapper) => {
-    const result = await getSettledValue(request.value);
-    process.send(<ResponseWrapper>{id: request.id, value: result});
-  });
+const server: EditorServer = new EditorServer();
+
+// stdin/stdout interface
+process.stdin.setEncoding('utf8');
```
FredKSchott: can this logic be wrapped in a function or process event listener? Right now it looks like there are side effects to just loading this file which we generally want to avoid

src/test/editor-service/editor-server-test.ts
```diff
@@ -0,0 +1,155 @@
...
+                           file: 'malformed.html',
+                           start: {line: 266, column: 0},
+                           end: {line: 266, column: 0}
+                         }
+                       }]);
+    });
+  }
+
+  let suiteName =
+      'from node with child_process.fork() and process.send() for IPC';
```
FredKSchott: stylistic bikeshed: I hate using heavier stuff like creating unnecessary variables to get around character limits, can we ignore or do something more like 

```javascript
suite(
  'from node with child_process.fork() and process.send() for IPC',
  function() {
```

justinfagnani:
Can you add a description? What's this for?

src/editor-service/editor-server.ts
justinfagnani: +1

Also, how do the stdin and IPC interfaces interact? Any benefit from setting up one or the other?

src/test/editor-service/editor-server-test.ts
```diff
@@ -0,0 +1,155 @@
...
+      localPath: 'malformed.html'
+    };
+
+
+    test('can create and initialize', async function() {
+      await sendRequest({id: 0, value: initMessage});
+      const response = await getNextResponse(0);
+      assert.deepEqual(response, undefined);
+    });
+    test('initializing twice is an error', async function() {
```
justinfagnani: nit: let the code breeeeathe

src/test/editor-service/editor-server-test.ts
```diff
@@ -0,0 +1,155 @@
...
+      kind: 'init',
+      basedir: `${path.join(__dirname, 'static')}`
+    };
+    const getWarningsMessage = {
+      kind: 'getWarningsFor',
+      localPath: 'malformed.html'
+    };
+
+
+    test('can create and initialize', async function() {
```
justinfagnani: nit: `function ()` -> `() =>`

src/test/editor-service/editor-server-test.ts
justinfagnani: +1, also arrow functions

src/test/editor-service/editor-server-test.ts
```diff
@@ -0,0 +1,155 @@
...
+      child.stdout.setEncoding('utf8');
+      child.stdout.resume();
+      lines = child.stdout.pipe(split());
+    });
+    teardown(() => {
+      child.kill();
+    });
+
+    async function sendRequest(message: any) {
+      return new Promise<void>((resolve, reject) => {
```
justinfagnani: Since you're in an async function, throw an error rather than create a new Promise.

src/test/editor-service/editor-server-test.ts
```diff
@@ -0,0 +1,155 @@
...
+        });
+      });
+    };
+
+    async function getNextResponse(expectedId: number) {
+      const line =
+          await new Promise<string>(resolve => lines.once('data', resolve));
+      const message = JSON.parse(line);
+      assert.equal(message.id, expectedId);
+      return new Promise((resolve, reject) => {
```
justinfagnani: do you need this Promise? Just return or throw

src/test/editor-service/editor-server-test.ts
```diff
@@ -0,0 +1,155 @@
...
+          reject(message.value.rejection);
+        }
+      });
+    };
+
+    editorServiceInterfaceTests(sendRequest, getNextResponse);
+  });
+
+  suiteName = 'from the command line with stdin and stdout';
+  suite(suiteName, function() {
```
justinfagnani: try to eliminate the variable as above

src/editor-service/editor-server.ts
rictic: This file exports nothing and should only ever be run directly with `node editor-server.js` or `child_process.fork`. Added docs.

src/test/editor-service/editor-server-test.ts
rictic: +1. This was mostly to work around an issue with clang-format, which was super indenting test bodies very deeply when test names wrapped.

src/editor-service/editor-server.ts
```diff
@@ -21,74 +22,78 @@ import {LocalEditorService} from './local-editor-service';
...
-    this._localEditorService = new LocalEditorService({
-      urlLoader: new FSUrlLoader(basedir),
-      urlResolver: new PackageUrlResolver()
-    });
-  }
+  private _editorService: LocalEditorService;
 
   async handleMessage(message: Request): Promise<any> {
+    if (message.kind === 'init') {
+      if (this._editorService) {
```
rictic: We don't, and I agree. I haven't dived into logging at all. Recommend punting this right now.

src/test/editor-service/editor-server-test.ts
rictic: Here and elsewhere. Done. (I think I did it this way because of a clang-format issue that is now resolved. At least, that's my story and I'm sticking to it.)

src/test/editor-service/editor-server-test.ts
rictic: Here and elsewhere: Done

src/test/editor-service/editor-server-test.ts
rictic: Done

src/test/editor-service/editor-server-test.ts
rictic: Oh, right! Done.

rictic:
Description added and all comments addressed. PTAL!

src/test/editor-service/editor-server-test.ts
justinfagnani: nevermind, the callback may be async

src/editor-service/editor-server.ts
FredKSchott: works for me with docs & moving to a demo/ folder

rictic: This file isn't a demo, it's part of the shipped interface. I kinda think we should break the editor stuff out into another repo, but we'll probably punt that for now.

PR #313, opened 2016-09-22
Add stringifier types to shady-css-typings

rictic:
Useful for polyup


rictic:
Yeah, and this should live in shady-css-parser. I pinged Chris about it, I think we'll be landing it soon. That PR is here: https://github.com/PolymerLabs/shady-css-parser/pull/16

PR #314, opened 2016-09-23
Fix #73, by evaluating unary operators correctly.

### LINKED ISSUE #73: element-finder check for 'notify' property breaks on javascript compilation
In the check at https://github.com/PolymerLabs/hydrolysis/blob/master/lib/ast-utils/element-finder.js#L76, `true` and `false` are usually changed to `!0` and `!1` respectively by javascript compilers (Closure Compiler and uglifyjs), which are `UnaryExpression` nodes, and should be allowed.

Also, the `undefined` check is probably unnecessary.
@jklein24

rictic:


src/javascript/ast-value.ts
```diff
@@ -30,12 +30,27 @@ function literalToValue(literal: estree.Literal): LiteralValue {
 /**
  * converts unary to string
  */
-function unaryToValue(unary: estree.UnaryExpression): string {
-  const argValue = expressionToValue(unary.argument);
-  if (argValue === undefined) {
-    return;
+function unaryToValue(unary: estree.UnaryExpression): LiteralValue {
```
justinfagnani: who's calling this function? Is the call site going to be ok with return values changing from `"!0"` to `true`? It doesn't look like the intent of the current version is to actually evaluate the operator.

rictic: The semantics of this family of functions is in fact to early evaluate to a literal value, or return undefined if this is not possible. This function is not called directly but instead is called by expressionToValue, which has these semantics.

justinfagnani: So update the doc comment?

Also, there has to be existing constant evaluation libraries that we can use...

justinfagnani:
LGTM

src/javascript/ast-value.ts
rictic: Good point. Also updated some of the other comments, most of which are of limited usefulness now that we've got good estree typings.

Issue #315, opened 2016-09-25
Support cloning Documents, mutating them, and stringifying the results


rictic:
The tricky part is inline documents. 

 - [x] `node-clone` needs some changes to be able to clone Documents.
   - [x] Clone maps and sets. We can probably get by with just this one for the time being. https://github.com/pvorb/node-clone/pull/64
   - [x] Clone promises. https://github.com/pvorb/node-clone/pull/65
   - [x] Clone symbols. https://github.com/pvorb/node-clone/pull/66
   - [x] Released as 2.0.0
 - [ ] Add AST nodes back onto Feature interface
 - [ ] Move serialization up to Document from ParsedDocument. Basic handling of serializing mutated inline documents.
 - [ ] Add optional indentation arg to serialize method.

rictic:
Marking P0 because it's in the critical path of polymer-upgrade.

PR #316, opened 2016-09-26
Add stringification to ParsedDocument.

rictic:
Doesn't yet handle inline documents. See: https://github.com/Polymer/polymer-analyzer/issues/315


src/css/css-document.ts
```diff
@@ -86,4 +87,13 @@ export class ParsedCssDocument extends ParsedDocument<shady.Node, Visitor> {
   sourceRangeForNode(_node: shady.Node): SourceRange {
     throw new Error('Not implemented');
   }
+
+  stringify() {
+    return cssbeautify(
+               shadyStringifier.stringify(this.ast),
```
justinfagnani: This is my daily venting against clang.

src/html/html-document.ts
```diff
@@ -101,6 +101,22 @@ export class ParsedHtmlDocument extends ParsedDocument<ASTNode, HtmlVisitor> {
       }
     };
   }
+
+  stringify() {
+    // TODO(issues/315): serialize inline documents correctly.
+    let result = parse5.serialize(this.ast);
+
+    // Strip out inferred boilerplate nodes that are injected.
```
justinfagnani: hmm... they're not always injected. What about checking the original source to see if they are there?

src/json/json-document.ts
```diff
@@ -54,4 +54,8 @@ export class ParsedJsonDocument extends ParsedDocument<Json, Visitor> {
   sourceRangeForNode(_node: Json): SourceRange {
     throw new Error('Not Implemented.');
   }
+
+  stringify() {
+    return JSON.stringify(this.ast, null, 2);
```
justinfagnani: FYI, I found this JSON parser that can report source positions: https://github.com/dscape/clarinet

Would need some work to use...

src/html/html-document.ts
rictic: I do. L112: `!this.contents.includes(`<${tagName}`)`

justinfagnani: oops :/

justinfagnani: alternatively, can you check the `location`?

src/json/json-document.ts
rictic: +1 to this or something similar. Writing a JSON parser is really easy if it comes to that.

Looks like what we'd do here is query the state of the input stream as we got events and either have a WeakMap alongside (if we only care about objects and arrays) or build up our own kind of annotated-node based tree?

justinfagnani: Yeah, something like that. I didn't look very hard, so there might be a non-SAX parser that does this too. JSON isn't so important in the scheme of things though.

PR #317, opened 2016-09-26
await on `warningPrinter.printWarnings()` in demo

FredKSchott:
Missing `await` statement led to process exiting before errors get printed.

/cc @rictic


rictic:
Oh, good catch!

FredKSchott:
@rictic It's a shame that the text of the source range has to be re-loaded at print-time (making printing async). What do you think about attaching the line to the warning at creation time? At that point we're likely to have the file already in memory

rictic:
It's a good point. OTOH many times you'll never need that info, either because you're not interested in warnings for a given file or because you're not displaying them in such a way that you need to get the lines.

I'm on the fence. What do you think @justinfagnani ?

PR #318, opened 2016-09-26
Update vscode settings to be shared

FredKSchott:
Two small changes I made to get vscode set up on my machine:

- I sometimes need to dig into the lib folder to see the code that is being executed. Excluding the folder from the tree feels a lot like personal preference.
- added `"typescript.tsdk"` to always match the same version of typescript we use to build, lint, etc.

Is there a way to add personal, project-scoped preferences to vscode? You should still be able to exclude the lib folder if you personally would like to.

/cc @rictic @justinfagnani

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -3,8 +3,8 @@
   "clang-format.style": "file",
   "clang-format.formatOnSave": true,
   "search.exclude": {
-    "node_modules/**": true,
-    "lib/**": true
+    "**/node_modules": true,
+    "**/lib": true
   },
   "json.schemas": [{
     "fileMatch": [
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -2,13 +2,11 @@
 {
   "clang-format.style": "file",
   "clang-format.formatOnSave": true,
-  "files.exclude": {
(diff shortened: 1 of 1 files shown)


.vscode/settings.json
```diff
@@ -2,13 +2,11 @@
 {
   "clang-format.style": "file",
   "clang-format.formatOnSave": true,
```
rictic: fair enough on excluding them entirely. what about just excluding them from search?

.vscode/settings.json
```diff
@@ -2,13 +2,11 @@
...
-    "lib/**": true
-  },
   "json.schemas": [{
     "fileMatch": [
       "elements.json"
     ],
     "url": "./lib/analysis.schema.json"
-  }]
+  }],
+  "typescript.tsdk": "node_modules/typescript/lib"
```
rictic: +1

rictic:
I dunno about multiple levels of package settings, but I think we can come to something agreeable.

.vscode/settings.json
FredKSchott: That definitely sounds fair, it looks like the `"search.exclude"` rule exists for this.

PR #319, opened 2016-09-26
Rootless

justinfagnani:
Remove the concept of root documents. Cache analyzed document. This requires that documents import all of their dependencies. It increases performance of repeated analysis in the benchmark by 2-3x.

<script> tags need special handling, since we want them to see dependencies imported by their container. We support this by adding a document feature pointing to its container document.

Other changes:
 * Added HtmlScriptTagImport and ScannedHtmlScriptTagImport
 * Renamed InlineParsedDocument to ScannedInlineDocument
 * Added InlineDocument
 * Added `deep` option to getFeatures()
 * Imports no longer automatically add their document as a feature
 * Added analyzed document cache to analyzer, and getDocument()

### DIFF
--- a/src/analyzer.ts
+++ b/src/analyzer.ts
@@ -23,7 +23,7 @@ import {HtmlScriptScanner} from './html/html-script-scanner';
 import {HtmlStyleScanner} from './html/html-style-scanner';
 import {JavaScriptParser} from './javascript/javascript-parser';
 import {JsonParser} from './json/json-parser';
-import {correctSourceRange, Document, InlineParsedDocument, LocationOffset, ScannedDocument, ScannedElement, ScannedFeature, ScannedImport} from './model/model';
+import {correctSourceRange, Document, LocationOffset, ScannedDocument, ScannedElement, ScannedFeature, ScannedImport, ScannedInlineDocument} from './model/model';
 import {ParsedDocument} from './parser/document';
 import {Parser} from './parser/parser';
 import {Measurement, TelemetryTracker} from './perf/telemetry';
@@ -82,9 +82,16 @@ export class Analyzer {
 
   private _loader: UrlLoader;
   private _resolver: UrlResolver|undefined;
-  private _parsedDocuments =
+
+  private _parsedDocumentPromises =
       new Map<string, Promise<ParsedDocument<any, any>>>();
-  private _scannedDocuments = new Map<string, Promise<ScannedDocument>>();
+  private _scannedDocumentPromises =
+      new Map<string, Promise<ScannedDocument>>();
+  private _analyzedDocumentPromises = new Map<string, Promise<Document>>();
+
+  private _scannedDocuments = new Map<string, ScannedDocument>();
+  private _analyzedDocuments = new Map<string, Document>();
+
   private _telemetryTracker = new TelemetryTracker();
 
   constructor(options: Options) {
@@ -116,16 +123,65 @@ export class Analyzer {
     // re-analyze a particular file. Editors can use a UrlLoader that reads from
     // it's internal buffers.
     if (contents != null) {
+      this._scannedDocumentPromises.delete(resolvedUrl);
       this._scannedDocuments.delete(resolvedUrl);
-      this._parsedDocuments.delete(resolvedUrl);
+      this._parsedDocumentPromises.delete(resolvedUrl);
+      this._analyzedDocuments.delete(resolvedUrl);
+      this._analyzedDocumentPromises.delete(resolvedUrl);
     }
 
-    const scannedDocument = await this._scan(resolvedUrl, contents);
-    const doneTiming =
-        this._telemetryTracker.start('Document.makeRootDocument', url);
-    const document = Document.makeRootDocument(scannedDocument);
-    doneTiming();
+    const cachedResult = this._analyzedDocumentPromises.get(resolvedUrl);
+    if (cachedResult) {
+      return cachedResult;
+    }
+    // TODO(justinfagnani): also check _analyzedDocuments?
+
+    const promise = (async() => {
+      // Make sure we wait and return a Promise before doing any work, so that
+      // the Promise is cached before anything else happens.
+      await Promise.resolve();
+      const doneTiming = this._telemetryTracker.start('analyze: make document', url);
+      const scannedDocument = await this._scan(resolvedUrl, contents);
+      const document = this._makeDocument(scannedDocument);
+      doneTiming();
+      return document;
+    })();
+    this._analyzedDocumentPromises.set(resolvedUrl, promise);
+    return promise;
+  }
+
+  /**
+   * Constructs a new analyzed Document and adds it to the analyzed Document
+   * cache.
+   */
+  private _makeDocument(scannedDocument: ScannedDocument): Document {
+    if (this._analyzedDocuments.has(scannedDocument.url)) {
+      throw new Error(
+          `Internal error: document ${scannedDocument.url} already exists`);
+    }
+    const document = new Document(scannedDocument, this);
+    document.resolve();
+    // TODO(justinfagnani): should this add to _analyzedDocumentPromises?
+    this._analyzedDocuments.set(scannedDocument.url, document);
     return document;
+  }
+
+  /**
+   * Gets an analyzed Document from the document cache. This is only useful for
+   * Analyzer plugins. You almost certainly want to use `analyze()` instead.
+   *
+   * If a document has been analyzed, it returns the analyzed Document. If not
+   * the scanned document cache is used and a new analyzed Document is returned.
+   * If a file is in neither cache, it returns `null`.
+   */
+  _getDocument(url: string): Document {
+    const resolvedUrl = this._resolveUrl(url);
+    let document = this._analyzedDocuments.get(resolvedUrl);
+    if (document) {
+      return document;
+    }
+    const scannedDocument = this._scannedDocuments.get(resolvedUrl);
+    return scannedDocument && this._makeDocument(scannedDocument);
   }
 
   async getTelemetryMeasurements(): Promise<Measurement[]> {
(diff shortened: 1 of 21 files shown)


src/polymer/element-descriptor.ts
```diff
@@ -197,8 +197,10 @@ function _getFlattenedAndResolvedBehaviors(
     const behavior = document.getOnlyAtId('behavior', behaviorName);
     if (!behavior) {
       throw new Error(
-          `Unable to resolve behavior \`${behaviorName}\` ` +
-          `Did you import it? Is it annotated with @polymerBehavior?`);
+          `In ${document &&
```
rictic: Recommend reverting this for now, @FredKSchott is working on a more proper Warning based error here.

src/test/analyzer_test.ts
```diff
@@ -47,16 +48,155 @@ suite('Analyzer', () => {
...
 
-
   suite('analyze()', () => {
 
+    test(
+        'analyzes a document with an inline Polymer element feature',
+        async() => {
+          const document = await analyzer.analyze(
+              'static/analysis/simple/simple-element.html');
+          const elements = document.getByKind('element');
```
rictic: This will give you *much* better error messages if this test ever breaks:

```typescript
const elements = Array.from(document.getByKind('element'));
assert.deepEqual(elements.map(e => e.tagName), ['simple-element']);
```

Asserting on size and then on individual items means that if there are e.g. two elements that are returned, all you find out is that there are two, not _which_ two. The first thing you'll want to do then is to go and add some console.logs or whatever to tell you that info.

With the above construct you'll get a nice diff like:

```diff
    [
      'simple-element',
+     'other-element',
    ]
```

src/test/analyzer_test.ts
```diff
@@ -47,16 +48,155 @@ suite('Analyzer', () => {
...
+          const subBehavior =
+              scriptDocument.getOnlyAtId('behavior', 'TestBehavior');
+          assert.equal(subBehavior.className, 'TestBehavior');
+        });
+
+
+    // This test is nearly identical to the previous, but covers a different
+    // issue.
+    // The Polymer element-descriptor must find behaviors while resolving,
+    // and if inline document's don't add a document feature for their container
```
rictic: Nit: `if inline document's` -> `if inline documents`

src/test/analyzer_test.ts
```diff
@@ -47,16 +48,155 @@ suite('Analyzer', () => {
...
+          assert.equal(inlineDocuments.length, 1);
+
+          // This is the main purpose of the test: get a feature from the inline
+          // document that's imported by the container document
+          const behaviorJsDocument = inlineDocuments[0];
+          const subBehavior = behaviorJsDocument.getOnlyAtId(
+              'behavior', 'MyNamespace.SubBehavior');
+          assert.equal(subBehavior.className, 'MyNamespace.SubBehavior');
+        });
+
```
rictic: Consider adding a TODO such that there is a way to resolve a js file directly and still get the context of the html file that references it.

src/perf/parse-all-benchmark.ts
```diff
@@ -67,7 +67,7 @@ async function measure() {
   const start = now();
   let document: any;
   for (let i = 0; i < 10; i++) {
-    document = await analyzer.analyzeRoot('ephemeral.html', fakeFileContents);
+    document = await analyzer.analyze('ephemeral.html', fakeFileContents);
```
rictic: Double check that this benchmark still, the output didn't look right to me

rictic:
Looks like the polymer-lint demo may need to be updated as well.

src/analyzer.ts
```diff
@@ -116,18 +123,66 @@ export class Analyzer {
...
+    // TODO(justinfagnani): also check _analyzedDocuments?
+
+    const promise = (async() => {
+      // Make sure we wait and return a Promise before doing any work, so that
+      // the Promise is cached before anything else happens.
+      await Promise.resolve();
+      const scannedDocument = await this._scan(resolvedUrl, contents);
+      return this._makeDocument(scannedDocument);
+    })();
+    this._analyzedDocumentPromises.set(resolvedUrl, promise);
```
rictic: Yeah ideally imports would be able to influence the scanned document that's produced..

Hm.

src/model/document.ts
```diff
@@ -56,110 +54,96 @@ export class ScannedDocument {
...
+    return this.parsedDocument.type;
+  }
+
+  /**
+   * Resolves all features of this document, so that they have references to all
+   * their dependencies.
+   *
+   * This method can only be called once
+   */
+  resolve() {
```
justinfagnani: It's called outside of this class now. I'd like to make ScannedDoument a resolvable and do this work in its `resolve()`

src/analyzer.ts
```diff
@@ -116,18 +123,66 @@ export class Analyzer {
...
   }
 
+  /**
+   * Gets an analyzed Document from the document cache.
+   *
+   * If a document has been analyzed, it returns the analyzed Document. If not
+   * the scanned document cache is used and a new analyzed Document is returned.
+   * If a file is in neither cache, it returns `null`.
+   */
+  getDocument(url: string): Document {
```
rictic: Can we make it available to Import.resolve somehow without exposing it to general users? I dislike having two APIs for the same thing, one that always works and the other than has caveats.

At the least I think we should add docs to this method that say "You almost certainly don't want to use this and should use `resolve` instead."

src/model/document.ts
```diff
@@ -233,45 +217,66 @@ export class Document implements Feature {
...
     }
-    documentsWalked.add(this);
+    visited.add(this);
     for (const feature of this._localFeatures) {
-      inProgress.add(feature);
-      if (feature instanceof Document) {
-        feature._getFeatures(inProgress, documentsWalked);
+      result.add(feature);
+      if (deep) {
+        if (feature.kinds.has('document')) {
```
justinfagnani: I intentionally moved to kinds because we have this mix of instanceof and checking kinds. This allows for other implementations of feature interfaces.

src/html/html-script-tag.ts
```diff
@@ -0,0 +1,64 @@
...
+    // analysis of the external script, but the document the analyzer has
+    // doesn't have its container as a feature.
+    // A better design might be to have the import itself be in charge of
+    // producing document objects. This will fit better with JS modules, where
+    // the type attribute drives how the document is parsed.
+
+    if (this.scannedDocument) {
+      const importedDocument =
+          new Document(this.scannedDocument, document.analyzer);
+      importedDocument._addFeature(document);
```
rictic: Let's be sure of this before landing this PR

src/test/analyzer_test.ts
justinfagnani: great suggestion. done

src/test/analyzer_test.ts
justinfagnani: That's really tricky. I think it maybe falls either under a "find references to..." feature on analyzer. Find the import, the import has the document with the right context.

src/test/analyzer_test.ts
justinfagnani: done

src/html/html-script-tag.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  resolve(document: Document): HtmlScriptTagImport {
+    // TODO(justinfagnani): warn if the same URL is loaded from more than one
+    // non-module script tag
+
+    // TODO(justinfagnani): Use the analyzer cache, since this is duplicating an
+    // analysis of the external script, but the document the analyzer has
+    // doesn't have its container as a feature.
+    // A better design might be to have the import itself be in charge of
+    // producing document objects. This will fit better with JS modules, where
```
rictic: Yeah, this is heading in the right direction.

src/model/document.ts
rictic: Yeah, but we're using this to access a private API here. I can see the argument for `.has('import')` but it's a bit more fuzzy with something you expect to be type compatible with `this` because it's ok for you to access private members of your own class.

It still might be fine for the really creative plugin writer, but they're also likely to break on minor releases.

src/model/document.ts
```diff
@@ -170,15 +154,15 @@ export class Document implements Feature {
...
+    // if (this._featuresByKind) {
+    //   // We have a fast index! Use that.
+    //   return this._featuresByKind.get(kind) || new Set();
+    // } else if (this._doneResolving) {
+    //   // We're done discovering features in this document and its children so
+    //   // we can safely build up the indexes.
+    //   this._buildIndexes();
+    //   return this.getByKind(kind);
+    // }
     return this._getByKind(kind, new Set());
```
rictic: [This is fixed now]

src/polymer/element-descriptor.ts
rictic: Acknowledged.

src/test/analyzer_test.ts
rictic: I agree that it's tricky, but I think supporting it one way or another is needed. Design TBD, but I think a TODO documenting the intention would be nice here.

rictic:
Nice! Looks like making the `resolve` stage cachable improved incremental analysis of large projects by 2-3x (~60ms to ~20ms).

 * Before: https://travis-ci.org/Polymer/polymer-analyzer/jobs/162923992
 * After: https://travis-ci.org/Polymer/polymer-analyzer/jobs/163263123

src/model/document.ts
rictic: Add a TODO then to move this logic to ScannedDocument?

rictic:
As discussed, html referencing script tags is still awkward, but this is a big improvement and looks ready to land. I added a couple of comments requesting TODOs but otherwise this LGTM.

src/model/document.ts
justinfagnani: done

src/test/analyzer_test.ts
justinfagnani: I'll file an issue

PR #322, opened 2016-09-28
Add ast node back on ScannedFeature and Feature

rictic:
Useful in the bundler, as well as in the upgrader.

I initially tried to track the type of ast nodes by a type parameter, but it was more viral than I remember it being when we removed the property last.

I'm not super opinionated about the decision, but it felt kinda ugh if `Document` and the `Feature`s it returns require type parameters.

On the other hand, we could make `Document` extend `Feature<null>` or `Feature<any>` and it would be somewhat rare for a user to spend too much time with undifferentiated `Feature`s.

Hm. WDYT?

Motivated by https://github.com/Polymer/polymer-analyzer/issues/315


rictic:
This wasn't quite right, I'm taking another shot at this, PR out shortly.

PR #320, opened 2016-09-26
Small improvement to "unable to find behavior" error message

FredKSchott:
I ran into this while tracking down a user issue:

```
$ polymer build
info:    Building application...
info:    Generating build/unbundled...
info:    Generating build/bundled...
error:   Promise rejection: Error: Unable to resolve behavior `Polymer.GDGSpainAnimatePage` Did you import it? Is it annotated with @polymerBehavior?
```

I had to search the project myself to find where this behavior was referenced, and it required a bit of trial and error to fix. This change adds the document URL to the error log, to make it easier to find and fix.

/cc @rictic @justinfagnani

PR #326, opened 2016-09-28
Put all responsibility of location offsets on ParsedDocument

rictic:
Should make #320 much more feasible. `ParsedDocument.getSourceRange` should now just Do The Right Thing in all circumstances.

Parsed Documents are now aware of their location offsets from the moment that they're created, and all implementations of `getSourceRange` now automatically apply the appropriate location offsets (if any).

### DIFF
--- a/src/analyzer.ts
+++ b/src/analyzer.ts
@@ -140,7 +140,8 @@ export class Analyzer {
       // Make sure we wait and return a Promise before doing any work, so that
       // the Promise is cached before anything else happens.
       await Promise.resolve();
-      const doneTiming = this._telemetryTracker.start('analyze: make document', url);
+      const doneTiming =
+          this._telemetryTracker.start('analyze: make document', url);
       const scannedDocument = await this._scan(resolvedUrl, contents);
       const document = this._makeDocument(scannedDocument);
       doneTiming();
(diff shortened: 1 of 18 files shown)


src/analyzer.ts
```diff
@@ -378,22 +369,24 @@ export class Analyzer {
...
     })();
     this._parsedDocumentPromises.set(resolvedUrl, promise);
     return promise;
   }
 
-  private _parseContents(type: string, contents: string, url: string):
-      ParsedDocument<any, any> {
+  private _parseContents(
+      type: string, contents: string, url: string,
+      locationOffset: LocationOffset|null): ParsedDocument<any, any> {
```
justinfagnani: why not make `locationOffset` optional?

src/model/document.ts
```diff
@@ -271,8 +269,8 @@ export class Document implements Feature {
   }
 
   /**
-   * Adds and indexes a feature to this document. Thie method can only be
-   * called before resolve().
+   * Adds and indexes a fea
+        ure to this documentled before resolve().
```
justinfagnani: wat?

src/analyzer.ts
rictic: rgr

PR #327, opened 2016-09-28
Get document stringification working properly for mutated inline documents

### LINKED ISSUE #315: Support cloning Documents, mutating them, and stringifying the results

rictic:
Fixes #315 

Also adds astNode back on Feature and ScannedFeature. This unblocks polymer-upgrade.

### DIFF
--- a/src/css/css-document.ts
+++ b/src/css/css-document.ts
@@ -90,15 +90,16 @@ export class ParsedCssDocument extends ParsedDocument<shady.Node, Visitor> {
 
   stringify(options?: StringifyOptions) {
     options = options || {};
-    const beautifulResults =
-        cssbeautify(
-            shadyStringifier.stringify(this.ast),
-            {indent: '  ', autosemicolon: true, openbrace: 'end-of-line'}) +
-        '\n';
+    const beautifulResults = cssbeautify(
+        shadyStringifier.stringify(this.ast),
+        {indent: '  ', autosemicolon: true, openbrace: 'end-of-line'});
+
+    const indent = '  '.repeat(options.indent || 0);
 
     return beautifulResults.split('\n')
-        .map(line => line && '  '.repeat(options.indent || 0) + line)
-        .join('\n');
+               .map(line => line === '' ? '' : indent + line)
+               .join('\n') +
+        '\n';
   }
 }
 
--- a/src/css/css-document.ts
+++ b/src/css/css-document.ts
@@ -16,7 +16,7 @@ import * as cssbeautify from 'cssbeautify';
 import * as shady from 'shady-css-parser';
 
 import {SourceRange} from '../model/model';
-import {Options, ParsedDocument} from '../parser/document';
+import {Options, ParsedDocument, StringifyOptions} from '../parser/document';
 
 export interface Visitor { visit(node: shady.Node, path: shady.Node[]): void; }
 
@@ -88,11 +88,17 @@ export class ParsedCssDocument extends ParsedDocument<shady.Node, Visitor> {
     throw new Error('Not implemented');
   }
 
-  stringify() {
-    return cssbeautify(
-               shadyStringifier.stringify(this.ast),
-               {indent: '  ', autosemicolon: true, openbrace: 'end-of-line'}) +
+  stringify(options?: StringifyOptions) {
+    options = options || {};
+    const beautifulResults =
+        cssbeautify(
+            shadyStringifier.stringify(this.ast),
+            {indent: '  ', autosemicolon: true, openbrace: 'end-of-line'}) +
         '\n';
+
+    return beautifulResults.split('\n')
+        .map(line => line && '  '.repeat(options.indent || 0) + line)
+        .join('\n');
   }
 }
 
--- a/src/html/html-document.ts
+++ b/src/html/html-document.ts
@@ -115,8 +115,12 @@ export class ParsedHtmlDocument extends ParsedDocument<ASTNode, HtmlVisitor> {
     const self = mutableDocuments.shift();
 
     for (const doc of mutableDocuments) {
+      // TODO(rictic): infer this from doc.astNode's indentation.
+      const expectedIndentation = 2;
+
       dom5.setTextContent(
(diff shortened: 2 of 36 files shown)


src/css/css-document.ts
```diff
@@ -88,11 +88,17 @@ export class ParsedCssDocument extends ParsedDocument<shady.Node, Visitor> {
...
+  stringify(options?: StringifyOptions) {
+    options = options || {};
+    const beautifulResults =
+        cssbeautify(
+            shadyStringifier.stringify(this.ast),
+            {indent: '  ', autosemicolon: true, openbrace: 'end-of-line'}) +
         '\n';
+
+    return beautifulResults.split('\n')
+        .map(line => line && '  '.repeat(options.indent || 0) + line)
```
justinfagnani: can you store the indent string in a variable?

justinfagnani: when is `line` null? Does split() do that?

src/css/css-document.ts
```diff
@@ -88,11 +88,17 @@ export class ParsedCssDocument extends ParsedDocument<shady.Node, Visitor> {
...
-  stringify() {
-    return cssbeautify(
-               shadyStringifier.stringify(this.ast),
-               {indent: '  ', autosemicolon: true, openbrace: 'end-of-line'}) +
+  stringify(options?: StringifyOptions) {
+    options = options || {};
+    const beautifulResults =
+        cssbeautify(
+            shadyStringifier.stringify(this.ast),
+            {indent: '  ', autosemicolon: true, openbrace: 'end-of-line'}) +
```
justinfagnani: I assume cssbeautify has no option to indent an initial amount?

src/css/css-document.ts
```diff
@@ -88,11 +88,17 @@ export class ParsedCssDocument extends ParsedDocument<shady.Node, Visitor> {
...
         '\n';
```
justinfagnani: might be a little cleaner to at the final newline after the join?

src/html/html-document.ts
```diff
@@ -102,21 +103,42 @@ export class ParsedHtmlDocument extends ParsedDocument<ASTNode, HtmlVisitor> {
...
-        result =
-            result.replace(RegExp(`<${tagName}>([^]*)?</${tagName}>`), '$1');
-      }
+    for (const doc of mutableDocuments) {
+      // TODO(rictic): infer this from doc.astNode's indentation.
+      const expectedIndentation = 2;
+
+      dom5.setTextContent(
+          doc.astNode,
+          '\n' + doc.stringify({indent: expectedIndentation}) + '  '.repeat(1));
```
justinfagnani: TODO: infer the newline too?

justinfagnani: `'  '.repeat(1)`?

src/html/html-document.ts
```diff
@@ -102,21 +103,42 @@ export class ParsedHtmlDocument extends ParsedDocument<ASTNode, HtmlVisitor> {
     };
   }
 
-  stringify() {
-    // TODO(issues/315): serialize inline documents correctly.
-    let result = parse5.serialize(this.ast);
+  stringify(options?: StringifyOptions) {
+    options = options || {};
+    // We want to make of copy of `this` and all of the inline documents such
```
justinfagnani: can you add why you need to clone? something about inline documents?

src/html/html-document.ts
```diff
@@ -102,21 +103,42 @@ export class ParsedHtmlDocument extends ParsedDocument<ASTNode, HtmlVisitor> {
...
+  }
+}
+
+function prettyPrint(ast: dom5.Node, contents: string) {
+  let result = parse5.serialize(ast);
+
+  // Strip out inferred boilerplate nodes that are injected.
+  const injectedTagNames = ['html', 'head', 'body'];
+  for (const tagName of injectedTagNames) {
+    if (!contents.includes(`<${tagName}`)) {
```
justinfagnani: I can already imaging funky things with import docs that contain strings or comments with `<body>`. Valdrin was showing me a bug related to that the other day. Can you check the source range of the ast nodes first, then remove html/head/body when they don't have a source range, and serialize?

src/css/css-document.ts
rictic: It does not :/

src/css/css-document.ts
rictic: Done.

src/css/css-document.ts
rictic: Done

rictic: It's when it's the empty string, we don't want to splat out a bunch of trailing whitespace.

src/html/html-document.ts
rictic: Yeah, this is the craziest trick I've pulled in a little while. Expanded this comment out a bunch.

src/html/html-document.ts
rictic: I guess the thinking is that we only need newlines if the total content of this line is >80cols or whatever?

That's way more clever than I think we're likely to ever make this method, IMO.

rictic: Fixed `'  '.repeat(1)` to something more sensible.

It made sense at the time, I promise!

src/html/html-document.ts
rictic: Really good point. I forgot that parser-hallucinated nodes wouldn't have __location info!

Done.

PR #330, opened 2016-09-29
Enable strict null checks.

rictic:
Also, put an array of warnings on all features and scanned features so
that instead of bailing out or throwing we can instead shove a warning
someplace that may be useful in future.

Was otherwise rather straightforward. Caught several bugs. Should have done this a long time
ago I think.

### DIFF
--- a/src/analyzer.ts
+++ b/src/analyzer.ts
@@ -23,7 +23,7 @@ import {HtmlScriptScanner} from './html/html-script-scanner';
 import {HtmlStyleScanner} from './html/html-style-scanner';
 import {JavaScriptParser} from './javascript/javascript-parser';
 import {JsonParser} from './json/json-parser';
-import {correctSourceRange, Document, InlineDocInfo, LocationOffset, ScannedDocument, ScannedElement, ScannedFeature, ScannedImport, ScannedInlineDocument} from './model/model';
+import {Document, InlineDocInfo, LocationOffset, ScannedDocument, ScannedElement, ScannedFeature, ScannedImport, ScannedInlineDocument} from './model/model';
 import {ParsedDocument} from './parser/document';
 import {Parser} from './parser/parser';
 import {Measurement, TelemetryTracker} from './perf/telemetry';
@@ -173,9 +173,9 @@ export class Analyzer {
    *
    * If a document has been analyzed, it returns the analyzed Document. If not
    * the scanned document cache is used and a new analyzed Document is returned.
-   * If a file is in neither cache, it returns `null`.
+   * If a file is in neither cache, it returns `undefined`.
    */
-  _getDocument(url: string): Document {
+  _getDocument(url: string): Document|undefined {
     const resolvedUrl = this._resolveUrl(url);
     let document = this._analyzedDocuments.get(resolvedUrl);
     if (document) {
@@ -240,11 +240,6 @@ export class Analyzer {
   private async _scanDocument(
       document: ParsedDocument<any, any>,
       maybeAttachedComment?: string): Promise<ScannedDocument> {
-    // TODO(rictic): We shouldn't be calling _scanDocument with
-    // null/undefined.
-    if (document == null) {
-      return null;
-    }
     const warnings: Warning[] = [];
     const scannedFeatures = await this._getScannedFeatures(document);
     // If there's an HTML comment that applies to this document then we assume
@@ -269,8 +264,8 @@ export class Analyzer {
           }
         });
 
-    const dependencies =
-        (await Promise.all(scannedSubDocuments)).filter(s => !!s);
+    const dependencies = (await Promise.all(scannedSubDocuments))
+                             .filter(s => !!s) as ScannedDocument[];
 
     const scannedDocument =
         new ScannedDocument(document, dependencies, scannedFeatures, warnings);
@@ -300,10 +295,7 @@ export class Analyzer {
       return scannedDocument;
     } catch (err) {
       if (err instanceof WarningCarryingException) {
-        const e: WarningCarryingException = err;
-        e.warning.sourceRange =
-            correctSourceRange(e.warning.sourceRange, locationOffset);
-        warnings.push(e.warning);
+        warnings.push(err.warning);
         return null;
       }
       throw err;
@@ -372,7 +364,7 @@ export class Analyzer {
 
       const doneTiming = this._telemetryTracker.start('parse', 'resolvedUrl');
       const parsedDoc =
-          this._parseContents(extension, content, resolvedUrl, null);
+          this._parseContents(extension, content, resolvedUrl, undefined);
       doneTiming();
       return parsedDoc;
     })();
@@ -382,7 +374,7 @@ export class Analyzer {
 
   private _parseContents(
       type: string, contents: string, url: string,
-      inlineInfo: InlineDocInfo<any>|null): ParsedDocument<any, any> {
+      inlineInfo?: InlineDocInfo<any>): ParsedDocument<any, any> {
     const parser = this._parsers.get(type);
     if (parser == null) {
       throw new NoKnownParserError(`No parser for for file type ${type}`);
--- a/src/analyzer.ts
+++ b/src/analyzer.ts
@@ -363,8 +363,7 @@ export class Analyzer {
       const extension = path.extname(resolvedUrl).substring(1);
 
       const doneTiming = this._telemetryTracker.start('parse', 'resolvedUrl');
-      const parsedDoc =
-          this._parseContents(extension, content, resolvedUrl, undefined);
+      const parsedDoc = this._parseContents(extension, content, resolvedUrl);
       doneTiming();
       return parsedDoc;
     })();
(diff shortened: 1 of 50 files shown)


src/editor-service/editor-server.ts
```diff
@@ -97,7 +97,7 @@ process.stdin.pipe(split()).on('data', async function(line: string) {
 // node child_process.fork() IPC interface
 process.on('message', async function(request: RequestWrapper) {
   const result = await getSettledValue(request.value);
-  process.send(<ResponseWrapper>{id: request.id, value: result});
+  process.send!(<ResponseWrapper>{id: request.id, value: result});
```
FredKSchott: can you remind me (any anyone else reviewing) what `!` does when used like this?

src/editor-service/local-editor-service.ts
```diff
@@ -80,7 +85,7 @@ export class LocalEditorService extends EditorService {
       let attributes: AttributeCompletion[] = [];
       for (const element of elements) {
         // A map from the inheritedFrom to a sort prefix.
-        let sortPrefixes = new Map<string, string>();
+        let sortPrefixes = new Map<string|undefined|null, string>();
```
FredKSchott: can we get rid of undefined and null as valid keys into this map?

src/generate-elements.ts
```diff
@@ -85,7 +86,7 @@ function serializeElement(
   const events = resolvedElement.events.map(
       e => ({
         name: e.name,
-        description: e.description,
+        description: e.description || '',
```
FredKSchott: can we leave as undefined if no description is found? It would make detection easier and less error prone downstream.

src/html/html-script-scanner.ts
```diff
@@ -43,16 +43,16 @@ export class HtmlScriptScanner implements HtmlScanner {
...
           const importUrl = resolveUrl(document.url, src);
           features.push(new ScannedScriptTagImport(
-              'html-script', importUrl, document.sourceRangeForNode(node),
-              document.sourceRangeForAttribute(node, 'src'), node));
+              'html-script', importUrl, document.sourceRangeForNode(node)!,
+              document.sourceRangeForAttribute(node, 'src')!, node));
         } else {
           const locationOffset = getLocationOffsetOfStartOfTextContent(node);
-          const attachedCommentText = getAttachedCommentText(node);
+          const attachedCommentText = getAttachedCommentText(node) || '';
```
FredKSchott: same here and for more below, if there's no attached comment text can we leave as undefined?

src/javascript/esutil.ts
```diff
@@ -157,12 +165,24 @@ export function toScannedPolymerProperty(
   let description =
       jsdoc.removeLeadingAsterisks(getAttachedComment(node) || '').trim();
 
+  const name = objectKeyToString(node.key);
+  if (name == null) {
```
FredKSchott: this will not be null anymore, use `typeof === 'undefined'` if possible.

src/javascript/javascript-document.ts
```diff
@@ -23,7 +23,7 @@ import {Visitor, VisitResult} from './estree-visitor';
 
 export {Visitor} from './estree-visitor';
 
-let __exampleNode: Node;
+const __exampleNode: Node = <any>null;
```
FredKSchott: can you add a comment explaining this? Curious about its usage here

src/model/element.ts
```diff
@@ -32,13 +33,14 @@ export class ScannedElement implements Resolvable {
   className?: string;
   superClass?: string;
   extends?: string;
-  properties: ScannedProperty[] = [];
-  attributes: ScannedAttribute[] = [];
+  properties = [] as ScannedProperty[];
```
FredKSchott: why the change to the new format?

src/editor-service/editor-server.ts
rictic: It means "I solemnly swear the expression to the left of the `!` is never null or undefined (right here just this one time)"

src/html/html-script-scanner.ts
rictic: see comment above

src/editor-service/local-editor-service.ts
FredKSchott: Got it, that makes sense. Can you document this somewhere?

src/javascript/esutil.ts
FredKSchott: Oh right, I remember we've talked about this before. Using `(name == null)` to check for falseyness is deceptive to me as a reviewer, because it feels like the intention is to check something more specific about the variable & `null` and maybe the `==` was a mistake.  `(!name)` more explicitly checks for falseyness.

Can we switch this to `(!name)` here and going forward?

src/model/element.ts
FredKSchott: I like that you're experimenting! :)

I personally prefer the first form (matches non default property syntax & uses types to describe the property instead of the value being set to that property) but I guess that's subjective. I'm more concerned with moving away from the format we've been using everywhere else so far and would prefer we stick to that.

src/generate-elements.ts
FredKSchott: Got it, thanks for the explanation. I still feel like `undefined` is more true here and will be an easier interface for users to interact with, but I do see the value in simplifying our types and elements.json format. I'll leave the decision up to you :)

src/javascript/esutil.ts
rictic: Here, I think you're totally right, `!name` is better, as an empty property name doesn't make sense as a polymer property.

In general I think we should keep `foo == null` in our repertoire, as it only catches `null` and `undefined` whereas `!foo` also catches `false`, `0`, `''`, `NaN`, (and in the browser, `document.all` because lol), and many times you do want to distinguish between nullish and falsy.

src/model/element.ts
rictic: @justinfagnani care to weigh in on a question of style?

src/javascript/javascript-document.ts
rictic: Added a comment. 

(suddenly I can reply here. still figuring out github's new review ui..)

src/editor-service/editor-server.ts
justinfagnani: It's a cast. If you have `o: T|null`, `o!` is the same as `(o as T)`. If TypeScript had subtraction types, it'd be `(o as (typeof o - null|undefined))`.

src/javascript/esutil.ts
FredKSchott: wait... `false != null`? How did I forget that???

Nevermind then, there's definitely value in distinguishing between the two

src/generate-elements.ts
justinfagnani: Before null-tracking I would say that I prefer non-null fields so that client code doesn't have to branch too many times to deal with null. I still think this is true, especially with collections. If the primary client branching would come from detecting empty though, it might not buy much, and with null-tracking you at least get warnings.

So `null` could make sense here. I'm not so sure about `undefined` as the 'not present' value. Here I'd like to represent that the object has a description property, that may be null, rather than not having a description property at all. Of course, to be maximally confusing, setting a property to `undefined` in JS means that the object has an own property name, but it definitely clouds the issue.

justinfagnani:
Before I get too far in this, what do we think about the default 'no value' value being `null`, not `undefined`?

src/model/element.ts
justinfagnani: I think it's a toss-up here. One is a type annotation to fill in a gap in type inference, the other is a cast, which then makes type inference work. Here, I think I prefer the type annotation so that the list of variables and their types is relatively consistent. It's also good to have a consistent `let or const` `name` `type` `initializer` ordering.

src/generate-elements.ts
rictic: Client code for elements.json is potentially really diverse, it's not just nodejs and not just code that we write, so I wanted to make it as easy to consume as possible. IMO defaulting to a description of empty string is the easiest thing for a user to consume.

Agreed on generally preferring nulls to undefined to represent a 'not present' value on a field.

src/analyzer.ts
```diff
@@ -269,8 +264,8 @@ export class Analyzer {
           }
         });
 
-    const dependencies =
-        (await Promise.all(scannedSubDocuments)).filter(s => !!s);
+    const dependencies = (await Promise.all(scannedSubDocuments))
+                             .filter(s => !!s) as ScannedDocument[];
```
rictic: Not without sacrificing local correctness, because we're only filtering out the nulls here.

src/model/element.ts
```diff
@@ -70,6 +72,7 @@ export class Element implements Feature {
   jsdoc?: jsdoc.Annotation;
   astNode: estree.Node|null;
   kinds: Set<string> = new Set(['element']);
+  warnings: Warning[] = [];
```
FredKSchott: What kind of feature can have warnings now? Does it make sense to add directly to the Feature class? Then all documents and/or elements could to have a `getWarnings()` method that returns all warnings contained within.

WDYT? I've been thinking about how to return all warnings found within a file, which is why this is top of mind.

rictic: Feature is just an interface so we need to add this onto all implementing classes, but yes. This PR adds an array of warnings onto ~everything, for getWarnings to gather up

rictic:
Rebased and comments addressed, PTAL

justinfagnani:
LGTM

PR #331, opened 2016-09-29
"behavior not found" errors are now WarningCarryingExceptions

FredKSchott:
This shouldn't conflict with #330 but an additional PR will be needed so that we can push this warning onto the document itself, instead of throwing a `WarningCarryingException`.

/cc @rictic @justinfagnani

### DIFF
--- a/src/model/element.ts
+++ b/src/model/element.ts
@@ -20,6 +20,22 @@ import {Document, Event, Feature, Property, Resolvable, ScannedEvent, ScannedPro
 
 export {Visitor} from '../javascript/estree-visitor';
 
+/**
+ * A scanned behavior assignment of a Polymer element. This is only a
+ * reference to the behavior and not the actual behavior definition itself.
+ *
+ * ex:
+ *   Polymer({
+ *     is: 'custom-element',
+ *     behaviors: [Polymer.SomeBehavior]
+ *                 ^^^^^^^^^^^^^^^^^^^^
+ *   });
+ */
+export interface ScannedBehaviorAssignment {
+  name: string;
+  sourceRange: SourceRange;
+}
+
 export interface ScannedAttribute {
   name: string;
   sourceRange: SourceRange;
--- a/src/model/element.ts
+++ b/src/model/element.ts
@@ -20,22 +20,6 @@ import {Document, Event, Feature, Property, Resolvable, ScannedEvent, ScannedPro
 
 export {Visitor} from '../javascript/estree-visitor';
 
-/**
- * A scanned behavior assignment of a Polymer element. This is only a
- * reference to the behavior and not the actual behavior definition itself.
- *
- * ex:
- *   Polymer({
- *     is: 'custom-element',
- *     behaviors: [Polymer.SomeBehavior]
- *                 ^^^^^^^^^^^^^^^^^^^^
- *   });
- */
-export interface ScannedBehaviorAssignment {
-  name: string;
-  sourceRange: SourceRange;
-}
-
 export interface ScannedAttribute {
   name: string;
   sourceRange: SourceRange;
--- a/src/test/analyzer_test.ts
+++ b/src/test/analyzer_test.ts
@@ -94,8 +94,8 @@ suite('Analyzer', () => {
         assert.instanceOf(err, WarningCarryingException);
         const warning: Warning = err.warning;
         assert.equal(
-            err.message, 'In static/html-missing-behaviors.html: ' +
-                'Unable to resolve behavior `Polymer.ExpectedMissingBehavior`. ' +
+            err.message,
+            'Unable to resolve behavior `Polymer.ExpectedMissingBehavior`. ' +
                 'Did you import it? Is it annotated with @polymerBehavior?');
         assert.deepEqual(warning.sourceRange, {
           file: 'static/html-missing-behaviors.html',
--- a/src/test/analyzer_test.ts
+++ b/src/test/analyzer_test.ts
@@ -435,7 +435,8 @@ suite('Analyzer', () => {
           elements.map(e => e.tagName), ['test-seed', 'test-element']);
       const testSeed = elements[0];
 
-      assert.deepEqual(testSeed.behaviors, ['Behavior1', 'Behavior2']);
+      assert.deepEqual(
+          testSeed.behaviorAssignments, ['Behavior1', 'Behavior2']);
       assert.equal(testSeed.tagName, 'test-seed');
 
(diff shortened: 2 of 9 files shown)


src/model/element.ts
```diff
@@ -20,6 +20,22 @@ import {Document, Event, Feature, Property, Resolvable, ScannedEvent, ScannedPro
...
+ * reference to the behavior and not the actual behavior definition itself.
+ *
+ * ex:
+ *   Polymer({
+ *     is: 'custom-element',
+ *     behaviors: [Polymer.SomeBehavior]
+ *                 ^^^^^^^^^^^^^^^^^^^^
+ *   });
+ */
+export interface ScannedBehaviorAssignment {
```
rictic: Behaviors are polymer specific and should live in `src/polymer`

src/polymer/element-descriptor.ts
```diff
@@ -182,33 +182,36 @@ function resolveElement(
...
     //     could be a mild warning and we could just take the last one in the
     //     array, which should be the most recently defined one.
-    const behavior = document.getOnlyAtId('behavior', behaviorName);
-    if (!behavior) {
-      throw new Error(
-          `In ${document &&
-          document.url}:` +
-              `Unable to resolve behavior \`${behaviorName}\` ` +
-              `Did you import it? Is it annotated with @polymerBehavior?`);
+    const foundBehavior = document.getOnlyAtId('behavior', behavior.name);
```
rictic: What about pushing the warning onto the document (if the urls match) and continuing? This way we don't stop the analysis

```typescript
if (!foundBehavior) {
  if (document.url === behavior.sourceRange.file) {
    document.warnings.push({/* the warning */});
  }
  continue;
}
```

src/polymer/element-descriptor.ts
```diff
@@ -182,33 +182,36 @@ function resolveElement(
...
+    if (!foundBehavior) {
+      throw new WarningCarryingException({
+        message: `In ${document && document.url}: Unable to resolve behavior ` +
```
rictic: Now that this is a Warning and we have a sourceRange the url here is superfluous

src/polymer/element-descriptor.ts
```diff
@@ -182,33 +182,36 @@ function resolveElement(
...
   const resolvedBehaviors = new Set<Behavior>();
   _getFlattenedAndResolvedBehaviors(behaviors, document, resolvedBehaviors);
   return resolvedBehaviors;
 }
 
 function _getFlattenedAndResolvedBehaviors(
-    behaviors: string[], document: Document, resolvedBehaviors: Set<Behavior>) {
-  const toLookup = behaviors.slice();
-  for (let behaviorName of toLookup) {
+    behaviors: ScannedBehaviorAssignment[], document: Document,
```
rictic: this was in the existing code, but rereading it I think that the variable name `behaviors` might be better as something like `behaviorReferences`

src/test/analyzer_test.ts
```diff
@@ -84,6 +85,32 @@ suite('Analyzer', () => {
           ['MyNamespace.SubBehavior', 'MyNamespace.SimpleBehavior']);
     });
 
+    // TODO(fks) 09-28-2016: Fix off-by-one error in warning sourceRange line
+    // numbers.
+    test('throws when cant find behavior', async() => {
```
rictic: I tend towards more integration tests and fewer unit tests myself, so I think this is great. 

It seems like otherwise you'd need to either set up parsing, scanning, resolving, etc yourself, ending up with your own mini-analyzer just for testing, or you'd have to  hand-constructing complex middle-stage objects. Both options seem too unreliable and divergent from the code as it is used.

src/polymer/element-descriptor.ts
FredKSchott: That's the ultimate goal, but I was going to keep this simple and make that change in a later PR that adds a `document.addWarning()` method (right now `_warnings` is a private property).

src/polymer/element-descriptor.ts
FredKSchott: Woo!

src/test/analyzer_test.ts
FredKSchott: 👍 that was my understanding as well but I wanted to double check that I wasn't missing some easier way.

src/polymer/element-descriptor.ts
rictic: hm, consider just dropping the underscore and making `_warnings` public

FredKSchott: Any objection to me doing that in another PR immediately after this one? 

My concern is that that will be far reaching, and I'd rather not mix the two now that this has become more of a "introduce the concept of `ScannedBehaviorAssignment`" PR

FredKSchott:
@rictic PTAL

rictic:
LGTM!

src/polymer/element-descriptor.ts
justinfagnani: I'd like to kill `WarningCarryingException` asap...

src/test/analyzer_test.ts
justinfagnani: I find targeted tests easier to debug when they break, especially if you can get an idea from which tests broke what went wrong. I'd love to test Polymer-specific things in the Polymer-specific tests. We should create some analyzer setup utilities for tests on various plugins. This is fine for now though.

src/test/analyzer_test.ts
```diff
@@ -84,6 +85,32 @@ suite('Analyzer', () => {
...
+        assert.instanceOf(err, WarningCarryingException);
+        const warning: Warning = err.warning;
+        assert.equal(
+            err.message,
+            'Unable to resolve behavior `Polymer.ExpectedMissingBehavior`. ' +
+                'Did you import it? Is it annotated with @polymerBehavior?');
+        assert.deepEqual(warning.sourceRange, {
+          file: 'static/html-missing-behaviors.html',
+          start: {
+            line: 23,  // Actual Expected: 24
```
justinfagnani: what does this `// Actual Expected: 24` comment mean?

FredKSchott:
Thanks everyone! Onwards to killing `WarningCarryingException`...

src/test/analyzer_test.ts
FredKSchott: Our line & position numbers are currently expected to be index-zero. This is weird in the warning use case, since most text editors are index-one.

Started an issue to discuss here: https://github.com/Polymer/polymer-analyzer/issues/332

PR #337, opened 2016-09-30
document SourceRange & Position indexing

### LINKED ISSUE #332: Are line & column numbers 0-indexed? 1-indexed?
This is the type definition for position:

```
export interface Position {
  /** The line number, starting from zero. */
  line: number;
  /** The column offset within the line, starting from zero. */
  column: number;
}
```

However, most (all?) text editors treat line numbers and column numbers as starting from one. This mismatch makes reporting errors difficult, since they will all be off by one (see https://github.com/Polymer/polymer-analyzer/blob/master/src/test/analyzer_test.ts#L103).

I'm opening this issue to discuss. Can we change our internal logic to treat lines and columns as 1-indexed? Or should we account for this as consumers of analyzed documents (ex: handle in `WarningPrinter`)?

/cc @rictic @justinfagnani

FredKSchott:
Resolves #332

/cc @rictic @justinfagnani

### DIFF
--- a/src/model/source-range.ts
+++ b/src/model/source-range.ts
@@ -12,6 +12,12 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
+/**
+ * Describes a range of text within a source file.
+ *
+ * NOTE: See `Position` documentation below for an important information on
+ * how to report `SourceRange` objects to the user.
+ */
 export interface SourceRange {
   /* The resolved path to the file. */
   file: string;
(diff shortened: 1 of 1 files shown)


src/model/source-range.ts
```diff
@@ -12,13 +12,27 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
+/**
+ * Describes a range of text within a source file.
+ *
+ * NOTE: See `Position` documentation below for an important information on
```
rictic: What about something a bit shorter and sweeter. What about:

`
Note: columns and line positions are indexed from zero. Consider displaying them to users as one-indexed to match text editor convention.
`

And then dropping the note from Position.

FredKSchott:
@rictic PTAL

PR #338, opened 2016-09-30
document.getWarnings()

FredKSchott:
`document.getWarnings()` will now return all warnings on both the document and features within the document. It takes an optional `deep` argument which can be used to filter out warnings on non-local features. This behaves similar to `document.getFeatures().`

This grants our editor plugins, build pipeline, and others a single interface for getting warnings from a given document.

/cc @justinfagnani @rictic @usergenic @garlicnation

### DIFF
--- a/src/model/document.ts
+++ b/src/model/document.ts
@@ -290,9 +290,9 @@ export class Document implements Feature {
     if (deep == null) {
       deep = true;
     }
-    warnings.push(...this.warnings);
+    warnings.push.apply(warnings, this.warnings);
     for (const feature of this.getFeatures(deep)) {
-      warnings.push(...feature.warnings);
+      warnings.push.apply(warnings, feature.warnings);
     }
     return warnings;
   }
--- a/src/model/document.ts
+++ b/src/model/document.ts
@@ -146,6 +146,17 @@ export class Document implements Feature {
     this._doneResolving = true;
   }
 
+  /**
+   * Adds and indexes a feature to this documentled before resolve().
+   */
+  _addFeature(feature: Feature) {
+    if (this._doneResolving) {
+      throw new Error('_addFeature can not be called after _resolve()');
+    }
+    this._indexFeature(feature);
+    this._localFeatures.add(feature);
+  }
+
   getByKind(kind: 'element'): Set<Element>;
   getByKind(kind: 'polymer-element'): Set<PolymerElement>;
   getByKind(kind: 'behavior'): Set<Behavior>;
@@ -267,17 +278,6 @@ export class Document implements Feature {
     }
   }
 
-  /**
-   * Adds and indexes a feature to this documentled before resolve().
-   */
-  _addFeature(feature: Feature) {
-    if (this._doneResolving) {
-      throw new Error('_addFeature can not be called after _resolve()');
-    }
-    this._indexFeature(feature);
-    this._localFeatures.add(feature);
-  }
-
   toString(): string {
     return this._toString(new Set()).join('\n');
   }
@@ -297,8 +297,7 @@ export class Document implements Feature {
       } else {
         let subResult = localFeature.toString();
         if (subResult === '[object Object]') {
-          subResult =
-              `<${localFeature.constructor.name} kinds="${Array
+          subResult = `<${localFeature.constructor.name} kinds="${Array
                   .from(localFeature.kinds)
                   .join(', ')}" ids="${Array.from(localFeature.identifiers)
                   .join(',')}">}`;
--- a/src/model/document.ts
+++ b/src/model/document.ts
@@ -247,8 +247,8 @@ export class Document implements Feature {
   }
 
   /**
-   * Get features for this document. If `deep` is true, also get features for
-   * all documents reachable from imports in this document.
+   * Get features for all documents reachable via imports in this document.
+   * If `deep` is false, only return features in this document.
    */
   getFeatures(deep?: boolean): Set<Feature> {
     if (deep == null) {
@@ -279,8 +279,8 @@ export class Document implements Feature {
   }
 
   /**
-   * Get warnings for this document. If `deep` is true, also get warnings for
-   * all documents reachable from imports in this document.
+   * Get warnings for all documents and features reachable via imports in this
+   * document. If `deep` is false, only return warnings in this document.
    */
   getWarnings(deep?: boolean): Warning[] {
     const warnings: Warning[] = [];
--- a/src/model/document.ts
+++ b/src/model/document.ts
@@ -54,7 +54,7 @@ export class Document implements Feature {
   kinds: Set<string> = new Set(['document']);
   identifiers: Set<string> = new Set();
   analyzer: Analyzer;
-  warnings: Warning[] = [];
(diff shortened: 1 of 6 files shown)


src/model/document.ts
```diff
@@ -88,7 +88,7 @@ export class Document implements Feature {
       this.identifiers.add(this.url);
     }
     this.kinds.add(`${this.parsedDocument.type}-document`);
-    // this._resolve();
+    this.warnings = base.warnings.slice();
```
FredKSchott: see the removal of `get warnings()` below. @rictic is it ever possible for warnings to be added to a ScannedDocument after the Document has been created from it?

src/model/document.ts
```diff
@@ -268,14 +280,21 @@ export class Document implements Feature {
   }
 
   /**
-   * Adds and indexes a feature to this documentled before resolve().
+   * Return all warnings on local features contained within the document as well
```
rictic: What about:

```
Get warnings for this document. If `deep` is true, also get warnings for all documents reachable from imports in this document.
```

src/model/document.ts
```diff
@@ -268,14 +280,21 @@ export class Document implements Feature {
...
-    if (this._doneResolving) {
-      throw new Error('_addFeature can not be called after _resolve()');
+  getWarnings(deep?: boolean): Warning[] {
+    const warnings: Warning[] = [];
+    if (deep == null) {
+      deep = true;
     }
-    this._indexFeature(feature);
-    this._localFeatures.add(feature);
+    warnings.push(...this.warnings);
```
FredKSchott: oops, forgot that TypeScript won't compile `...` out. Tests are failing on Travis for Node v4 because of these.

src/model/document.ts
```diff
@@ -268,14 +280,21 @@ export class Document implements Feature {
...
+   * deep argument is set to false, only return warnings found on local
+   * features.
    */
-  _addFeature(feature: Feature) {
-    if (this._doneResolving) {
-      throw new Error('_addFeature can not be called after _resolve()');
+  getWarnings(deep?: boolean): Warning[] {
+    const warnings: Warning[] = [];
+    if (deep == null) {
+      deep = true;
```
rictic: I think `deep` should be false by default for this method.

src/model/document.ts
rictic: I think this will work: `warnings.push.apply(warnings, this.warnings)`

src/polymer/polymer-element.ts
```diff
@@ -221,14 +221,15 @@ function _getFlattenedAndResolvedBehaviors(
     //     array, which should be the most recently defined one.
     const foundBehavior = document.getOnlyAtId('behavior', behavior.name);
     if (!foundBehavior) {
-      throw new WarningCarryingException({
+      document.warnings.push({
```
rictic: Only want to push this here if `document.url === behavior.sourceRange.file` because the behaviorAssignment could be in another file as part of a chain of behaviors. e.g.

`index.html`
```html
<link rel="import" href="foo.html">
<script>
  Polymer({is: 'index-el', behaviors: [Foo]});
</script>
```

`foo.html`
```html
<script>
  /** @polymerBehavior */
  const Foo = {behaviors: [NotFoundBehavior]};
</script>
```

So we don't want to report that NotFoundBehavior wasn't found in index.html, only in foo.html.

src/model/document.ts
rictic: No, a ScannedDocument should be logically immutable by the time that we're creating Documents

src/polymer/polymer-element.ts
rictic: Or we could filter for this at `getWarnings` maybe, but we'd have a problem with duplicate warnings I think.

src/model/document.ts
FredKSchott: I actually originally had that behavior as the default before realizing that all other methods (`getFeatures()`, `getByKind()`, `getById()`) either defaulted to using or only worked on the deep import graph. Having getWarnings() as the only method returning non-deep objects by default might trip users up who expect to be working with the entire import graph. wdyt?

src/polymer/polymer-element.ts
FredKSchott: Good catch, I forgot to come back and fix this. I'll add a test as well.

src/model/document.ts
```diff
@@ -247,9 +247,8 @@ export class Document implements Feature {
   }
 
   /**
-   * Return all local features contained within the document as well as features
-   * detected by following through all imports. If the deep argument is set to
-   * false, only return local features.
+   * Get features for this document. If `deep` is true, also get features for
```
rictic: This might lead people in the wrong direction, as `deep` defaults to true here. 

I wrote the suggestion for getWarnings thinking that we'd go with `deep` defaulting to false there, so maybe we should say something like:

    Get features reachable by this document and its imports. If `deep` is false, only get features contained in this document.

src/model/document.ts
rictic: Good point, let's do the consistent thing.

src/polymer/polymer-element.ts
FredKSchott: Test added, but I just remembered this was already solved for during the scanning phase (by saving the proper sourceRange to be used here) instead of during this detection phase.  So while a warning is generated because the behavior wasn't detected for the `index.html` document, it is generated with the `sourceRange` cached in the `ScannedBehaviorAssignment` pointing to where the behavior assignment was referenced.

rictic: Hm, good point. I think the solution there is for us to also call something like getFlattenedAndResolvedBehaviors as part of resolving Behaviors.

FredKSchott: As discussed in person, leaving this as future work (I'll add a TODO)

FredKSchott:
@justinfagnani PTAL when you can

src/model/document.ts
justinfagnani: arg, I have pending comments I forgot to send, here where I'm favoring non-deep by default. I think this can change with other changes upcoming wrt resolving and querying. My logic here is that you don't expect to get warnings from a dependency when you ask for warnings from a file.

Outside of warnings I think we'll shortly need to refine what the concept of "deep" even is. So far we've been dealing with globally scoped features of documents, either a `customElements.define()` call, or a global symbol. We're going to have to adapt to scoped exports/imports, in which case, what does querying a document mean? Are you getting the exported symbols from a document, or the import symbols visible within a document, or all exported symbols reachable from a node in the dependency graph?

I'd argue that the last query isn't particularly useful most of the time (though it could be for some editor service suggestions), and so what you want by default is a shallow query for what's either exported from or imported into a document. Only getting local warnings would be consistent with this. Globals are kind of like having an `export * from x` for all x that a document imports, and so deep == false would be sort of like exports except re-exports. It could be that deep is to vague of a word for all these concepts.

FredKSchott: coming back to this, I'll change the default behavior to default false for now because that's what I would ultimately expect from the method. But we should figure out our overall position here asap

PR #340, opened 2016-09-30
Fix issue with not properly caching analyzed documents

justinfagnani:


src/analyzer.ts
```diff
@@ -173,14 +172,17 @@ export class Analyzer {
...
    */
   private _makeDocument(scannedDocument: ScannedDocument): Document {
-    if (this._analyzedDocuments.has(scannedDocument.url)) {
+    const resolvedUrl = scannedDocument.url;
+
+    if (this._analyzedDocuments.has(resolvedUrl)) {
       throw new Error(
-          `Internal error: document ${scannedDocument.url} already exists`);
+          `Internal error: document ${resolvedUrl} already exists`);
     }
```
rictic: check and throw for `this._analyzedDocumentPromises.has(resolvedUrl)` as well?

justinfagnani: can't do that, since this is called from `analyze()` which does set a Promise in the cache.

PR #343, opened 2016-10-05
Sync dom5 parse5

usergenic:
Analyzer using dom5 and dom5 using a different version of parse5 than analyzer.  This created weird downstream type issues for clients of analyzer who used dom5 or parse5 as well.

Bumped up analyzer to use dom5 latest and latest parse5 typings.


rictic:
LGTM

PR #344, opened 2016-10-06
Accurate SourceRange production for ParsedHtmlDocument

usergenic:
SourceRange production was inaccurate `<zoidberg>`and you should feel inaccurate`</zoidberg>`.

- Added code to correct SourceRange calculations for multi-line nodes.
- Changed the way element nodes report SourceRanges now to include closing tag when present.
- When closing tag is not present, element nodes now report multi-line SourceRanges correctly.
- Added test file for html-document and source-ranges html fixture.

### DIFF
--- a/src/html/html-document.ts
+++ b/src/html/html-document.ts
@@ -63,6 +63,17 @@ export class ParsedHtmlDocument extends ParsedDocument<ASTNode, HtmlVisitor> {
         end: {line: location.endTag.line - 1, column: location.endTag.col - 1}
       };
     }
+    // text node value can provide accurate line and column count
+    if (parse5.treeAdapters.default.isTextNode(node)) {
+      const lines = (node.value || '').split(/\n/);
+      const lastLine = lines[lines.length - 1];
+      const endColumn = lines.length === 1 ? location.col + lastLine.length - 1 : lastLine.length;
+      return {
+        file: this.url,
+        start: {line: location.line - 1, column: location.col - 1},
+        end: {line: location.line + lines.length - 2, column: endColumn}
+      };
+    }
     return {
       file: this.url,
       // one indexed to zero indexed
--- a/src/html/html-document.ts
+++ b/src/html/html-document.ts
@@ -47,14 +47,37 @@ export class ParsedHtmlDocument extends ParsedDocument<ASTNode, HtmlVisitor> {
     });
   }
 
-  _sourceRangeForNode(node: ASTNode): SourceRange|undefined {
-    if (!node || !node.__location) {
-      return;
-    }
-    // dom5 locations are 1 based but ours are 0 based.
+  // For multi-line comment nodes, we must calculate the ending line and
+  // column ourselves.
+  _sourceRangeForCommentNode(node: ASTNode): SourceRange|undefined {
     const location = node.__location;
 
-    // element node
+    if (!location || isElementLocationInfo(location) ||
+        !parse5.treeAdapters.default.isCommentNode(node)) {
+      return;
+    }
+
+    const lines = (parse5.treeAdapters.default.getCommentNodeContent(node) ||
+        '').split(/\n/);
+    const endLength = lines[lines.length - 1].length;
+
+    // Adjust length for single-line comment by 6 for `<!---->` and adjust by 3
+    // for multi-line comment for `-->` only.
+    const endColumn =
+        lines.length === 1 ? location.col + endLength + 6 : endLength + 3;
+    return {
+      file: this.url,
+      start: {line: location.line - 1, column: location.col - 1},
+      end: {line: location.line + lines.length - 2, column: endColumn}
+    };
+  }
+
+  // An element node with end tag information will produce a source range that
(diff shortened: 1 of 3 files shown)


src/html/html-document.ts
```diff
@@ -63,6 +63,17 @@ export class ParsedHtmlDocument extends ParsedDocument<ASTNode, HtmlVisitor> {
         end: {line: location.endTag.line - 1, column: location.endTag.col - 1}
       };
     }
+    // text node value can provide accurate line and column count
```
rictic: This comment wasn't helpful until after I'd already read and understood the code. Is this accurate?

`// We need to determine the ending line and column ourselves for other kinds of nodes. This is tricky and we probably don't do it right for all cases yet, but for text nodes at least we can just examine the node's content.`

src/test/html/html-document_test.ts
```diff
@@ -0,0 +1,66 @@
...
+    const file = fs.readFileSync(
+        path.resolve(__dirname, `../${url}`), 'utf8');
+    const document: ParsedHtmlDocument = parser.parse(file, url);
+
+    test.skip('can report correct range for comments', () => {});
+
+    test.skip('can report correct range for elements', () => {
+      // TODO(usergenic): If element source range includes their children,
+      // shouldn't they extend to include their closing tag as well?  They
+      // don't do so currently.  Add test after we discuss/and/or fix.
```
rictic: Makes sense that if you want the entire element to be squiggled then that should include the end tag if any.

src/test/html/html-document_test.ts
```diff
@@ -0,0 +1,66 @@
...
+      assert.deepEqual(document.sourceRangeForNode(pTags[0]!.childNodes![0]!), {
+        file: url, start: {line: 13, column: 7}, end: {line: 15, column: 4}
+      });
+
+      // The second <p> tag text node child is single line, terminated by a
+      // closing tag.
+      assert.deepEqual(document.sourceRangeForNode(pTags[1]!.childNodes![0]!), {
+        file: url, start: {line: 15, column: 7}, end: {line: 15, column: 46}
+      });
+    });
```
rictic: Nice tests!

src/test/html/html-document_test.ts
usergenic: Ok, will fix.

src/html/html-document.ts
usergenic: Will update.  I also just committed fix for comment nodes.  About to send one for elements with comment updates.

usergenic:
@rictic, if you want to take another look I committed a complete rewrite of the _sourceRangeForNode that handles comments and fixes multi-line elements (without end tags) and elements with end tags to include the end tag.

src/html/html-document.ts
```diff
@@ -63,17 +65,35 @@ export class ParsedHtmlDocument extends ParsedDocument<ASTNode, HtmlVisitor> {
...
+        end: {line: location.line + lines.length - 2, column: endColumn}
+      };
+    }
+
+    // comment node
+    if (parse5.treeAdapters.default.isCommentNode(node)) {
+      const lines = (parse5.treeAdapters.default.getCommentNodeContent(node) ||
+          '').split(/\n/);
+      const lastLineLength = lines[lines.length - 1].length;
+      const endColumn = lines.length === 1 ? location.col + lastLineLength + 6
```
rictic: why +6 vs +3?

src/test/html/html-document_test.ts
```diff
@@ -28,7 +29,21 @@ suite('ParsedHtmlDocument', () => {
         path.resolve(__dirname, `../${url}`), 'utf8');
     const document: ParsedHtmlDocument = parser.parse(file, url);
 
-    test.skip('can report correct range for comments', () => {});
+    test('can report correct range for comments', () => {
+      const comments = dom5.nodeWalkAll(document.ast,
+          parse5.treeAdapters.default.isCommentNode);
```
rictic: Can simplify later lines by adding the non-null casting `!` here

src/html/html-document.ts
```diff
@@ -47,61 +47,118 @@ export class ParsedHtmlDocument extends ParsedDocument<ASTNode, HtmlVisitor> {
...
+      if (lastRange) {
+        return {
+          file: this.url,
+          start: {line: location.line - 1, column: location.col - 1},
+          end: lastRange.end
+        };
+      }
+    }
+
+    // Fallback to determining the node's dimensions by serializing it.
```
rictic: clever.

src/html/html-document.ts
usergenic: The +6 vs +3 is described in the comment for new _sourceRangeForCommentNode header here https://github.com/Polymer/polymer-analyzer/pull/344/files#diff-d9049ec942cc8c27254037cf14faa624R64

src/html/html-document.ts
usergenic: 👍

src/html/html-document.ts
rictic: ah, thanks!

PR #346, opened 2016-10-09
Teach warning printer to print multi line warnings.

rictic:
This will let us print better lint errors, and it also proves useful in an upcoming set of tests.

### DIFF
--- a/src/test/warning/warning-printer_test.ts
+++ b/src/test/warning/warning-printer_test.ts
@@ -33,6 +33,17 @@ const dumbNameWarning: Warning = {
   }
 };
 
+const goodJobWarning: Warning = {
+  message: 'Good job with this observedAttributes getter.',
+  code: 'cool-observed-attributes',
+  severity: Severity.INFO,
+  sourceRange: {
+    file: 'vanilla-elements.js',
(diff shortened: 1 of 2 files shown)


src/test/warning/warning-printer_test.ts
```diff
@@ -49,7 +60,7 @@ suite('WarningPrinter', () => {
 
   test('can handle printing no warnings', async() => {
     await printer.printWarnings([]);
-    assert.equal(output.toString(), '');
+    assert.deepEqual(output.toString(), '');
```
FredKSchott: for here and below: is this really necessary? Similar strings should have equality.

FredKSchott:
As an overall comment, thanks for breaking this up into smaller PRs! Much easier to review :)

src/test/warning/warning-printer_test.ts
rictic: They're semantically identical, but the output is *way* more legible for multiline strings, as it will display a diff of only the lines that didn't match, so I think deepEqual is the better default.

rictic:
Thanks for the quick review!

PR #347, opened 2016-10-09
Teach HTML document how to get many more kinds of SourceRanges.

rictic:
Builds on top of #346 

This lets us get source ranges for attribute names, attribute values, start tags, and end tags!

Most of the code is in tests. I changed our document source range tests to test against string representations of the source ranges, using some warning printer functionality. Makes it possible for a reviewer to eyeball the assertions rather than having to painstakingly track down line and column numbers.

No code uses these powers... yet. This is used in a follow-on PR.

### DIFF
--- a/src/test/html/html-document_test.ts
+++ b/src/test/html/html-document_test.ts
@@ -17,105 +17,501 @@ import * as dom5 from 'dom5';
 import * as fs from 'fs';
 import * as parse5 from 'parse5';
 import * as path from 'path';
-import {HtmlParser} from '../../html/html-parser';
+
+import {Analyzer} from '../../analyzer';
 import {ParsedHtmlDocument} from '../../html/html-document';
+import {HtmlParser} from '../../html/html-parser';
+import {SourceRange} from '../../model/model';
+import {FSUrlLoader} from '../../url-loader/fs-url-loader';
+import {WarningPrinter} from '../../warning/warning-printer';
 
 suite('ParsedHtmlDocument', () => {
+  const parser: HtmlParser = new HtmlParser();
+  const url = './source-ranges/html-complicated.html';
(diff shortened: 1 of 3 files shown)


rictic:
Swapping out Brendan for Justin, as Brendan has experience with this code and Justin is oversubscribed with summit business.

src/test/html/html-document_test.ts
```diff
@@ -17,105 +17,501 @@ import * as dom5 from 'dom5';
...
+  const warningPrinter = new WarningPrinter(null as any, {analyzer});
 
-  suite('sourceRangeForNode()', () => {
-    const parser: HtmlParser = new HtmlParser();
-    const url = '/static/source-ranges/html-complicated.html';
-    const file = fs.readFileSync(
-        path.resolve(__dirname, `../${url}`), 'utf8');
-    const document: ParsedHtmlDocument = parser.parse(file, url);
+  const getUnderlinedText = async(range: SourceRange | undefined) => {
+    if (range == null) {
```
FredKSchott: can this throw an error? Or is this ever expected to be undefined in the tests below?

rictic: This could throw an error, but IMO the error message is nicer by returning this string rather than throwing. 

e.g. this error message

```diff
 - expected, + actual
+  No source range produced.
-     <!-- Single Line Comment -->
-     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

That's easier to track down where the problem is than a message that mentions the error and a line number.

usergenic: The type is `SourceRange | undefined` but you're checking for null.  Even though this works in the `==` case it feels like the check should be for `undefined` which would also work for both cases.

rictic:
Merged!

PR #348, opened 2016-10-09
Rewrite code for mapping source positions to HTML ASTs

rictic:
This fixes a bunch of bugs, including Polymer/atom-plugin#7 Polymer/atom-plugin#8 and Polymer/atom-plugin#9 as well as a *lot* of subtle edge cases that we weren't getting right.

There's probably more work to be done here, I didn't test any multiline use cases here, though I'm hoping that since this builds on #347 that most of that stuff should Just Work.

Also I made some methods protected that should have been (they were only ever intended to be defined or called by subclasses anyways).

/cc @TimvdLippe

### DIFF
--- a/src/editor-service/ast-from-source-position.ts
+++ b/src/editor-service/ast-from-source-position.ts
@@ -13,54 +13,62 @@
 
 import * as parse5 from 'parse5';
 
-import {ParsedHtmlDocument} from '../html/html-document';
-import {SourceRange} from '../model/model';
+import { ParsedHtmlDocument } from '../html/html-document';
+import { SourceRange } from '../model/model';
 
-import {SourcePosition} from './editor-service';
+import { SourcePosition } from './editor-service';
 
 
-export type LocationResult = LocatedAttribute | LocatedAttributeValue |
-    LocatedTag | LocatedEndTag | LocatedInText | LocatedInScript |
-    LocatedInStyle | LocatedInComment;
+export type LocationResult = AttributesSection | AttributeValue |
+  TagName | EndTag | TextNode | ScriptContents |
+  StyleContents | Comment;
+
 /** In the tagname of a start tag. */
-export interface LocatedTag {
+export interface TagName {
   kind: 'tagName';
   element: parse5.ASTNode;
 }
 /** In an end tag. */
-export interface LocatedEndTag {
+export interface EndTag {
   kind: 'endTag';
   element: parse5.ASTNode;
 }
+
 /** In the attributes section of a start tag. Maybe in an attribute name. */
-export interface LocatedAttribute {
+export interface AttributesSection {
   kind: 'attribute';
-  attribute: string|null;
+  /** The attribute name that we're hovering over, if any. */
+  attribute: string | null;
+  /** The element whose start tag we're in. */
   element: parse5.ASTNode;
 }
+
 /** In the value of an attribute of a start tag. */
-export interface LocatedAttributeValue {
+export interface AttributeValue {
   kind: 'attributeValue';
   attribute: string;
   element: parse5.ASTNode;
 }
+
 /** In a text node. */
-export interface LocatedInText {
+export interface TextNode {
   kind: 'text';
   textNode?: parse5.ASTNode;
 }
 /** In the text of a <script> */
-export interface LocatedInScript {
+export interface ScriptContents {
   kind: 'scriptTagContents';
   textNode?: parse5.ASTNode;
 }
+
 /** In the text of a <style> */
-export interface LocatedInStyle {
+export interface StyleContents {
   kind: 'styleTagContents';
   textNode?: parse5.ASTNode;
 }
-/** In a comment. */
-export interface LocatedInComment {
+
+/** In a <!-- comment --> */
+export interface Comment {
   kind: 'comment';
   commentNode: parse5.ASTNode;
 }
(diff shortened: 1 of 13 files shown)


TimvdLippe:
Confirmed it fixes most of the issues :tada: 

Some issues I still observed:

1. Type `<dom-module></dom-module>`
* Type `id=""` in the start tag
* Type `be` inside the quotes and observe that the suggestions list shows `<behavior-test-elem>`. Note that actually letting it autocomplete does not work (known issue).

And another one:

1. Type `<beh` and autocomplete to `<behavior-test-elem>`
* Go back to the start tag and remove the space
* Hit ctrl + space before the `>` of the start tag
* Observe that no suggestions are shown

rictic:
Hm. So for the first issue, I can reproduce something similar. I'm seeing autocompletions of attribute names for attribute values, which I tried but apparently did not succeed in fixing in this PR. Will take another shot at that.

 I haven't been able to reproduce getting element name suggestions for attribute values though. Mind pasting in the exact text of the file that you're working with, with a `|` where the cursor is?


For the second issue  `<behavior-test-elem|>`, is https://github.com/Polymer/atom-plugin/issues/12 a fair summary of the ideal behavior?

TimvdLippe:
For the second issue instead of showing docs, I would prefer to have attribute completion with the space added in front. The heuristic would probably that if the start tag is closed with a > then it should not autocomplete on the element name but the attributes instead.

rictic:
The logic that I'm going for here is that you get suggestions based on what would happen if you started typing at the current cursor position.

I can see a user wanting either docs for that tag and I can see them wanting to get attribute completions. Both seem like reasonable expectations to me, and I lean towards keeping it as is (but with displaying docs) for consistency reasons.

src/editor-service/ast-from-source-position.ts
```diff
@@ -0,0 +1,280 @@
...
+import {ParsedHtmlDocument} from '../html/html-document';
+import {SourceRange} from '../model/model';
+
+import {SourcePosition} from './editor-service';
+
+
+export type LocationResult = LocatedAttribute | LocatedAttributeValue |
+    LocatedTag | LocatedEndTag | LocatedInText | LocatedInScript |
+    LocatedInStyle | LocatedInComment;
+/** In the tagname of a start tag. */
```
justinfagnani: nit: add some newlines

src/editor-service/ast-from-source-position.ts
```diff
@@ -0,0 +1,280 @@
...
+}
+/** In an end tag. */
+export interface LocatedEndTag {
+  kind: 'endTag';
+  element: parse5.ASTNode;
+}
+/** In the attributes section of a start tag. Maybe in an attribute name. */
+export interface LocatedAttribute {
+  kind: 'attribute';
+  attribute: string|null;
```
justinfagnani: what does the value `null` mean?

src/editor-service/ast-from-source-position.ts
```diff
@@ -0,0 +1,280 @@
...
+import {SourcePosition} from './editor-service';
+
+
+export type LocationResult = LocatedAttribute | LocatedAttributeValue |
+    LocatedTag | LocatedEndTag | LocatedInText | LocatedInScript |
+    LocatedInStyle | LocatedInComment;
+/** In the tagname of a start tag. */
+export interface LocatedTag {
+  kind: 'tagName';
+  element: parse5.ASTNode;
```
justinfagnani: consider naming this `node` so it's consistent across all `LocationResults`

src/editor-service/ast-from-source-position.ts
```diff
@@ -0,0 +1,280 @@
...
+/**
+ * Given a position and an HTML document, try to describe what new text typed
+ * at the given position would be.
+ *
+ * Where possible we try to return the ASTNode describing that position, but
+ * sometimes there does not actually exist one. (for a simple case, the empty
+ * string should be interpreted as a text node, but there is no text node in
+ * an empty document, but there would be after the first character was typed).
+ */
+export function getLocationInfoForPosition(
```
justinfagnani: future: definitely seems like this could be a method on Document

src/editor-service/ast-from-source-position.ts
```diff
@@ -0,0 +1,280 @@
...
+    return {kind: 'text'};
+  }
+  return location;
+}
+function _getLocationInfoForPosition(
+    node: parse5.ASTNode, position: SourcePosition,
+    document: ParsedHtmlDocument): undefined|LocationResult {
+  const sourceRange = document.sourceRangeForNode(node);
+  const location = node.__location;
+  if (!(sourceRange && location)) {
```
justinfagnani: can you document the reasoning behind this condition? it seems to be that if you can't get a SourceRange for a node, you recurse into its children, but how would you know that's the right thing to do if you can't get a SourceRange? Naively I would think this case should return `null`.

src/editor-service/ast-from-source-position.ts
```diff
@@ -0,0 +1,280 @@
...
+  }
+
+  const locationInChildren = _findLocationInChildren(node, position, document);
+  if (locationInChildren) {
+    return locationInChildren;
+  }
+
+  /**
+   * TODO(rictic): upstream the fact that regular locations can have attrs
+   * sometimes.
```
justinfagnani: what does that mean?

src/editor-service/ast-from-source-position.ts
```diff
@@ -0,0 +1,280 @@
...
+   */
+  const attrs: parse5.AttributesLocationInfo =
+      (isElementLocationInfo(location) && location.startTag.attrs) ||
+      location['attrs'] || {};
+
+  for (const attrName in attrs) {
```
justinfagnani: consider factoring this into a function

src/editor-service/ast-from-source-position.ts
```diff
@@ -0,0 +1,280 @@
...
+
+  const startTagRange = document.sourceRangeForStartTag(node);
+  const endTagRange = document.sourceRangeForEndTag(node);
+
+  // If we're in the end tag... we're in the end tag.
+  if (isPositionInsideRange(position, endTagRange, false)) {
+    return {kind: 'endTag', element: node};
+  }
+
+  if (startTagRange && isPositionInsideRange(position, startTagRange, false)) {
```
justinfagnani: why a truthy check on `startTagRange` here, but not on `endTagRange` above?

src/editor-service/ast-from-source-position.ts
rictic: It's a typings thing, expanded the comment.

src/editor-service/ast-from-source-position.ts
rictic: Done

src/editor-service/ast-from-source-position.ts
rictic: For the compiler's sake, because we use startTagRange inside the if.

PR #349, opened 2016-10-09
A few miscellaneous warning improvements.

rictic:
* for import errors, underline just the url
* fix the error code for behavior-not-found
* use getUnderlinedText to test source ranges in getWarningsForFile

### DIFF
--- a/src/polymer/polymer-element.ts
+++ b/src/polymer/polymer-element.ts
@@ -224,7 +224,7 @@ function _getFlattenedAndResolvedBehaviors(
               `\`${behavior.name}\`. Did you import it? Is it annotated with ` +
               `@polymerBehavior?`,
           severity: Severity.ERROR,
-          code: 'parse-error',
+          code: 'unknown-polymer-behavior',
           sourceRange: behavior.sourceRange
         });
       }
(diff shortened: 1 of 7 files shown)


src/polymer/polymer-element.ts
```diff
@@ -226,7 +226,7 @@ function _getFlattenedAndResolvedBehaviors(
             `\`${behavior.name}\`. Did you import it? Is it annotated with ` +
             `@polymerBehavior?`,
         severity: Severity.ERROR,
-        code: 'parse-error',
+        code: 'unknown-polymer-behavior',
```
FredKSchott: 👍

src/test/editor-service/editor-service_test.ts
```diff
@@ -450,59 +461,6 @@ function editorTests(editorFactory: (basedir: string) => EditorService) {
                     }]);
         });
   });
-
-  suite('getWarnings', function() {
```
FredKSchott: are these tested somewhere else? or are these no longer valuable?

rictic: Turns out they'd gotten duplicated by some merge^Wrebase along the way. See `suite('getWarningsForFile', ...)` above.

PR #354, opened 2016-10-11
Element reference scanner

43081j:
This is far from complete but looking for feedback.

Essentially, this scanner will find custom element references (tags in the HTML document). This is a fairly useful scanner for something like polymer-linter to retrieve a full list of referenced custom elements.

I'm still getting my head around much of the analyser so forgive me if I make any obvious mistakes and such...

## Feedback/Questions

**What is the best way to populate `attributes` with the tag's attributes from parse5?**

I would like to populate the `attributes` property of `ElementReference` but I haven't quite figured out where I should do that, yet.

**Should the classes be prefixed with `Custom` and a base implementation added?**

Currently, we confusingly call it `ElementReference` but it is actually specific to _custom_ elements. Maybe we should have an `ElementReferenceScanner` to find _all_ elements, and a `CustomElementReferenceScanner` which extends it to find only _custom_ elements?

Or simply prefix it all with `Custom` and have no inherited base, because maybe we don't ever expect to find every single HTML element.

**Currently `<template>` content isn't traversed, so I'm unable to test it**

I was told by @rictic that this is soon to be fixed in parse5, so will leave it for now.

### DIFF
--- a/src/html/html-element-reference-scanner.ts
+++ b/src/html/html-element-reference-scanner.ts
@@ -1,55 +1,9 @@
 import {HtmlScanner} from './html-scanner';
 import {ParsedHtmlDocument, HtmlVisitor} from './html-document';
-import {Resolvable, SourceRange, Feature} from '../model/model';
-import {Warning} from '../warning/warning';
 import * as dom5 from 'dom5';
+import {ScannedElementReference} from '../model/element-reference';
 
 const isCustomElement = dom5.predicates.hasMatchingTagName(/(.+-)+.+/);
-
-export interface Attribute {
-  name: string;
-  sourceRange: SourceRange|undefined;
-  value?: string;
-}
-
-export class ElementReference implements Feature {
-  tagName?: string;
-  attributes: Attribute[] = [];
-  sourceRange: SourceRange;
-  astNode: dom5.Node;
-  warnings: Warning[];
-  kinds: Set<string> = new Set(['element-reference']);
-
-  get identifiers(): Set<string> {
-    const result: Set<string> = new Set();
-    if (this.tagName) {
-      result.add(this.tagName);
-    }
-    return result;
-  }
-}
-
-export class ScannedElementReference implements Resolvable {
-  tagName?: string;
-  attributes: Attribute[] = [];
-  sourceRange: SourceRange;
-  astNode: dom5.Node;
-  warnings: Warning[];
-
-  constructor(
-    tagName: string, sourceRange: SourceRange,
-    ast: dom5.Node) {
-      this.tagName = tagName;
-      this.sourceRange = sourceRange;
-      this.astNode = ast;
-  }
-
-  resolve(): ElementReference {
-    const ref = new ElementReference();
-    Object.assign(ref, this);
-    return ref;
-  }
-}
 
 export class HtmlElementReferenceScanner implements HtmlScanner {
   async scan(
--- a/src/html/html-element-reference-scanner.ts
+++ b/src/html/html-element-reference-scanner.ts
@@ -19,7 +19,7 @@ export class HtmlElementReferenceScanner implements HtmlScanner {
             document.sourceRangeForNode(node)!,
             node);
 
-          if(node.attrs) {
+          if (node.attrs) {
             for (const attr of node.attrs) {
               element.attributes.push({
                 name: attr.name,
--- a/src/html/html-element-reference-scanner.ts
+++ b/src/html/html-element-reference-scanner.ts
@@ -6,7 +6,7 @@ import * as dom5 from 'dom5';
 
 const isCustomElement = dom5.predicates.hasMatchingTagName(/(.+-)+.+/);
 
-export interface ScannedAttribute {
+export interface Attribute {
   name: string;
   sourceRange: SourceRange|undefined;
   value?: string;
@@ -14,7 +14,7 @@ export interface ScannedAttribute {
 
 export class ElementReference implements Feature {
   tagName?: string;
-  attributes: ScannedAttribute[] = [];
+  attributes: Attribute[] = [];
   sourceRange: SourceRange;
   astNode: dom5.Node;
   warnings: Warning[];
@@ -31,7 +31,7 @@ export class ElementReference implements Feature {
 
 export class ScannedElementReference implements Resolvable {
   tagName?: string;
-  attributes: ScannedAttribute[] = [];
+  attributes: Attribute[] = [];
   sourceRange: SourceRange;
   astNode: dom5.Node;
   warnings: Warning[];
@@ -59,7 +59,6 @@ export class HtmlElementReferenceScanner implements HtmlScanner {
       let elements: ScannedElementReference[] = [];
 
       await visit((node) => {
-        console.log(node.nodeName, isCustomElement(node));
         if (isCustomElement(node) && node.nodeName !== 'dom-module') {
           elements.push(new ScannedElementReference(
             node.nodeName,
--- a/src/html/html-element-reference-scanner.ts
+++ b/src/html/html-element-reference-scanner.ts
@@ -19,14 +19,16 @@ export class HtmlElementReferenceScanner implements HtmlScanner {
             document.sourceRangeForNode(node)!,
             node);
 
-          for (const attr of node.attrs) {
-            element.attributes.push({
-              name: attr.name,
-              value: attr.value,
-              sourceRange: document.sourceRangeForAttribute(node, attr.name),
-              nameSourceRange: document.sourceRangeForAttributeName(node, attr.name),
-              valueSourceRange: document.sourceRangeForAttributeValue(node, attr.name)
-            });
+          if(node.attrs) {
+            for (const attr of node.attrs) {
+              element.attributes.push({
+                name: attr.name,
+                value: attr.value,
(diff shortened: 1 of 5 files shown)


src/test/html/html-element-reference-scanner_test.ts
```diff
@@ -0,0 +1,42 @@
...
+        </body></html>`;
+      const document = new HtmlParser().parse(contents, 'test-document.html');
+      let visit = async(visitor: HtmlVisitor) => document.visit([visitor]);
+
+      const features = await scanner.scan(document, visit);
+      assert.equal(features.length, 2);
+      assert.instanceOf(features[0], ScannedElementReference);
+      assert.instanceOf(features[1], ScannedElementReference);
+
+      const feature0 = <ScannedElementReference>features[0];
```
rictic: Do you need these casts? The result of HtmlElementReferenceScanner is a promise of ScannedElementReference[] isn't it?

src/test/html/html-element-reference-scanner_test.ts
```diff
@@ -0,0 +1,42 @@
...
+      assert.equal(feature0.tagName, 'x-foo');
```
rictic: I like using `assert.deepEquals` more than a lot of little assertions, because you get way better error messages, with diffs.

So e.g.

```typescript
assert.deepEquals(features.map(f => f.tagName), ['x-foo', 'x-bar']);
```

rictic:
Oh, also please enable this scanner by default by adding it to the scanners on `Analyzer`. I want to see what the impact is on our benchmark.

src/test/html/html-element-reference-scanner_test.ts
43081j: Yup I was thinking they would be `Feature` but actually they are already the correct type!

43081j:
@rictic done most of your suggestions i think and expanded tests.

Will do a little renaming and such soon to add the base class i mentioned (so we have one for "all elements" and one for "custom elements").

src/test/html/html-element-reference-scanner_test.ts
```diff
@@ -0,0 +1,132 @@
...
+          <x-foo a=5 b="test" c></x-foo>
+          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~`,
+        `
+            <x-bar></x-bar>
+            ~~~~~~~~~~~~~~~`
+      ]);
+
+      const attrRanges = await Promise.all(features.map(
+        async f => await Promise.all(f.attributes.map(
+          async a => await getUnderlinedText(a.sourceRange)))));
```
43081j: if this is a little too crazy or could be done better, do let me know..

43081j:
I noticed occasionally, I keep getting CI failures because of the local service tests, any idea what's causing that? (seems intermittent)

rictic:
[_bump for clabot_]

rictic:
Finally tracked down the issue! You need to initialize `warnings` on ElementReference. Looks like the typescript compiler doesn't check that a declared property is initialized. Arg, that makes `strictNullChecks` way less useful.

Should be as simple as changing `warnings: Warning[];` to `warnings: Warning[] = [];` on ScannedElementReference.

src/model/element-reference.ts
```diff
@@ -0,0 +1,47 @@
...
+    return new Set([this.tagName]);
+  }
+}
+
+export class ScannedElementReference implements Resolvable {
+  tagName: string;
+  attributes: Attribute[] = [];
+  sourceRange: SourceRange;
+  astNode: dom5.Node;
+  warnings: Warning[];
```
rictic: Need to initialize this as well.

```typescript
warnings: Warning[] = [];
```

43081j:
@rictic i changed those warning refs.

However, due to this:
https://downloads.globalsign.com/acton/fs/blocks/showLandingPage/a/2674/p/p-008f/t/page/fm/0

I'm unable to see why the build is failing (chrome says travis-ci's SSL is revoked, well done globalsign!)

rictic:
Copy pasting from travis (why yes, I _can't_ sleep on the flight to london). Looks like one last failure, this one more obviously related to your code at least :)

```
 1) HtmlElementReferenceScanner scan() finds custom element references:
      AssertionError: expected [ Array(2) ] to deeply equal [ Array(2) ]
      + expected - actual
       [
         [
           "\n          <x-foo a=5 b=\"test\" c></x-foo>\n                   ~"
           "\n          <x-foo a=5 b=\"test\" c></x-foo>\n                       ~~~~~~"
      -    "No source range produced"
      +    "\n          <x-foo a=5 b=\"test\" c></x-foo>\n                              ~"
         ]
         []
       ]
      
      at Function.assert.deepEqual (node_modules/chai/lib/chai/interface/assert.js:205:32)
      at Object.<anonymous> (lib/test/html/test/html/html-element-reference-scanner_test.ts:114:14)
      at next (native)
      at fulfilled (lib/test/html/html-element-reference-scanner_test.js:4:58)
```

43081j:
oh that's a strange one!

its because the second one is `No source range produced`. This doesn't happen locally... maybe I missed some null check or similar somewhere. Will have to debug as that source range _should_ exist.

rictic:
If you run mocha directly you can end up testing stale sources. We've configured `npm test` to be very reliable, I have not seen any cases where `npm test` for the analyzer passed locally but failed on travis. The only thing it doesn't do is ensure node_modules is the same as it will be on travis (and I'm thinking about trying out yarn for just that purpose).

43081j:
Yup I'm sure it was an issue on my end, I'll try it again later and sort the tests out.

Still need to find time to separate out into a base scanner too when I can...

src/html/html-element-reference-scanner.ts
```diff
@@ -1,40 +1,52 @@
-import {HtmlScanner} from './html-scanner';
-import {ParsedHtmlDocument, HtmlVisitor} from './html-document';
 import * as dom5 from 'dom5';
```
rictic: Please add license header

43081j:
@rictic once you get chance can you check out the latest changes?

seems to be building now. I did separate the logic out a bit into the base class, seems to work fine but could do with a once over.

src/html/html-element-reference-scanner.ts
```diff
@@ -0,0 +1,52 @@
+import * as dom5 from 'dom5';
+import {ASTNode} from 'parse5';
+import {ScannedElementReference} from '../model/element-reference';
+import {HtmlVisitor, ParsedHtmlDocument} from './html-document';
+import {HtmlScanner} from './html-scanner';
+
+const isCustomElement = dom5.predicates.hasMatchingTagName(/(.+-)+.+/);
+
+export class HtmlElementReferenceScanner implements HtmlScanner {
```
rictic: Mind adding docs for this class? Also worth pointing out in the docs that this scanner isn't hooked up by default, but its subclass is.

src/test/html/html-element-reference-scanner_test.ts
```diff
@@ -0,0 +1,152 @@
...
+
+      assert.deepEqual(attrValueRanges, [
+        [
+          `
+          <x-foo a=5 b="test" c></x-foo>
+                   ~`,
+          `
+          <x-foo a=5 b="test" c></x-foo>
+                       ~~~~~~`,
+          `No source range produced`
```
rictic: Oh! Right, now I recall the relevant detail for why we had trouble with this assertion. I recently changed the code to get attribute value ranges to return undefined for boolean attributes.

So when it was failing before that was because you had tested locally without that change merged in, but travis was seeing it with it merged in.

src/test/html/html-element-reference-scanner_test.ts
```diff
@@ -0,0 +1,152 @@
+import {assert} from 'chai';
```
rictic: Please add license headers.

src/model/element-reference.ts
```diff
@@ -0,0 +1,47 @@
+import {Resolvable, SourceRange, Feature} from '../model/model';
+import {Warning} from '../warning/warning';
+import * as dom5 from 'dom5';
+
+export interface Attribute {
+  name: string;
+  sourceRange: SourceRange|undefined;
```
rictic: We should be able to type sourceRange and nameSourceRange without `undefined`. Just add `!` after the relevant expressions when you're creating the Attribute values. Only `valueSourceRange` really can be undefined, as far as I'm aware.

src/model/element-reference.ts
```diff
@@ -0,0 +1,47 @@
+import {Resolvable, SourceRange, Feature} from '../model/model';
```
rictic: Please add license header.

src/html/html-element-reference-scanner.ts
```diff
@@ -0,0 +1,52 @@
+import * as dom5 from 'dom5';
```
rictic: Please add license header.

src/test/html/html-element-reference-scanner_test.ts
43081j: aha i suspected that when i saw your recent commits, but wasn't sure. now it makes sense

43081j:
I think that's most changes or all done now. 

Also sorry that it changed the import scanner test, seems when i ran clang-format that happened without me realising. Would have removed the changes otherwise.

PR #355, opened 2016-10-11
Fix duplicate warnings

usergenic:
`Document#getWarnings()` was returning duplicate warnings because of the way `this.warnings` and `feature.getWarnings()` apparently contained the same items.  I just put a guard against pushing existing warnings into the result array.  Separated commits so there's one that proves the dup.

### DIFF
--- a/src/model/document.ts
+++ b/src/model/document.ts
@@ -291,7 +291,7 @@ export class Document implements Feature {
     warnings.push.apply(warnings, this.warnings);
     for (const feature of this.getFeatures(deep)) {
       for (const warning of feature.warnings) {
-        if (!warnings.includes(warning)) {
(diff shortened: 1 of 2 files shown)


src/model/document.ts
```diff
@@ -290,7 +290,11 @@ export class Document implements Feature {
     }
     warnings.push.apply(warnings, this.warnings);
     for (const feature of this.getFeatures(deep)) {
-      warnings.push.apply(warnings, feature.warnings);
+      for (const warning of feature.warnings) {
+        if (warnings.indexOf(warning) === -1) {
```
rictic: Hm, this is O(n^2) in the number of warnings. Consider either making `warnings` a Set or maybe check whether `feature === this`

PR #358, opened 2016-10-13
mark computed props as readOnly

### LINKED ISSUE #350: Computed properties are readonly
We should mark computed properties readOnly

43081j:
Fixes #350.

I also added a bunch of polymer-element tests so I was able to test that `readOnly` is set correctly.

I did pass through clang-format but if you think there's anything horrible going on in here, do review!

If you are wondering why I use an `isComputed` flag, it is to avoid tripping up on the following (also found in tests):

```javascript
{
    computed: '_computeMe()',
    readOnly: false
}
```

So it will still mark as `readOnly: true` instead of being mistakenly overridden by the property there. If this is wrong behaviour, do let me know.

### DIFF
--- a/src/test/polymer/polymer-element-scanner_test.ts
+++ b/src/test/polymer/polymer-element-scanner_test.ts
@@ -1,5 +1,18 @@
-import {assert} from 'chai';
+/**
+ * @license
+ * Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at
+ * http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
 
+import {assert} from 'chai';
 import {Visitor} from '../../javascript/estree-visitor';
 import {JavaScriptParser} from '../../javascript/javascript-parser';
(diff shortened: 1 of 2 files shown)


src/test/polymer/polymer-element-scanner_test.ts
```diff
@@ -0,0 +1,119 @@
+import {assert} from 'chai';
```
rictic: Please add the license header here

src/test/polymer/polymer-element-scanner_test.ts
```diff
@@ -0,0 +1,119 @@
...
+                         sourceType: 'script'
+                       }).parse(contents, 'test-document.html');
+      let visit = async(visitor: Visitor) => document.visit([visitor]);
+
+      const features = await scanner.scan(document, visit);
+
+      assert.equal(features.length, 2);
+      assert.instanceOf(features[0], ScannedPolymerElement);
+      assert.instanceOf(features[1], ScannedPolymerElement);
+      assert.deepEqual(features.map(f => f.tagName), ['x-foo', 'x-bar']);
```
rictic: This deepEqual is better than the assert on `.length` because it asserts the same thing, but will give a much better error message if the assertion ever fails (e.g. instead of `expected 1 to equal 2` you'll get `expected ['x-bar'] to equal ['x-foo', 'x-bar']`. Recommend removing the length assertion.

The instanceOf asserts are fine but IMO elidable, the type system does a fine job of it.

43081j: I agree, the tag assertion is enough and we don't need to check the instances seeing as our type checking would have found it. Will remove those

43081j:
This should be sorted now, added the license & removed some redundant assertions.

PR #360, opened 2016-10-18
better support of lazy imports

garlicnation:
Avoid returning features defined in lazy imports.

This is suboptimal. The better solution would be for lazy imports to "color" features that are transitively loaded through them, but I'm unsure of what that would look like.


rictic:
Please add a test and document this change in the CHANGELOG

rictic:
Support for this was added in #543

PR #368, opened 2016-10-23
Use yarn

rictic:
Should speed up our test runs a bit (hopefully by rather a lot once https://github.com/yarnpkg/yarn/pull/1319 lands and is released), but more importantly it will give us consistency. Precious, precious consistency.


.travis.yml
```diff
@@ -2,5 +2,9 @@ language: node_js
 node_js:
   - "4"
   - "6"
+before_install:
+- npm i -g yarn
+install:
+- yarn install --ignore-scripts --prefer-offline
 script:
 - npm test && npm run benchmark
```
FredKSchott: can we move these to yarn as well? If we're going to do this lets go all the way :)

justinfagnani: +1

FredKSchott:
Oh interesting, looks like yarn team says otherwise about checking in yarn.lock https://yarnpkg.com/blog/2016/11/24/lockfiles-for-all

justinfagnani:
Yeah, but I don't think that's a slam-dunk. The yarn.lock is still ignored for transitive dependencies, but it would be used for tests. Basically by default we're only going to get coverage of one variant on tests, the question is do we want a fixed variant by using a lockfile, or do we want to pick up the latest dependencies.

If we only ever test the specific dependencies versions in the lock file, I'm not sure why we wouldn't just lock our actual dependencies to fixed versions and not ranges, since we can't actually say that we work with a range. Only an app using our library might get newer versions of our deps and discover any problems.

On the other hand the lock file could in some cases help figure out problems if we do have a failure.

I don't know the right answer here, except that it'd be nice to test under more than one variant if we advertize that we support a range of dependencies. One potential solution for testing the deps we claim we support, but getting some reproducibility would be to generate a lockfile on the CI server and echo it, so in the case of a failure we can copy it from the logs and install the exact versions of deps.

.travis.yml
```diff
@@ -2,5 +2,9 @@ language: node_js
 node_js:
   - "4"
   - "6"
+before_install:
+- npm i -g yarn
```
abdonrd: Now Travis CI supports Yarn :)

https://blog.travis-ci.com/2016-11-21-travis-ci-now-supports-yarn

TimvdLippe:
Would love to use yarn for tools such as the analyzer. Speeds up the workflow a lot! Also the package.json still references npm. Update these to yarn too?

.travis.yml
```diff
@@ -2,5 +2,9 @@ language: node_js
...
+install:
+- yarn install --ignore-scripts --prefer-offline
```
stramel: Use `yarn` instead of `yarn install`

.travis.yml
stramel: :+1:

.travis.yml
```diff
@@ -2,5 +2,9 @@ language: node_js
 node_js:
   - "4"
```
stramel: Remove node v4 for v7 or v8

FredKSchott:
Closing this PR as abandoned.

PR #373, opened 2016-10-30
Better warnings when resolving behaviors.

rictic:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+### Fixed
+* [Polymer] A number of fixes around warnings when resolving behaviors
+  * Warn, don't throw when a behavior is declared twice.
+  * Warn when there's a problem mixing behaviors into other behaviors, the same way that we warn when mixing behaviors into elements.
+
 ## [2.0.0-alpha.17] - 2016-10-28 - [minor]
 
 ### Added
--- a/src/polymer/polymer-element.ts
+++ b/src/polymer/polymer-element.ts
@@ -17,7 +17,7 @@ import * as estree from 'estree';
 
 import * as jsdoc from '../javascript/jsdoc';
 import {Document, Element, LiteralValue, Property, ScannedAttribute, ScannedElement, ScannedEvent, ScannedProperty, SourceRange} from '../model/model';
-import {Severity} from '../warning/warning';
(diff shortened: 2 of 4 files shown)


src/polymer/polymer-element.ts
```diff
@@ -204,39 +212,53 @@ function resolveElement(
...
+      warnings.push({
+        message: `Found more than one behavior named ${behavior.name}.`,
+        severity: Severity.WARNING,
+        code: 'multiple-polymer-behaviors',
+        sourceRange: behavior.sourceRange
+      });
+      // Don't skip processing this behavior, just take the most recently
+      // declared instance.
+    }
+    const foundBehavior = Array.from(foundBehaviors)[foundBehaviors.size - 1];
```
FredKSchott: code golf just because I was curious: `foundBehavior.values().next()`

src/test/analyzer_test.ts
```diff
@@ -89,7 +89,9 @@ suite('Analyzer', () => {
         async() => {
           let document =
               await analyzer.analyze('static/html-missing-behaviors.html');
-          let warnings = document.getWarnings();
+          // TODO(#372): this should be false, we should treat inline documents
```
FredKSchott: :( good catch, and good followup for creating the issue

CHANGELOG.md
```diff
@@ -8,6 +8,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+### Fixed
+* [Polymer] A number of fixes around warnings when resolving behaviors
```
FredKSchott: 👍 loving this

src/polymer/polymer-element.ts
rictic: That gets you the first one, but we actually want the last one, as last write wins.

(which itself isn't quite right because you can get different orderings of the same imports based on the shape of the total import graph and the starting node, but this is at least closer to correct)

(ok there's even more caveats because it could be that you want one in the _middle_ of the list of results because at runtime we look up the identifier at the time of the behavior reference… but we definitely don't want to bother with that (even though at one point in this code we totally did, the caching story is a nightmare))

rictic:
Thanks for the review!

src/polymer/polymer-element.ts
FredKSchott: I regret this entire thread

rictic: :fire: :computer: :fire:

PR #380, opened 2016-11-04
handle SequenceExpression nodes in behavior parsing

### LINKED ISSUE #379: Parse SequenceExpression statements in JavaScript
Right now we skip over JavaScript `SequenceExpression`s when we scan for Polymer behaviors. This means we miss any expressions that are joined by a comma, like this:

```js
/**
   * @polymerBehavior Polymer.TestBehavior
   */
Polymer.TestBehaviorImpl={properties:{}},/**
   * @polymerBehavior
   */
Polymer.TestBehavior=[Polymer.TestBehaviorImpl,SomeSimpleBehavior];
```

These `SequenceExpression`s are common in minified code, which means analyzing minified JavaScript is pretty broken right now.

See https://github.com/Polymer/polymer-analyzer/commit/2b293f99ba75460be274d2ed6e51d73e0eb79bfe on the `wip-sequence-expression-behaviors` branch for my progress towards fixing this. We'll need to add a `enterSequenceExpression()` method to `BehaviorVisitor` that can properly parse these. I added a basic version of this, but couldn't figure out how to get the different `initBehavior` & `finishBehavior` methods called in the proper order.

Also, we'll need handle how comments relate to `SequenceExpression`, where leading and trailing comments are on the `SequenceExpression` itself and not its first & last expressions.

/cc @rictic

FredKSchott:
Fixes #379

This fix lets our behavior scanner understand Sequence expressions (ex: `EXPRESSION, EXPRESSION, EXPRESSION;`) while scanning. Users will run into this problem if they do either of these two things:

1. **Use a strange coding style that involves lots of SequenceExpressions.** Unknown if this is actually anyone, and if there are it would be fair to just not support them.
2. **Minify/Preprocess their code before analysis.** This will be less common once we start analyzing apps before minification in the CLI, but there are still reasonable cases where this could occur (using JSX, CoffeeScript, other languages/features that may not get support right away in the analyzer).

As I opened this issue I was 50-50 on whether this was a useful fix, but as I wrote out (2.) above I can now see how this actually a useful fix. @rictic I'll let you make the ultimate call on merging this, but I think it's worth it as long as the code itself (where we patch the comments) doesn't feel like too much of a hack to you.

/cc @justinfagnani @usergenic

### DIFF
--- a/src/polymer/behavior-scanner.ts
+++ b/src/polymer/behavior-scanner.ts
@@ -67,8 +67,31 @@ class BehaviorVisitor implements Visitor {
   propertyHandlers: PropertyHandlers|null = null;
 
   document: JavaScriptDocument;
+
   constructor(document: JavaScriptDocument) {
     this.document = document;
+  }
+
+  /**
+   * Fix how comments are attached to SequenceExpressions. In espree, any
+   * leading comments to the expression are parsed as leading comments to the
+   * entire sequence and not the first expression in that sequence.
+   *
+   * For our behaviors we expect our `@polymerBehavior` comments to be the
+   * leading comments of the actual first expression, so we copy them over here
+   * before so that other enter functions can properly read them.
+   */
+  enterSequenceExpression(
+      node: estree.SequenceExpression, parent: estree.Node) {
+    node.expressions.forEach((expression, i) => {
+      if (i === 0 && !expression.leadingComments) {
+        node.leadingComments = node.leadingComments || parent.leadingComments;
+      } else if (
+          i === node.expressions.length - 1 && !expression.trailingComments) {
+        node.trailingComments =
+            node.trailingComments || parent.trailingComments;
+      }
+    });
   }
 
   /**
--- a/src/test/analyzer_test.ts
+++ b/src/test/analyzer_test.ts
@@ -323,6 +323,16 @@ suite('Analyzer', () => {
       await analyzer.analyze('static/caching/file2.html');
     });
 
+    test('can analyze behaviors documented in SequenceExpressions', async() => {
+      const document =
+          await analyzer.analyze('static/minified-javascript/index.html');
+      assert.equal(document.getWarnings(true).length, 0);
+      assert.equal(document.getByKind('behavior').size, 2);
+      assert.isObject(document.getOnlyAtId('behavior', 'Polymer.TestBehavior'));
+      assert.isObject(
+          document.getOnlyAtId('behavior', 'Polymer.SimpleBehavior'));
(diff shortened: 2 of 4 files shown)


src/polymer/behavior-scanner.ts
```diff
@@ -67,11 +67,34 @@ class BehaviorVisitor implements Visitor {
...
+   *
+   * For our behaviors we expect our `@polymerBehavior` comments to be the
+   * leading comments of the actual first expression, so we copy them over here
+   * before so that other enter functions can properly read them.
+   */
+  enterSequenceExpression(
+      node: estree.SequenceExpression, parent: estree.Node) {
+    node.expressions.forEach((expression, i) => {
+      if (i === 0 && !expression.leadingComments) {
+        node.leadingComments = node.leadingComments || parent.leadingComments;
```
rictic: Consider extracting this out instead to a `getLeadingComments` function somewhere, maybe on JavascriptDocument. There are a lot of cases where we want to do this sort of thing, and I bet this isn't the only place where we could do some parental walking to properly associate the right comments on the right nodes.

FredKSchott: That would be great, I think one even already exists in esutil. 

The problem is that by the time we walk to the expression inside the sequence, the comments no longer exist on `node` or `parent` so we would have nothing to call `getLeadingComments()` with. Is there a way that you know of to get a `Node`'s parent?

rictic: Damn, good point. Ideally we would have access to the complete stack of ancestors inside this visitor.

I'm not super happy with a visitor modifying the nodes that it's walking, as that means that potentially other visitors would be affected by the order in which they are run, tying us to the implementation details of the tree walking algorithm. Do you see a way to do this without modifying estree nodes?

justinfagnani: This can be solved by a base class that keeps a stack of ancestors. Given that we know we have an open ecosystem of visitors, I'd really discourage modifying the AST in any way. If for some reason maintaining a path stack doesn't work, a WeakMap of comments would be better. One pass to build the map, another pass to use it.

FredKSchott: Yea it's a total hack on top of the AST :/ 

I'm going to table this for now to work on the analyzer, since this isn't a high priority bug atm. hopefully we can revisit it soon, maybe when we have a better reason to rewrite the behavior scanners altogether.

rictic:
Removing myself for now as this is tabled. Please re-add me if you'd like me to take another look!

PR #381, opened 2016-11-04
Add .vscode debugging launch script

FredKSchott:
Useful for debugging in .vscode. The debug "play" button should just work for all of us now, since it's using the mocha dependency of the analyzer itself.

/cc @justinfagnani  @rictic @usergenic

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,22 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Launch Program",
+      "program": "${workspaceRoot}/node_modules/.bin/_mocha",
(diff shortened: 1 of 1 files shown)


justinfagnani:
Please no comments on json files

.vscode/launch.json
```diff
@@ -0,0 +1,25 @@
+{
+  // Use IntelliSense to learn about possible Node.js debug attributes.
```
justinfagnani: back to a computer so I can make an inline comment: while VS Code supports a superset of JSON with comments, not all tools do and we've had problems in the past with that. So let's keep all of our JSON valid, as restrictive as that may be.

FredKSchott: I thought about this but figured it was fine since this is a config file for vcscode, and vscode supports comments. But happy to change if it bothers you.

PR #383, opened 2016-11-04
remove unneccesary node_modules/.bin path from package.json

FredKSchott:
node_modules/.bin is added to the npm run PATH automatically. (The runs for me without clang-format installed)

/cc @rictic


FredKSchott:
Yea npm adds `node_modules/.bin` to the front of the PATH chain. Didn't know about `node-gyp` though, maybe something in Node relies on that being absolutely first

PR #385, opened 2016-11-08
Scan A-Frame components; scanners return warnings.

rictic:

Adds first support for a non-polymer web components library: A-Frame! This patch will recognize certain kinds of invalid A-Frame component registrations. It also sets the groundwork for doing more interesting stuff later once we also recognize A-Frame entity definitions.

In order to make that work, I noticed that there was no way to return warnings from a scanner unless those warnings had a feature to attach to, so I refactored the return type of #scan().

### DIFF
--- a/src/a-frame/component-scanner.ts
+++ b/src/a-frame/component-scanner.ts
@@ -1,0 +1,131 @@
+/**
+ * @license
+ * Copyright (c) 2015 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at
+ * http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import * as estree from 'estree';
+
+import * as astValue from '../javascript/ast-value';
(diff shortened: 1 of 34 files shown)


src/scanning/scanner.ts
```diff
@@ -15,10 +15,15 @@
 import {Analyzer} from '../analyzer';
 import {ScannedFeature} from '../model/model';
 import {ParsedDocument} from '../parser/document';
+import {Warning} from '../warning/warning';
+
+export interface ScanResult {
```
FredKSchott: my reaction to reading this: are these all warnings found during scanning? Or are these just feature-unrelated warnings? If there's a warning on a feature, does it show up in the warnings array, in the feature.warnings property, or both?

Perhaps rename `warnings` to `otherWarnings` or something that clarifies?

src/a-frame/component-scanner.ts
```diff
@@ -0,0 +1,131 @@
+/**
```
FredKSchott: could you add some more inline documentation to this file? I haven't used a-frame before and had to make some assumptions about what you were trying to do here

- scanComponent
- enterCallExpression
- other?

src/javascript/javascript-import-scanner.ts
```diff
@@ -39,7 +40,7 @@ export class JavaScriptImportScanner implements JavaScriptScanner {
             document.sourceRangeForNode(node.source)!, node));
       }
     });
-    return imports;
+    return {features: imports, warnings: []};
```
FredKSchott: for these and others, would it be safer to return `warnings: visitor.warnings` even if we're assuming today that visitor will have no warnings?

PR #389, opened 2016-11-12
Handle mutually recursive import graphs.

### LINKED ISSUE #388: Circular imports result in analyze promise never being resolved nor rejected
Title says it all. Seems circular imports result in a stalemate situation where two promises wait for each other to be resolved/rejected which means neither ever are.

rictic:
This required two changes:
  * break up scanning a document and its dependencies. cache the scanned document once it has been scanned but before its dependencies are.
  * track a scanned document's inline status a bit better, and don't cache inline scanned documents. This was always a hidden error that was masked by the earlier order of operations.

Fixes #388

### DIFF
--- a/src/analyzer.ts
+++ b/src/analyzer.ts
@@ -231,6 +231,9 @@ export class Analyzer {
     this._analyzedDocumentPromises.clear();
   }
 
+  /**
+   * Scan a toplevel document given its url and optionally its contents.
+   */
   private async _scan(resolvedUrl: string, contents?: string):
       Promise<ScannedDocument> {
     const cachedResult = this._scannedDocumentPromises.get(resolvedUrl);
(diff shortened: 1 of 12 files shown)


src/analyzer.ts
```diff
@@ -242,30 +245,35 @@ export class Analyzer {
...
+    await this._scanDependencies(scannedDoc);
+    return scannedDoc;
   }
 
   /**
    * Scans a parsed Document object.
    */
   private async _scanDocument(
-      document: ParsedDocument<any, any>,
+      document: ParsedDocument<any, any>, isInline: boolean,
```
FredKSchott: is there no property on ParsedDocument that could convey this instead? Would a property like this be useful for other things during analysis? I'd like to see that more than a one-off argument here if possible.

FredKSchott: Also, it's confusing when I'd want to use `this._scanInlineDocument()` vs. `this._scanDocument(isInline: true)`. Can this logic go in to _scanInlineDocument instead? If not, can you document the differences / when you'd want to use either?

rictic: +1 to moving this property to ParsedDocument, that's way more elegant.

When you'd want to call _scanInlineDocument vs _scan vs _scanDocument is fairly clear based on their now I think, PTAL

JohnRiv:
fwiw, I'm also experiencing the issue #388, so I tested this out and it worked

src/analyzer.ts
```diff
@@ -276,34 +283,47 @@ export class Analyzer {
...
+    return scannedDocument;
+  }
+
+  /**
+   * Scan all the dependencies of the given scanned document.
+   *
+   * This must be called exactly once per scanned document, as we mutate
+   * the given scannedDocument by adding warnings.
+   */
+  private async _scanDependencies(scannedDocument: ScannedDocument) {
```
FredKSchott: nit: can you add a return type here? it's hard to tell what this function is meant to return

src/css/css-parser.ts
```diff
@@ -29,13 +29,14 @@ export class CssParser implements Parser<ParsedCssDocument> {
...
     let ast = this._parser.parse(contents);
+    const isInline = !!inlineInfo;
     inlineInfo = inlineInfo || {};
     return new ParsedCssDocument({
       url,
       contents,
       ast,
       locationOffset: inlineInfo.astNode,
-      astNode: inlineInfo.astNode
+      astNode: inlineInfo.astNode, isInline,
```
FredKSchott: should be new line?

src/html/html-parser.ts
```diff
@@ -29,13 +29,14 @@ export class HtmlParser implements Parser<ParsedHtmlDocument> {
...
     let ast = parseHtml(contents, {locationInfo: true});
+    const isInline = !!inlineInfo;
     inlineInfo = inlineInfo || {};
     return new ParsedHtmlDocument({
       url,
       contents,
       ast,
       locationOffset: inlineInfo.locationOffset,
-      astNode: inlineInfo.astNode
+      astNode: inlineInfo.astNode, isInline,
```
FredKSchott: ditto (for this and others), should be new line?

src/css/css-parser.ts
rictic: This is what clang does. 🏌

FredKSchott: clang you're so silly. Sounds good then

src/analyzer.ts
rictic: Sure thing, done! Rebased it into the previous commit

PR #390, opened 2016-11-12
Add test that demonstrates issues with caching.

rictic:
The test is skipped for now, but it demonstrates that our caching strategy doesn't work properly with modifications of files coming in while other analyses are happening.

I believe this demonstrates the bug that's happening in https://github.com/Polymer/polymer-editor-service/issues/28

 - [x] CHANGELOG.md not updated, entirely internal change.

### DIFF
--- a/src/test/analyzer_test.ts
+++ b/src/test/analyzer_test.ts
@@ -27,7 +27,8 @@ import {HtmlParser} from '../html/html-parser';
 import {ScriptTagImport} from '../html/html-script-tag';
 import {JavaScriptDocument} from '../javascript/javascript-document';
 import {ParsedCssDocument} from '../css/css-document';
-import {Document, Import, ScannedImport, ScannedInlineDocument} from '../model/model';
+import {Document, ScannedImport, Import, ScannedInlineDocument} from '../model/model';
+import {UrlLoader} from '../url-loader/url-loader';
 import {FSUrlLoader} from '../url-loader/fs-url-loader';
 import {UrlResolver} from '../url-loader/url-resolver';
 
@@ -521,6 +522,65 @@ suite('Analyzer', () => {
 
       assert.deepEqual(
           testSeed.events.map(e => e.name), ['fired-event', 'data-changed']);
+    });
+  });
+
+  suite.skip('race conditions and caching', () => {
+    const wait = () =>
+        new Promise((resolve) => setTimeout(resolve, Math.random() * 30));
+    class RacyUrlLoader implements UrlLoader {
+      constructor(public map: Map<string, string>) {
+      }
+      canLoad() {
+        return true;
+      }
+      async load(path: string) {
+        await wait();
+        const result = this.map.get(path);
+        if (result != null) {
+          return result;
+        }
+        throw new Error(`no known contents for ${path}`);
+      }
+    }
+    test('editor simulator of imports that import a common dep', async() => {
+      // Here we're simulating a lot of noop-changes to base.html, which has
+      // two imports, which mutually import a common dep. This stresses the
+      // analyzer's caching.
+      const contentsMap = new Map<string, string>([
+        [
+          'base.html',
+          `<link rel="import" href="a.html">\n<link rel="import" href="b.html">`
+        ],
+        ['a.html', `<link rel="import" href="common.html">`],
+        ['b.html', `<link rel="import" href="common.html">`],
+        ['common.html', `<custom-el></custom-el>`],
+      ]);
+      const analyzer =
+          new Analyzer({urlLoader: new RacyUrlLoader(contentsMap)});
+      const promises: Promise<Document>[] = [];
+      for (let i = 0; i < 30; i++) {
+        await wait();
+        for (const key of contentsMap) {
+          if (Math.random() > 0.5) {
+            analyzer.analyze(key[0], key[1]);
+          }
+        }
+        promises.push(analyzer.analyze('base.html'));
+      }
+      const documents = await Promise.all(promises);
+      for (const document of documents) {
+        const imports = Array.from(document.getByKind('import'));
+        assert.deepEqual(
+            imports.map(m => m.url).sort(),
+            ['a.html', 'b.html', 'common.html', 'common.html']);
+        const docs = Array.from(document.getByKind('document'));
+        assert.deepEqual(
+            docs.map(d => d.url).sort(),
+            ['a.html', 'b.html', 'base.html', 'common.html']);
+        const refs = Array.from(document.getByKind('element-reference'));
+        assert.deepEqual(refs.map(ref => ref.tagName), ['custom-el']);
+      }
     });
   });
 });
--- a/src/test/analyzer_test.ts
+++ b/src/test/analyzer_test.ts
@@ -141,7 +141,6 @@ suite('Analyzer', () => {
               chainedDocument.getWarnings(true), [expectedWarning]);
         });
 
-    // currently passing
     test('analyzes multiple imports of the same behavior', async() => {
       const documentA = await analyzer.analyze(
           'static/multiple-behavior-imports/element-a.html');
@@ -151,7 +150,6 @@ suite('Analyzer', () => {
       assert.deepEqual(documentB.getWarnings(true), []);
     });
 
-    // currently failing
     test.skip(
         'analyzes multiple imports of the same behavior simultaneously',
         async() => {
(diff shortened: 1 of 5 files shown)


src/test/analyzer_test.ts
```diff
@@ -523,4 +524,63 @@ suite('Analyzer', () => {
           testSeed.events.map(e => e.name), ['fired-event', 'data-changed']);
     });
   });
+
+  suite.skip('race conditions and caching', () => {
+    const wait = () =>
+        new Promise((resolve) => setTimeout(resolve, Math.random() * 30));
```
FredKSchott: -1 for randomness in tests, can you choose a number instead?

FredKSchott:
+1000 for adding breaking tests :)

src/test/analyzer_test.ts
rictic: Yeah, I hear you. I think this is still worthwhile to have around though, as a stress test of unexpected routes through our analysis and caching pipeline.

How about I add deterministic tests of all the three failures states that I noticed this test encountering?

FredKSchott: +1 for adding tests for those known failure states

The problem with a random test like this is that if a bug is introduced, it takes a while for the test to catch it and then takes even longer to track down where the bug was introduced since the test itself is unreliable. When it does start failing, we're much more likely to ignore it and just rerun until passing (ex: wct currently has a flakey test that's been ignored or rerun for months now).

So I'm still -1 on this test. At best it will never catch anything, and at worst we are introducing a flakey, slow-to-signal, hard-to-fix test.

src/test/analyzer_test.ts
```diff
@@ -140,6 +142,30 @@ suite('Analyzer', () => {
               chainedDocument.getWarnings(true), [expectedWarning]);
         });
 
+    // currently passing
```
FredKSchott: remove comment?

src/test/analyzer_test.ts
```diff
@@ -523,4 +549,193 @@ suite('Analyzer', () => {
...
   });
+
+  suite('race conditions and caching', () => {
+    test.skip('editor simulator of imports that import a common dep', async() => {
+      // Here we're simulating a lot of noop-changes to base.html, which has
+      // two imports, which mutually import a common dep. This stresses the
+      // analyzer's caching.
+
+      const wait = () =>
+          new Promise((resolve) => setTimeout(resolve, Math.random() * 30));
```
FredKSchott: Still don't love these tests, but with the deterministic tests below I'll survive. Could we at least log each `Math.random()` value we use so that we can know how big/small some of these wait times were if/when the test breaks?

src/test/analyzer_test.ts
```diff
@@ -523,4 +549,193 @@ suite('Analyzer', () => {
...
+
+      test.skip('two edits of the same file back to back', async() => {
+        const analyzer = new Analyzer({urlLoader: new NoopUrlLoader()});
+        await Promise.all([
+          analyzer.analyze('leaf.html', 'Hello'),
+          analyzer.analyze('leaf.html', 'World')
+        ]);
+      });
+
+      test.skip('something about the order of scanning?', async() => {
```
FredKSchott: :)

src/test/analyzer_test.ts
rictic: I refactored this code to log the wait times that are used in the case that it fails, and added a skipped test that uses such an array of wait times. Sadly, this didn't seem to reliably reproduce the same failure, it looks like, between the setTimeout scheduler's degrees of freedom and other system factors there's still enough nondeterminism to keep the failures from being deterministic. 

Still, it's better than nothing and may be useful the next time we're called back into this code, so I'll leave it for now and let us decide then whether it should be kept.

src/test/analyzer_test.ts
```diff
@@ -523,4 +546,211 @@ suite('Analyzer', () => {
...
+     * This is a tool for reproducing and debugging a failure of the editor
+     * simulator test above, but only at the exact same commit, as it's
+     * sensitive to the order of internal operations of the analyzer. So this
+     * code with a defined list of wait times should not be checked in.
+     *
+     * It's also worth noting that this code will be dependent on many other
+     * system factors, so it's only somewhat more reproducible, and may not end
+     * up being very useful. If it isn't, we should delete it.
+     */
+    test.skip('somewhat more reproducable editor simulator', async() => {
```
justinfagnani: might as well remove the test.skip if the whole suite is skipped

src/test/analyzer_test.ts
```diff
@@ -523,4 +546,211 @@ suite('Analyzer', () => {
...
+        ['common.html', `<custom-el></custom-el>`],
+      ]);
+      const analyzer =
+          new Analyzer({urlLoader: new RacyUrlLoader(contentsMap, waitFn)});
+      const promises: Promise<Document>[] = [];
+      for (let i = 0; i < 30; i++) {
+        await waitFn();
+        for (const key of contentsMap) {
+          if (Math.random() > 0.5) {
+            analyzer.analyze(key[0], key[1]);
```
justinfagnani: micro-nit 2: until we have destructuring, I like to assign entries into variables that describe what they are.

src/test/analyzer_test.ts
```diff
@@ -523,4 +546,211 @@ suite('Analyzer', () => {
...
+        ['a.html', `<link rel="import" href="common.html">`],
+        ['b.html', `<link rel="import" href="common.html">`],
+        ['common.html', `<custom-el></custom-el>`],
+      ]);
+      const analyzer =
+          new Analyzer({urlLoader: new RacyUrlLoader(contentsMap, waitFn)});
+      const promises: Promise<Document>[] = [];
+      for (let i = 0; i < 30; i++) {
+        await waitFn();
+        for (const key of contentsMap) {
```
justinfagnani: micro-nit: `key` -> `entry`

src/test/analyzer_test.ts
```diff
@@ -523,4 +546,211 @@ suite('Analyzer', () => {
           testSeed.events.map(e => e.name), ['fired-event', 'data-changed']);
     });
   });
+
+  suite.skip('race conditions and caching', () => {
+
+    class RacyUrlLoader implements UrlLoader {
+      constructor(
+          public map: Map<string, string>,
```
justinfagnani: nit: `map` -> `files`

src/test/analyzer_test.ts
```diff
@@ -523,4 +546,211 @@ suite('Analyzer', () => {
...
+          if (Math.random() > 0.5) {
+            analyzer.analyze(key[0], key[1]);
+          }
+        }
+        promises.push(analyzer.analyze('base.html'));
+      }
+      const documents = await Promise.all(promises);
+      for (const document of documents) {
+        const imports = Array.from(document.getByKind('import'));
+        assert.deepEqual(
```
justinfagnani: After the `map()` this looks like `assert.sameMembers()`. Might give a better error message.

src/test/analyzer_test.ts
```diff
@@ -523,4 +546,211 @@ suite('Analyzer', () => {
...
+        setTimeout(resolve, waitTime);
+      });
+      await editorSimulator(reproducableWait);
+    });
+
+    suite('deterministic tests', () => {
+      // Deterministic tests extracted from various failures of the above random
+      // test.
+
+      class Deferred<T> {
```
justinfagnani: thinking out loud: We really should have/import this somewhere. Would also be nice to add:

```typescript
get nodeCallback() {
  return (error, result) => {
    if (error) {
      this.reject(error)
    } else {
      this.resolve(result);
    }
  };
}
```

for manual promisification.

src/test/analyzer_test.ts
```diff
@@ -523,4 +546,211 @@ suite('Analyzer', () => {
...
+        resolve: (result: T) => void;
+        reject: (error: Error) => void;
+        constructor() {
+          this.promise = new Promise((resolve, reject) => {
+            this.resolve = resolve;
+            this.reject = reject;
+          });
+        }
+      }
+      class KeyedQueue<Key, Result> {
```
justinfagnani: can you add a small comment here about what this is for?

src/test/analyzer_test.ts
```diff
@@ -523,4 +546,211 @@ suite('Analyzer', () => {
...
+        private _requests = new Map<Key, Deferred<Result>[]>();
+        private _results = new Map<Key, Result[]>();
+        async request(key: Key): Promise<Result> {
```
justinfagnani: nit: add a newline

src/test/analyzer_test.ts
```diff
@@ -523,4 +546,211 @@ suite('Analyzer', () => {
...
+        }
+
+        toString() {
+          return JSON.stringify({
+            openRequests: Array.from(this._requests.keys()),
+            openResponses: Array.from(this._results.keys())
+          });
+        }
+      }
+      class DeterministicUrlLoader implements UrlLoader {
```
justinfagnani: nit: newline

src/test/analyzer_test.ts
```diff
@@ -523,4 +546,211 @@ suite('Analyzer', () => {
...
+        outstandingRequests = new KeyedQueue<string, string>();
+        canLoad(_url: string) {
+          return true;
+        }
+
+        async load(url: string) {
+          return this.outstandingRequests.request(url);
+        }
+      }
+      class NoopUrlLoader implements UrlLoader {
```
justinfagnani: nit: newline

src/test/analyzer_test.ts
```diff
@@ -523,4 +546,211 @@ suite('Analyzer', () => {
...
+        canLoad() {
+          return true;
+        }
+        async load(): Promise<string> {
+          throw new Error(
```
justinfagnani: if this throws, whould `canLoad`?

src/test/analyzer_test.ts
```diff
@@ -523,4 +546,211 @@ suite('Analyzer', () => {
...
+              `Noop Url Loader isn't supposed to be actually called.`);
+        }
+      }
+
+      test('two edits of the same file back to back', async() => {
+        const analyzer = new Analyzer({urlLoader: new NoopUrlLoader()});
```
justinfagnani: For regression tests on tricky bugs like this, it's nice to have a comment that describes how this was triggering a bug, or what the failure condition was, or maybe a link to an issue that this tests against.

justinfagnani:
Also, it'd be nice to squash/edit out the commit message with "todo".

rictic:
Edited the commit message to remove the `todo` part and clarify.

PR #384, opened 2016-11-04
Add common clang-format rules

FredKSchott:
- [X] CHANGELOG.md has been updated

***Only merge when there are no outstanding PRs, and remember to run `npm run format` after merging!***

/cc @rictic

PR #393, opened 2016-11-13
Format all code

rictic:

The formatting options introduced with #384 cause clang-format to produce invalid code with async methods whose signatures are a certain length. Specifically `BinPackParameters: false` will turn

```typescript
class C {
  async method(aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa, b, c) {
  }
}
```

into

```typescript
class C {
  async
  method(aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa, b, c) {
  }
}
```

which is semantically different code. I would report a bug upstream but account creation in their bug tracker is disabled: https://llvm.org/bugs/enter_bug.cgi

Given this, I'm leaning towards the idea that future changes that affect formatting should also include the results of `npm run format`. There will be merge conflicts, but resolving them with the `ours` strategy and then following with `npm run format` should make them trivial.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,5 +1,6 @@
 BasedOnStyle: Google
 AlignAfterOpenBracket: AlwaysBreak
+AllowAllParametersOfDeclarationOnNextLine: false
 AllowShortBlocksOnASingleLine: false
 AllowShortCaseLabelsOnASingleLine: false
 AllowShortFunctionsOnASingleLine: None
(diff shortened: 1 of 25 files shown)


FredKSchott:
+1, for both this PR and the idea of always including reformatting in the same PR as clang config changes

rictic:
Thanks for the review!

Issue #369, opened 2016-10-26
When a dependency of a Document changes, we must remove that Document from the cache.


rictic:
Suppose `a`, `b`, and `c` are documents, an `a` imports `b`, and `b` imports `c`. If you call `analyzer.analyze('a')`, then we will resolve and cache all three. 

If you then call `analyzer.analyzer('c', 'new contents')` then we must remove `a` and `b` from the `Document` cache, as they must be re-resolved.

TimvdLippe:
I think this is fixed in #394 but not autoclosed as it wasnt correctly referenced in the PR description

rictic:
Thanks Tim!

PR #394, opened 2016-11-14
Add forkable analysis caches for concurrency-safe analysis

rictic:

This should fix a number of issues that have arisen:
  * proper cache invalidation of dependant documents, fixes https://github.com/Polymer/polymer-editor-service/issues/21 and #369 
  * I believe that this will fix a number of errors that sound like race conditions, like https://github.com/Polymer/polymer-editor-service/issues/27 https://github.com/Polymer/polymer-editor-service/issues/28 https://github.com/Polymer/polymer-editor-service/issues/26 and https://github.com/Polymer/polymer-editor-service/issues/16

Note that this PR builds on top of a chain of other PRs


justinfagnani:
Can you rebase?

src/analysis-cache.ts
```diff
@@ -0,0 +1,169 @@
...
+    for (const keyValue of newCache.analyzedDocuments) {
+      newCache.analyzedDocumentPromises.set(
+          keyValue[0], Promise.resolve(keyValue[1]));
+    }
+
+    return newCache;
+  }
+
+  private _clone(): AnalysisCache {
+    const newCache = new AnalysisCache();
```
justinfagnani: I'd use a copy constructor instead so you can use the copy constructor of `Map`:

```javascript
constructor(from?: AnalysisCache) {
  this.parsedDocumentPromises = new Map(from && from.parsedDocumentPromises);
  // etc...
}
```

rictic:
Rebased. Note that this is merging into #390 which contains a WIP commit as I continue to hack on making deterministic tests for all of the different failure modes.

src/analysis-cache.ts
```diff
@@ -0,0 +1,169 @@
...
+    }
+  }
+}
+
+/**
+ * Trawl the import graph and return the paths of (transitive) dependants on
+ * the given path.
+ *
+ * This should eventually live somewhere like Document, but we really need
+ * something like a Project, which encapsulates all of the known Documents
```
justinfagnani: Originally I had an `AnalysisContext` which had all of this, but as a simplification I reduced that to just an `Analyzer`. `AnalysisCache` along with this as a method is very much like a context, and does know about all the documents, so maybe a rename would make the cache a good home for this function.

src/analysis-cache.ts
```diff
@@ -0,0 +1,169 @@
...
+    for (const importer of importers) {
+      if (!visited.has(importer)) {
+        toVisit.add(importer);
+      }
+    }
+  }
+  return visited;
+}
+
+function _buildInvertedIndex(
```
justinfagnani: Let's describe the index in a comment - what's this an index of?

src/analysis-cache.ts
```diff
@@ -0,0 +1,169 @@
...
+ * inside a basedir, and which has the same interface as Document. That is the
+ * object that the AnalysisCache needs to know the results of calling
+ * getImportersOf.
+ */
+export function getImportersOf(
+    path: string,
+    documents: Iterable<Document>,
+    scannedDocuments: Iterable<ScannedDocument>,
+    urlResolver: (url: string) => string): Set<string> {
+  const invertedIndex =
```
justinfagnani: how often is `getImportersOf` called on the same inputs? It'd be nice to preserve the index as much as we can. If it's part of the cache that may be possible - we can invalidate only part of the index.

src/analysis-cache.ts
```diff
@@ -0,0 +1,169 @@
...
+      _buildInvertedIndex(documents, scannedDocuments, urlResolver);
+  const visited = new Set<string>();
+  const toVisit = new Set<string>([path]);
+  while (toVisit.size > 0) {
+    const path = toVisit.keys().next().value!;
```
justinfagnani: interesting... Can you use `entries()` for clarify? `keys()` on `Set` is odd to read

src/analysis-cache.ts
```diff
@@ -0,0 +1,169 @@
...
+    urlResolver: (url: string) => string): Map<string, Set<string>> {
+  const invertedIndex = new Map<string, Set<string>>();
+  const docsSeen = new Set<string>();
+  for (const doc of docs) {
+    docsSeen.add(doc.url);
+    _addFeaturesToInvertedIndex(doc, invertedIndex);
+  }
+  for (const scannedDoc of scannedDocuments) {
+    if (docsSeen.has(scannedDoc.url)) {
+      continue;
```
justinfagnani: why `continue`? Does `_addFeaturesToInvertedIndex` also add scanned features?

src/analysis-cache.ts
```diff
@@ -0,0 +1,169 @@
...
+  }
+}
+
+function _addScannedFeaturesToInvertedIndex(
+    scannedDoc: ScannedDocument,
+    invertedIndex: Map<string, Set<string>>,
+    urlResolver: (url: string) => string) {
+  for (const feature of scannedDoc.features) {
+    if (feature instanceof ScannedImport) {
+      const imported = urlResolver(feature.url);
```
justinfagnani: why do you resolve this URL but not in `_addFeaturesToInvertedIndex`?

src/model/document.ts
```diff
@@ -12,7 +12,7 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
-import {Analyzer} from '../analyzer';
+import {AnalyzerCacheContext} from '../analyzer';
```
justinfagnani: where's this?

src/analysis-cache.ts
```diff
@@ -0,0 +1,169 @@
...
+  dependenciesScanned = new Map<string, Promise<void>>();
+
+  scannedDocuments = new Map<string, ScannedDocument>();
+  analyzedDocuments = new Map<string, Document>();
+
+  /**
+   * Returns a copy of this cache, with the given path and all of its transitive
+   * dependants invalidated.
+   */
+  onPathChanged(path: string, dependants: Iterable<string>): AnalysisCache {
```
rictic: Because the AnalysisCache doesn't manage the dependency graph, so it doesn't have this information. It needs it because you need to invalidate the post-resolution caches of everything that depends on `path`

FredKSchott: How much complexity does this method add to the rest of the cache/analyzer? My bet is that paths are changed very rarely, and it could reduce a ton of complexity around tracking these different things if we just cleared the cache and rescanned whenever a path is changed.

src/analysis-cache.ts
FredKSchott: +1, I'm having trouble understanding these functions without knowing what your intention is with index, inverted index, etc.

rictic:
RE the complexity of tracking changes to files vs just rescanning everything (because I can't reply on an old line of code??): 

For the builder case that would be fine, but for the editor use case we get a lot of performance out of only rescanning and reanalyzing the files that have changed.

FredKSchott:
@rictic yea github is the worst :( I understand that the performance benefits are significant for the real-time use case, but I imagine path changes are a very rare event. And if a user is changing paths, they're not typing code at the same time so the analyzer should have some extra time to catch up. 

Since you've already built this it's great that we can now handle it, but I only brought it up in case this will add significant complexity to the rest of the analyzer.

rictic:
Oh, `path` there is the path to a document, so the meaning is that the document at that path changed. I'll rename the method.

src/model/document.ts
rictic: It was in `../analyzer` alongside `Analyzer` but that's not super clean. I broke it out into its own file and created a new directory `core` for code that implements the core abstract pieces of the analyzer.

rictic:
Sorry for the commit spam. I broke it down into small pieces so that it's easier to review.

On the plus side this fixes a few subtle issues with the way that we broke up the parse -> scan -> resolve phases.

justinfagnani:
Ugh, github. I can't respond to your comments :(

src/core/analyzer-cache-context.ts
```diff
@@ -0,0 +1,483 @@
...
+   */
+  private async _scan(
+      resolvedUrl: string, contents?: string,
+      visited?: Set<string>): Promise<ScannedDocument|'visited'> {
+    if (visited && visited.has(resolvedUrl)) {
+      return 'visited';
+    }
+    visited = visited || new Set();
+    visited.add(resolvedUrl);
+    const cachedResult = this._cache.scannedDocumentPromises.get(resolvedUrl);
```
justinfagnani: why not await here?

src/model/document.ts
```diff
@@ -48,6 +48,21 @@ export class ScannedDocument {
   get url() {
     return this.document.url;
   }
+
+  getNestedFeatures(): ScannedFeature[] {
```
justinfagnani: what are "nested" features?

justinfagnani: could take an optional argument which is a features array, then just push into the array as you recurse.

src/model/import.ts
```diff
@@ -30,12 +30,10 @@ export class ScannedImport implements Resolvable {
   type: 'html-import'|'html-script'|'html-style'|'js-import'|string;
 
   /**
-   * URL of the import, relative to the document containing the import.
+   * URL of the import, relative to the base directory.
```
justinfagnani: I think we may want two fields here. `url` used to represent what was actually in the import feature.

src/core/analysis-cache.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  parsedDocumentPromises: Map<string, Promise<ParsedDocument<any, any>>>;
+  scannedDocumentPromises: Map<string, Promise<ScannedDocument>>;
+  analyzedDocumentPromises: Map<string, Promise<Document>>;
+
+  /**
+   * This is a map from a resolved url to a promise that will resolve when
+   * that document's dependencies have been scanned.
+   *
+   * We need to keep track of this separate from just the scanned document
```
rictic: If we combined the "scan dependencies" and "resolve" steps then this cache wouldn't be necessary, but this PR already seems complex enough, and it seems better to punt it.

src/core/analysis-cache.ts
```diff
@@ -0,0 +1,106 @@
...
+   * the document itself, but we do need to rescan its dependencies.
+   */
+  dependenciesScanned: Map<string, Promise<void>>;
+
+  scannedDocuments: Map<string, ScannedDocument>;
+  analyzedDocuments: Map<string, Document>;
+
+  dependencyGraph: DependencyGraph;
+
+  constructor(from?: AnalysisCache, newDependencyGraph?: DependencyGraph) {
```
rictic: Done

src/core/analysis-cache.ts
```diff
@@ -0,0 +1,106 @@
...
+    let f: Partial<AnalysisCache> = from || {};
```
rictic: Done

src/core/analysis-cache.ts
```diff
@@ -0,0 +1,106 @@
...
+    this.dependencyGraph = newDependencyGraph || new DependencyGraph();
+  }
+
+  /**
+   * Returns a copy of this cache, with the given document and all of its
+   * transitive dependants invalidated.
+   *
+   * Must be called whenever a document changes.
+   */
+  invalidatePaths(documentPaths: string[]): AnalysisCache {
```
rictic: Renamed to invalidate

src/model/document.ts
rictic: Documented

rictic: Refactored into a public method that takes no arguments and a private method that works like this

src/model/import.ts
rictic: +1, but punting. This PR doesn't change this behavior I don't think, it just documents the current behavior.

src/core/analyzer-cache-context.ts
```diff
@@ -0,0 +1,483 @@
...
+    copy._telemetryTracker = this._telemetryTracker;
+    copy._cache = cache;
+    copy._generation = this._generation + 1;
+    return copy;
+  }
+
+  /**
+   * Scan a toplevel document given its url and optionally its contents.
+   */
+  private async _scan(
```
rictic: Updated documentation to emphasize what it does more and less on its arguments, which seem fairly clear.

src/core/analyzer-cache-context.ts
rictic: Oh, yeah that's easier to reason about, +1.

And +1 to typescript for correctly determining that `Promise<T> | undefined` awaits to `T | undefined`

src/core/analyzer-cache-context.ts
```diff
@@ -0,0 +1,483 @@
...
+            'Scanned document already in cache. This should never happen.');
+      }
+      this._cache.scannedDocuments.set(scannedDocument.url, scannedDocument);
+    }
+    await this._scanInlineDocuments(scannedDocument, visited);
+
+    return scannedDocument;
+  }
+
+  private async _scanDependenciesOfToplevelDoc(
```
rictic: +1. Added a bunch of docs and also did some mild reorganization for clarity.

The organization of code into private methods here had gotten really spread out and hard to follow, but thanks to other changes in this PR I was able to simplify through a bit of consolidation.

src/core/analyzer-cache-context.ts
```diff
@@ -0,0 +1,483 @@
...
+      scannedDocument: ScannedDocument, visited: Set<string>) {
+    const scanDepPromise =
+        this._cache.dependenciesScanned.get(scannedDocument.url);
+    if (scanDepPromise) {
+      return scanDepPromise;
+    }
+    const promise = (async() => {
+      await Promise.resolve();
```
rictic: Done

PR #396, opened 2016-11-19
Fix an unhandled promise rejection warning.

rictic:
In a future version of node, this will kill the process, even though we've handled the error through other means.

(mea culpa: I think I asked for the promise to reject back when the scan() function was originally checked in)

### DIFF
--- a/src/scanning/scan.ts
+++ b/src/scanning/scan.ts
@@ -17,9 +17,8 @@ import {ParsedDocument} from '../parser/document';
 
 import {Scanner} from './scanner';
 
-export async function scan(
-    document: ParsedDocument<any, any>,
-    scanners: Scanner<any, any, any>[]): Promise<ScannedFeature[]> {
+export async function scan<A, V, P extends ParsedDocument<A, V>>(
+    document: P, scanners: Scanner<P, A, V>[]): Promise<ScannedFeature[]> {
   // Scanners register a visitor to run via the `visit` callback passed to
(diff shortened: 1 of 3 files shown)


src/scanning/scan.ts
```diff
@@ -66,12 +65,11 @@ export async function scan(
   };
 
   // The callback passed to `scan()`
-  function visit(visitor: any) {
+  function visit(visitor: V) {
     visitors.push(visitor);
     if (!runner) {
       runner = Promise.resolve().then(runVisitors).catch((error) => {
         visitError = visitError || error;
-        throw error;
```
rictic: This is the only change that matters much. The rest were just for my own benefit.

src/scanning/scan.ts
```diff
@@ -17,9 +17,8 @@ import {ParsedDocument} from '../parser/document';
 
 import {Scanner} from './scanner';
 
-export async function scan(
-    document: ParsedDocument<any, any>,
-    scanners: Scanner<any, any, any>[]): Promise<ScannedFeature[]> {
+export async function scan<A, V, P extends ParsedDocument<A, V>>(
```
FredKSchott: Can you document these `A`, `V`, `P` types? No idea what is happening here 😕

rictic: They're just passed in through to ParsedDocument. I improved the docs on ParsedDocument.

PR #397, opened 2016-11-19
Add slots to PolymerElement

### LINKED ISSUE #391: Add insertionPoints on Element when parsing templates inside dom modules
For https://github.com/Polymer/polymer-editor-service/pull/29 an enhancement that I could not implement yet was autocompletion for insertionPoints.

E.g.

```html
<template>
  <slot name="my-slot1"></slot>
  <slot name="my-slot2"></slot>
</template>
```

When analyzed by the analyzer returns in `Element` an array `insertionPoints`:
```js
insertionPoints: ['my-slot1', 'my-slot2']
```

TimvdLippe:
This PR is now finished and fixes #391 

Please see the corresponding PR at https://github.com/Polymer/polymer-editor-service/pull/34 which uses this feature.

### DIFF
--- a/src/model/element.ts
+++ b/src/model/element.ts
@@ -41,6 +41,7 @@ export class ScannedElement implements Resolvable {
   sourceRange: SourceRange|undefined;
   astNode: estree.Node|null;
   warnings: Warning[] = [];
+  slots: Slot[] = [];
 
   jsdoc?: jsdoc.Annotation;
 
@@ -58,6 +59,16 @@ export class ScannedElement implements Resolvable {
 
 export interface Attribute extends ScannedAttribute { inheritedFrom?: string; }
 
+export class Slot {
+  name: string;
+  range: SourceRange;
+
+  constructor(name: string, range: SourceRange) {
+    this.name = name;
+    this.range = range;
+  }
+}
+
 export class Element implements Feature {
   tagName?: string;
   className?: string;
@@ -73,6 +84,7 @@ export class Element implements Feature {
   astNode: estree.Node|null;
   kinds: Set<string> = new Set(['element']);
   warnings: Warning[] = [];
+  slots: Slot[] = [];
   get identifiers(): Set<string> {
     const result: Set<string> = new Set();
     if (this.tagName) {
--- a/src/polymer/dom-module-scanner.ts
+++ b/src/polymer/dom-module-scanner.ts
@@ -13,11 +13,11 @@
  */
 
 import * as dom5 from 'dom5';
-import {ASTNode} from 'parse5';
+import {ASTNode, treeAdapters} from 'parse5';
 
 import {HtmlVisitor, ParsedHtmlDocument} from '../html/html-document';
 import {HtmlScanner} from '../html/html-scanner';
-import {Feature, getAttachedCommentText, Resolvable, SourceRange} from '../model/model';
+import {Feature, getAttachedCommentText, Resolvable, SourceRange, Slot} from '../model/model';
 import {Warning} from '../warning/warning';
 
 const p = dom5.predicates;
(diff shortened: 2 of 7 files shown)


src/polymer/polymer-element.ts
```diff
@@ -83,6 +83,7 @@ export class ScannedPolymerElement extends ScannedElement {
   //     level. Remove them here, they should only exist on PolymerElement.
   domModule?: dom5.Node;
   scriptElement?: dom5.Node;
+  slots: Slot[];
```
rictic: Ideally `slots` would live on ScannedElement and Element, as it's not a polymer specific thing (even if we only know how to recognize slots on polymer elements for the moment)

src/polymer/dom-module-scanner.ts
```diff
@@ -87,11 +93,15 @@ export class DomModuleScanner implements HtmlScanner {
 
     await visit((node) => {
       if (isDomModule(node)) {
+        const slots = dom5.queryAll(node, dom5.predicates.hasTagName('slot'))
```
rictic: Nitpick, but I think it's more correct to find the first `<template>` element that is a direct child of `node` and do this queryAll over that.

src/polymer/polymer-element.ts
rictic: Also the issue you're seeing with it being undefined is due it not giving it a default value. I wish the type system caught this, but apparently it does not.

So:

```typescript
slots: Slot[] = [];
```

src/model/document.ts
```diff
@@ -26,6 +26,14 @@ import {Import} from './import';
 import {isResolvable} from './resolvable';
 import {SourceRange} from './source-range';
 
+export class Slot {
+  name: string;
+
+  constructor(name: string) {
```
rictic: I think `name` is optional, and represents a slot that will capture all light children that aren't captured by another slot

TimvdLippe:
@rictic Ready for review :tada:

src/polymer/dom-module-scanner.ts
```diff
@@ -87,11 +93,20 @@ export class DomModuleScanner implements HtmlScanner {
 
     await visit((node) => {
       if (isDomModule(node)) {
+        const template = dom5.query(node, dom5.predicates.hasTagName('template'));
+        let slots: Slot[] = [];
+        if (template) {
+          slots = dom5.queryAll(
+            treeAdapters.default.getTemplateContent(template),
+            dom5.predicates.hasTagName('slot')
+          ).map(s => new Slot(dom5.getAttribute(s, 'name') || ''));
```
rictic: Oh nice catch! Can confirm that a slot name of `''` is the same as not providing a name attribute at all.

src/polymer/polymer-element.ts
```diff
@@ -207,6 +207,7 @@ function resolveElement(
   if (domModule) {
     clone.description = scannedElement.description || domModule.comment || '';
     clone.domModule = domModule.node;
+    clone.slots = domModule.slots;
```
rictic: may not be a bad idea to do a defensive copy here: `domModule.slots.slice()`

src/model/element.ts
```diff
@@ -58,6 +59,14 @@ export class ScannedElement implements Resolvable {
 
 export interface Attribute extends ScannedAttribute { inheritedFrom?: string; }
 
+export class Slot {
+  name: string;
+
```
rictic: Recommend adding a sourceRange property, that way we can do jump to definition when someone does `<div slot="foo"></div>`, as well as giving the linter more hooks to work with too.

`ParsedHtmlDocument#getSourceRangeForNode` should do all of the heavy lifting.

src/polymer/dom-module-scanner.ts
```diff
@@ -87,11 +93,20 @@ export class DomModuleScanner implements HtmlScanner {
 
     await visit((node) => {
       if (isDomModule(node)) {
+        const template = dom5.query(node, dom5.predicates.hasTagName('template'));
+        let slots: Slot[] = [];
+        if (template) {
+          slots = dom5.queryAll(
+            treeAdapters.default.getTemplateContent(template),
+            dom5.predicates.hasTagName('slot')
+          ).map(s => new Slot(dom5.getAttribute(s, 'name') || '', document.sourceRangeForNode(s)));
```
rictic: The only way `document.sourceRangeForNode(s)` would return undefined here I think is if `s` wasn't originally from `document`, or if `document` wasn't parsed with location annotations turned on. We can be confident that neither was the case, so you can cast this to definitively non-null with `!` and make `slot.range` non optional as well

PR #401, opened 2016-11-27
Track memory usage and leaking in the benchmark.

rictic:
- [x] Internal change, CHANGELOG.md not updated.

### DIFF
--- a/src/perf/parse-all-benchmark.ts
+++ b/src/perf/parse-all-benchmark.ts
@@ -71,6 +71,11 @@ function memUsed(usage?: number) {
 }
 
 async function measure() {
+  if (!global.gc) {
+    throw new Error(
+        'This benchmark must be run with node --expose-gc.\n' +
+        '      Just do:\n             npm run benchmark');
+  }
   global.gc();
   const initialMemUse = process.memoryUsage().rss;
   console.log(`Initial rss: ${memUsed(initialMemUse)}`);
--- a/src/perf/parse-all-benchmark.ts
+++ b/src/perf/parse-all-benchmark.ts
@@ -63,12 +63,25 @@ function padLeft(str: string, num: number): string {
   return str;
 }
 
+function memUsed(usage?: number) {
+  if (usage == null) {
+    usage = process.memoryUsage().rss;
+  }
+  return `${(usage / (1024 * 1024)).toFixed(1)}MiB`;
+}
+
 async function measure() {
+  global.gc();
+  const initialMemUse = process.memoryUsage().rss;
+  console.log(`Initial rss: ${memUsed(initialMemUse)}`);
   const start = now();
   let document: any;
   for (let i = 0; i < 10; i++) {
     document = await analyzer.analyze('ephemeral.html', fakeFileContents);
   }
+
+  global.gc();
+  const afterInitialAnalyses = process.memoryUsage().rss;
 
   const measurements = await analyzer.getTelemetryMeasurements();
   printMeasurements(measurements);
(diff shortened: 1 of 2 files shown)


src/perf/parse-all-benchmark.ts
```diff
@@ -63,19 +63,51 @@ function padLeft(str: string, num: number): string {
...
+  const afterInitialAnalyses = process.memoryUsage().rss;
+
   const measurements = await analyzer.getTelemetryMeasurements();
   printMeasurements(measurements);
 
   console.log(`\n\n\n${document.getFeatures().size} total features resolved.`);
   console.log(
       `${((now() - start) / 1000).toFixed(2)} seconds total elapsed time`);
+
+  for (let i = 0; i < 100; i++) {
```
FredKSchott: It seems like we're testing two different things here: speed & memory leaks. Can we split into two different functions and add a way to test each individually?

src/perf/parse-all-benchmark.ts
```diff
@@ -63,19 +63,51 @@ function padLeft(str: string, num: number): string {
...
+      `Additional memory used in analyzing all Polymer-owned code: ${memUsed(
+          afterInitialAnalyses - initialMemUse)}`);
+  const leakedMemory = afterMoreAnalyses - afterInitialAnalyses;
+  console.log(
+      `Additional memory used after 100 more incremental analyses: ${memUsed(
+          afterMoreAnalyses - afterInitialAnalyses)}`);
+  // TODO(rictic): looks like we've got a memory leak. Need to track this down.
+  //   This should be < 10MiB, not < 100 MiB.
+  if (leakedMemory > 100 * (1024 * 1024)) {
+    process.exit(1);
```
FredKSchott: can we log something here so that someone seeing this test run knows why it failed? You can throw an error with a message, instead of calling `.exit(1)`.

src/perf/parse-all-benchmark.ts
```diff
@@ -63,19 +63,51 @@ function padLeft(str: string, num: number): string {
...
 
+function memUsed(usage?: number) {
+  if (usage == null) {
+    usage = process.memoryUsage().rss;
+  }
+  return `${(usage / (1024 * 1024)).toFixed(1)}MiB`;
+}
+
 async function measure() {
+  global.gc();
```
FredKSchott: Does this break if `--expose-gc` is not enabled? If fails silently or with a hard to understand error, it would be nice to add a check here to make sure that `--expose-gc` is enabled before running.

src/perf/parse-all-benchmark.ts
justinfagnani: In general I probably agree here, but I can see that for know the test would essentially be the same except the gc() calls and memory calcs. If we have different scenarios to run where one stresses memory but the other stresses throughput, then I'd definitely separate, otherwise it might make the single scenario harder to update.

src/perf/parse-all-benchmark.ts
justinfagnani: +1

src/perf/parse-all-benchmark.ts
rictic: +1

src/perf/parse-all-benchmark.ts
rictic: Good catch, done

rictic:
All comments addressed, PTAL

src/perf/parse-all-benchmark.ts
FredKSchott: 👍 I don't feel too strongly here, sounds good

src/perf/parse-all-benchmark.ts
```diff
@@ -63,19 +63,56 @@ function padLeft(str: string, num: number): string {
...
+          afterInitialAnalyses - initialMemUse)}`);
+  const leakedMemory = afterMoreAnalyses - afterInitialAnalyses;
+  console.log(
+      `Additional memory used after 100 more incremental analyses: ${memUsed(
+          afterMoreAnalyses - afterInitialAnalyses)}`);
+  // TODO(rictic): looks like we've got a memory leak. Need to track this down.
+  //   This should be < 10MiB, not < 100 MiB.
+  const threshold = 100 * (1024 * 1024);
+  if (leakedMemory > threshold) {
+    console.error(
```
FredKSchott: output nit: after seeing the test failure in travis, it took me a while to find this relevant line in the long list of test output (it looked like regular output and not an error). It would be nice if it was highlighted a little more. Even just prepending with 'Error: ' would have helped me (since we're all used to looking for those in throw exceptions).

justinfagnani:
Yeah, fix the tests then it's good to go.

src/perf/parse-all-benchmark.ts
rictic: +1 Clarified and bumped up the threshold. Looks like memory usage is noticeably higher in newer versions of node

PR #402, opened 2016-11-30
Add notify field to Attribute

TimvdLippe:
Adds the notify field from the scanned properties to attribute.

### DIFF
--- a/src/polymer/polymer-element.ts
+++ b/src/polymer/polymer-element.ts
@@ -113,7 +113,7 @@ export class ScannedPolymerElement extends ScannedElement {
         sourceRange: prop.sourceRange,
         description: prop.description,
         type: prop.type,
-        changeEvent: prop.notify ? `${attributeName}-changed` : ''
+        changeEvent: prop.notify ? `${attributeName}-changed` : undefined
       });
     }
     if (prop.notify) {
--- a/src/polymer/polymer-element.ts
+++ b/src/polymer/polymer-element.ts
@@ -113,6 +113,7 @@ export class ScannedPolymerElement extends ScannedElement {
         sourceRange: prop.sourceRange,
         description: prop.description,
         type: prop.type,
+        changeEvent: prop.notify ? `${attributeName}-changed` : ''
       });
     }
     if (prop.notify) {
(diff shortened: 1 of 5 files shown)


rictic:
`notify` is a Polymer-specific thing, I'd like to keep it on the polymer specific models.

Hm. The corresponding generic custom elements concept is that some attributes and properties fire an event when they're changed, and the event isn't always named `${prop.name}-changed`, e.g. `<input>` fires the `input` event when its `value` changes.

So maybe a `changeEvent` string field on Attribute and Property, and we can autogenerate this for Polymer attributes and properties based on `notify`.

That way when we do autocompletion for Polymer databinding based on this we'll have the information needed to get the `<input>` case right (assuming that we generate Element objects for built-in elements, maybe based on mdn (something I've been wanting to do for a long time anyways))

src/polymer/polymer-element.ts
```diff
@@ -113,6 +113,7 @@ export class ScannedPolymerElement extends ScannedElement {
         sourceRange: prop.sourceRange,
```
rictic: Recommend instead something like:

```typescript
const attribute = {...};
if (prop.notify) {
  attribute.changeEvent = `${attributeName}-changed`;
}
this.attributes.push(attribute);

File: src/test/polymer/polymer-element-scanner_test.ts [hunk 2]
```diff
@@ -116,6 +116,17 @@ suite('PolymerElementScanner', () => {
...
       ]);
 
+      assert.deepEqual(features[0].attributes.map(p => [p.name, p.changeEvent]), [
+        ['a', undefined],
+        ['b', undefined],
+        ['c', undefined],
+        ['d', undefined],
+        ['e', 'e-changed'],
+        ['f', undefined],
+        ['g', undefined],
```
rictic: Looks like the test is finding `['h', undefined],` as well? Which is kinda weird because I don't see an 'h' property defined anywhere

TimvdLippe: Oh oops that's because I did not rebase this branch onto master. Since my other PR got merged which fixed short-hand property notation

src/polymer/polymer-element.ts
rictic: Actually, never mind. V8 will be happier with more consistent object shapes.

PR #403, opened 2016-11-30
Correctly parse shorthand property definition

### LINKED ISSUE #357: Does not understand shorthand property definitions
```
properties: {
    // detected
    propA: {
        type: String
    },
    // not detected
    propB: String
}
```

The only test I could find for this is https://github.com/Polymer/polymer-analyzer/blob/master/src/test/polymer/polymer2-element-scanner_test.ts which does not seem to account for `prop: type`-style property definitions.

TimvdLippe:
Small fix to correctly parse shorthand property definitions.

Fixes #357

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 * Bump JS parser up to parse ES2017 (ES8).
-* Correctly parse shorthand property definitions.
+* [Polymer] Property descriptors are allowed to be just a type name, like `value: String`.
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 * Bump JS parser up to parse ES2017 (ES8).
+* Correctly parse shorthand property definitions.
```
rictic: This isn't clear to a reader that doesn't already know about the bug I don't think. What about this:

```md
* [Polymer] Property descriptors are allowed to be just a type name, like `value: String`.
```

PR #405, opened 2016-12-07
Use ts 2.1 features to express Kind types.

rictic:


rictic:
The advantage here is that we aren't repeating the mapping from kind to type (which is error prone, there were more than a couple errors in the existing code).

It is more abstract though, so I should probably add some more comments.

FredKSchott:
Oh yea you're right, I've been in that file a bunch now and only now just saw those errors. tbh I didn't even realize that they were all exact copies of each other.

I'm convinced, LGTM

src/model/document.ts
```diff
@@ -50,6 +50,29 @@ export class ScannedDocument {
   }
 }
 
+// A map between kind string literal types and their feature types.
+export interface FeatureKinds {
```
justinfagnani: I think ultimately this should be in a different file and/or and split up. PolymerElement should extend FeatureKinds with:

```typescript
declare interface FeatureKinds {
  'polymer-element': PolymerElement;
}
```
etc.

rictic: Strong agreement. I experimented with this a bit but didn't get interface merging working right, and I didn't want to spend much time on this.

PR #407, opened 2016-12-12
Properly handle 'listeners' property when scanning polymer element...

MaKleSoft:
... as described [here](https://www.polymer-project.org/1.0/docs/devguide/events#event-listeners). Before, it was treated like any other property. Fixes https://github.com/Polymer/polymer-analyzer/issues/285

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
    this comment.
 -->
 
+### Added
+* [Polymer] Extract 'listeners' from 1.0-style declarations.
+
 ## [2.0.0-alpha.19] - 2016-12-12
 
 ### Added
--- a/src/polymer/declaration-property-handlers.ts
+++ b/src/polymer/declaration-property-handlers.ts
@@ -74,10 +74,11 @@ export function declarationPropertyHandlers(
         return;
       }
       for (let prop of node.properties) {
-        if (prop.key.type !== 'Literal' || prop.value.type !== 'Literal') {
-            continue;
+        if (prop.key.type !== 'Literal' || typeof prop.key.value !== 'string' ||
+            prop.value.type !== 'Literal' || typeof prop.value.value !== 'string') {
+          continue;
         }
-        declaration.listeners.push({event: <string>prop.key.value, handler: <string>prop.value.value});
+        declaration.listeners.push({event: prop.key.value, handler: prop.value.value});
       }
     }
   };
--- a/src/polymer/declaration-property-handlers.ts
+++ b/src/polymer/declaration-property-handlers.ts
@@ -68,6 +68,17 @@ export function declarationPropertyHandlers(
         }
         declaration.observers.push({javascriptNode: element, expression: v});
       }
+    },
+    listeners(node: estree.Node) {
+      if (node.type !== 'ObjectExpression') {
+        return;
+      }
+      for (let prop of node.properties) {
+        if (prop.key.type !== 'Literal' || prop.value.type !== 'Literal') {
+            continue;
+        }
+        declaration.listeners.push({event: <string>prop.key.value, handler: <string>prop.value.value});
+      }
     }
   };
 }
--- a/src/polymer/declaration-property-handlers.ts
+++ b/src/polymer/declaration-property-handlers.ts
@@ -19,6 +19,7 @@ import {JavaScriptDocument} from '../javascript/javascript-document';
 
 import {analyzeProperties} from './analyze-properties';
 import {ScannedPolymerElement} from './polymer-element';
+import {Severity} from '../warning/warning';
 
 export type PropertyHandlers = {
   [key: string]: (node: estree.Node) => void
@@ -71,11 +72,23 @@ export function declarationPropertyHandlers(
     },
     listeners(node: estree.Node) {
       if (node.type !== 'ObjectExpression') {
+        declaration.warnings.push({
+          code: 'invalid-listeners-declaration',
+          message: '`listeners` property should be an object expression',
+          severity: Severity.ERROR,
+          sourceRange: document.sourceRangeForNode(node)!
+        });
         return;
       }
       for (let prop of node.properties) {
         if (prop.key.type !== 'Literal' || typeof prop.key.value !== 'string' ||
             prop.value.type !== 'Literal' || typeof prop.value.value !== 'string') {
+          declaration.warnings.push({
+            code: 'invalid-listeners-declaration',
+            message: 'handler value should be a string',
+            severity: Severity.ERROR,
+            sourceRange: document.sourceRangeForNode(prop.value)!
+          });
           continue;
         }
         declaration.listeners.push({event: prop.key.value, handler: prop.value.value});
--- a/src/polymer/declaration-property-handlers.ts
+++ b/src/polymer/declaration-property-handlers.ts
@@ -71,6 +71,7 @@ export function declarationPropertyHandlers(
       }
     },
     listeners(node: estree.Node) {
+
       if (node.type !== 'ObjectExpression') {
         declaration.warnings.push({
           code: 'invalid-listeners-declaration',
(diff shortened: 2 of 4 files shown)


src/polymer/declaration-property-handlers.ts
```diff
@@ -68,6 +68,17 @@ export function declarationPropertyHandlers(
         }
         declaration.observers.push({javascriptNode: element, expression: v});
       }
+    },
+    listeners(node: estree.Node) {
+      if (node.type !== 'ObjectExpression') {
+        return;
+      }
+      for (let prop of node.properties) {
+        if (prop.key.type !== 'Literal' || prop.value.type !== 'Literal') {
```
rictic: Should also early exit if the `typeof` the literal values aren't strings rather than casting to string below. I think this will work and let you remove the casts:

```typescript
if (prop.key.type !== 'Literal' || typeof prop.key.value !== 'string' || prop.value.type !== 'Literal' || typeof prop.value.value !== 'string') {
  continue;
}
```

`npm run format` will handle the whitespace

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
    this comment.
 -->
 
+* Properly handle 'listeners' property when scanning polymer elements
```
rictic: Recommend:

```
* [Polymer] Extract 'listeners' from 1.0-style declarations.
```

src/polymer/declaration-property-handlers.ts
rictic: (Optional, for all the bonus points: add Warnings to `document` when any of these constraints don't hold)

src/polymer/polymer-element.ts
```diff
@@ -60,6 +60,10 @@ export interface Options {
     javascriptNode: estree.Expression | estree.SpreadElement,
     expression: LiteralValue
   }[];
+  listeners?: {
+    event: string,
+    handler: string
```
TimvdLippe: Please also add a `sourceRange` for jump-to-definition

src/polymer/polymer-element.ts
```diff
@@ -96,6 +104,7 @@ export class ScannedPolymerElement extends ScannedElement {
     if (options.properties) {
       options.properties.forEach((p) => this.addProperty(p));
     }
+    this.listeners = this.listeners || [];
```
TimvdLippe: To make it consistent with the other lists, just initialize it at the declaration (see `Slot[]`)

src/test/polymer/polymer-element-scanner_test.ts
```diff
@@ -73,7 +75,10 @@ suite('PolymerElementScanner', () => {
         observers: [
           '_anObserver()',
           '_anotherObserver()'
-        ]
+        ],
+        listeners: {
+          'event-a': '_handleA'
```
TimvdLippe: Add 1 more event to make sure it handles multiple? Also maybe discard bogus events

src/polymer/declaration-property-handlers.ts
```diff
@@ -68,6 +68,17 @@ export function declarationPropertyHandlers(
...
+            continue;
+        }
+        declaration.listeners.push({event: <string>prop.key.value, handler: <string>prop.value.value});
```
TimvdLippe: Are these typecasts required?

src/polymer/polymer-element.ts
rictic: What's the jump-to-definition use case? Jumping from the handler name to the actual method?

It would be good to validate the connection between the handler name and the corresponding method on the element, and if we did _that_ then I could see adding the source range of the method, but that's a big enough addition to this PR that it'd also be reasonable to punt that to another PR.

OTOH it would be pretty awesome.

src/test/polymer/polymer-element-scanner_test.ts
MaKleSoft: What do you consider a bogus event?

src/polymer/polymer-element.ts
TimvdLippe: I was thinking of adding autocompletion of listener values. E.g. `foo: '|'` autocompleting to all available methods in the current element.

For that I assume `getLocationInfoForPosition` needs to know the sourceRange of the block, to know whether the user currently wants to autcomplete listeners.

The comment is not a blocker, just something I was thinking of implementing.

src/test/polymer/polymer-element-scanner_test.ts
TimvdLippe: Seems you already updated the test, exactly what I meant nice!

web-padawan:
It would be nice to have some check for using `id` of element which is located inside of the `dom-if` template in `listeners` object. This causes errors like this: 
```
polymer.html:540 Uncaught TypeError: Invalid value used as weak map key
```
Maybe this check should be implemented in `polymer-linter` as well, even though 1.0 syntax is going to become deprecated.
@rictic WDYT?

TimvdLippe:
@web-padawan That check should be implemented in `polymer-linter`, but this PR is required to actually retrieve the data required for the linter to potentially report warnings about misuse of listeners.

web-padawan:
@TimvdLippe thanks, got it.

src/polymer/declaration-property-handlers.ts
```diff
@@ -71,11 +72,23 @@ export function declarationPropertyHandlers(
...
       }
       for (let prop of node.properties) {
         if (prop.key.type !== 'Literal' || typeof prop.key.value !== 'string' ||
             prop.value.type !== 'Literal' || typeof prop.value.value !== 'string') {
+          declaration.warnings.push({
+            code: 'invalid-listeners-declaration',
+            message: 'handler value should be a string',
+            severity: Severity.ERROR,
+            sourceRange: document.sourceRangeForNode(prop.value)!
+          });
```
rictic: Awesome!

One last change and I think this is good to go:

It's possible that `prop.key` isn't a Literal or that literal isn't a string. consider:

```javascript
const propName = 'foo';
Polymer({
  listeners: {
    [propName + '-changed']: 'handlePropChanged'
  }
});
```

As written this would give a warning on the 'handlePropName' string literal, not the computed object key, which is where the issue is. I think we'll need to split this out into two conditionals to get it right, one for the key and one for the value.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 * Add `changeEvent` to `Attribute`.
+* [Polymer] Extract 'listeners' from 1.0-style declarations.
```
rictic: Minor, I can do this when it's merged, but if you get the chance to rebase, this will need to move up to the `## Unreleased` section again, as we stamped a new release last night.

src/polymer/declaration-property-handlers.ts
MaKleSoft: I'm guessing simple identifiers should still be allowed though, right?
```js
listeners: {
  changed: 'changedHandler'
}
```
How should we treat computed keys? Strictly speaking they should be valid, correct? So should we just skip the entry without a warning or consider a code smell and add a warning with reduced severity? We probably can't extract the event name in that case. Or is there an easy way to evaluate the expression?

rictic: Ah, yeah, identifiers should be allowed for keys. I thought that that syntax was recognized as a string literal (as semantically it's a string literal, just one that's parsed with identifier syntax).

And yeah, you're right, we should probably only generate warnings by default on stuff that's definitely wrong, rather than stuff that's just impossible to statically analyze. (As an eventual goal it would be nice if people could opt in to warnings on non-statically analyzable code, as some people would want to enforce that in their codebases, but that will require more advanced Warning-twiddling, as well as an IGNORE severity, which both are out of scope of this change.)

So I guess the right thing to do here is to handle string literals or identifiers for keys, and only string literals for values, and if it's anything else just continue without emitting a warning.

MaKleSoft:
Rebased and (hopefully) ready to merge. Btw, how do you guys run tests? If I run `gulp test` it runs zero tests because tests are compiled to `lib/test` rather than `test`. I had to change the path in the gulp task for it to work.

TimvdLippe:
You can run tests with `npm test` or `npm run test:watch`

MaKleSoft:
@TimvdLippe Ok cool. So I suppose using gulp is deprecated?

rictic:
We've been reducing our usage of gulp after finding a bunch of issues with common gulp libraries. `gulp-mocha`'s results are undefined if you call it twice in the same process, `gulp-typescript` doesn't fail on errors by default and sometimes reports spurious errors, etc.

src/test/polymer/polymer-element-scanner_test.ts
```diff
@@ -143,6 +154,16 @@ suite('PolymerElementScanner', () => {
...
+      assert.deepEqual(features[0].listeners, [
+        {event: 'event-a', handler: '_handleA'},
+        {event: 'eventb', handler: '_handleB'}
+      ]);
+
+      // Skip not statically analizable entries without emitting a warning
+      assert.equal(features[0].warnings.filter(w => w.code === 'invalid-listeners-declaration').length, 0);
+      // Emit warning for non-object `listeners` literal
+      assert.equal(features[1].warnings.filter(w => w.code === 'invalid-listeners-declaration').length, 1);
     });
```
rictic: Nice tests!

Issue #187, opened 2015-10-29
hydrolysis doesn't always find pseudo elements


rictic:
``` bash
$ cat example.html <
<script>
  /**
   * Test.
   *
   * @pseudoElement pseudo-elem
   */
</script>


$ cat example.js <
var hydrolysis = require('hydrolysis');

hydrolysis.Analyzer.analyze('example.html').then(function(data) {
  console.log('Elements found:');
  console.log(data.elements);
});


$ node example.js
Elements found:
[]
```

rictic:
Update: AJ discovered that the issue was that pseudo-elements are only found in HTML comments, not jsdoc.

rictic:
We should find them in both HTML comments and JS comments. The analyzer appears not to be finding either, so it doesn't find iron-icons. Or as @MaKleSoft said on #377:

> [...] element descriptions for elements that aren't actually registered anywhere via the `@pseudoElement` tag (hence 'pseudo elements'). `iron-icons` uses this  for example: https://github.com/PolymerElements/iron-icons/blob/master/iron-icons.html#L36

MaKleSoft:
@rictic Actually, #409 only takes care of HTML comments, so we might wanna keep this open for now. (I'll create a PR for JS comments, soon)

justinfagnani:
You know, pseudo element is a really bad name for this, since that's already a thing in HTML/CSS. I'd  expect a pseudo element scanner to be looking for `foo::bar` CSS selectors. I wonder if we can migrate to something else over time.

rictic:
Yeah, agreed. If we can agree on a better name we could recognize both in the analyzer but emit a Warning if you use the old style.

Brainstorming for terminology 🚲🏠ing for this class of thing generally (as we'd like the ability to have pseudo behaviors, pseudo properties, etc):
  * unknown implementation
  * declared

justinfagnani:
Why not just `@element`?

rictic:
That's not bad.

@MaKleSoft has a use case that involves distinguishing these elements so he's adding a `pseudo-element` kind to the generated ScannedPolymerElement but I think that might be better as `implementation-unknown`

MaKleSoft:
[...]

PR #409, opened 2016-12-14
Extract pseudo elements from comments

MaKleSoft:
Adds support for extracting pseudo elements declared in comments via the `@pseudoElement` tag. (partially) fixes https://github.com/Polymer/polymer-analyzer/issues/187

Some notes:

* Only works for HTML comments for now. If you guys are happy with this, I'll add support for javascript comments as well.
* Right now the scanner extracts `ScannedPolymerElement`. We'll probably want to at least add a `pseudo` flag to mark elements as pseudo elements but a separate feature would probably be better. Maybe extend PolymerElement and add something like `pseudo-element` to the `kinds` set?
* I saw that there is already a [parsePseudoElements](https://github.com/Polymer/polymer-analyzer/blob/master/src/polymer/docs.ts#L292) function but it doesn't seem to be used anywhere yet. I copy/pasted some of the code instead of using it directly since it didn't support extracting the source range. Not sure what's the deal with that function though.

### DIFF
--- a/src/polymer/pseudo-element-scanner.ts
+++ b/src/polymer/pseudo-element-scanner.ts
@@ -1,0 +1,51 @@
+/**
+ * @license
+ * Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at
+ * http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import * as dom5 from 'dom5';
+import {ASTNode} from 'parse5';
+
+import * as jsdoc from '../javascript/jsdoc';
+import {annotateElementHeader} from './docs';
+
+import {HtmlVisitor, ParsedHtmlDocument} from '../html/html-document';
+import {HtmlScanner} from '../html/html-scanner';
+import {ScannedPolymerElement} from './polymer-element';
+
+export class PseudoElementScanner implements HtmlScanner {
+  async scan(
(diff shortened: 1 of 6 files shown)


src/polymer/pseudo-element-scanner.ts
```diff
@@ -0,0 +1,51 @@
...
+import {ASTNode} from 'parse5';
+
+import * as jsdoc from '../javascript/jsdoc';
+import {annotateElementHeader} from './docs';
+
+import {HtmlVisitor, ParsedHtmlDocument} from '../html/html-document';
+import {HtmlScanner} from '../html/html-scanner';
+import {ScannedPolymerElement} from './polymer-element';
+
+export class PseudoElementScanner implements HtmlScanner {
```
rictic: A pseudo element is probably a weird enough concept to be worth adding a comment about, and here seems like the natural place. I'd go with something like:

```javascript
/**
 * A Polymer pseudo-element is an element that is declared in an unusual way, such
 * that the analyzer couldn't normally analyze it, so instead it is declared in
 * comments.
 */
```

MaKleSoft:
@rictic Cool. Do you want me to add javascript comment support right here or in a separate PR? Also, any comments on the last two bullet points?

rictic:
Small PRs are best IMO, and github has good support for base branches, so you can have PRs that build off of each other.

I think exposing them as normal ScannedPolymerElements is reasonable for now. If we run into any issues we can revisit, but this seems about right.

The jsdoc related code is some of the oldest code that hasn't been well integrated into the new analyzer yet. It's not surprising to me that there's a function that kinda was doing pseudo element parsing but that nothing was using it. It's just there as inspiration for the new correct thing, which it looks like this code is, so maybe we should delete that unused function as part of this PR.

MaKleSoft:
Actually, for one of the tools I need this for I'll also need to be able to distinguish between pseudo elements and regular elements so I added some minor changes:

- Add `pseudo` property to `ScannedPolymerElement`
- If `pseudo` is true, add `pseudo-element` to `PolymerElement`s `kinds` set upon resolving

rictic:
Cool, that all makes total sense. Merging!

PR #410, opened 2016-12-14
Manage async work caching in a class.

rictic:
This lets us remove the difficult-to-reason-about code (including the IIAFE with the immediate `await Promise.resolve()`) into one place where it can declare its invariants and be better maintained.

 - [x] CHANGELOG.md has been not been updated, this is an internal noop refactoring.

### DIFF
--- a/src/core/analysis-cache.ts
+++ b/src/core/analysis-cache.ts
@@ -39,6 +39,10 @@ export class AnalysisCache {
    */
   dependenciesScanned: AsyncWorkCache<string, void>;
 
+  /**
+   * TODO(rictic): These synchronous caches need to be kept in sync with their
+   *     async work cache analogues above.
+   */
   scannedDocuments: Map<string, ScannedDocument>;
   analyzedDocuments: Map<string, Document>;
 
@@ -100,8 +104,7 @@ export class AnalysisCache {
       // definite, and not invalidated.
       newCache.analyzedDocumentPromises.clear();
       for (const keyValue of newCache.analyzedDocuments) {
-        newCache.analyzedDocumentPromises.set(
-            keyValue[0], Promise.resolve(keyValue[1]));
+        newCache.analyzedDocumentPromises.set(keyValue[0], keyValue[1]);
       }
     }
 
--- a/src/core/analysis-cache.ts
+++ b/src/core/analysis-cache.ts
@@ -15,6 +15,7 @@
 import {Document, ScannedDocument} from '../model/model';
 import {ParsedDocument} from '../parser/document';
 
+import {AsyncWorkCache} from './async-work-cache';
 import {DependencyGraph} from './dependency-graph';
 
 export class AnalysisCache {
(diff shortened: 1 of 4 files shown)


src/core/analysis-cache.ts
```diff
@@ -111,3 +114,10 @@ function shallowCopyMap<K, V>(from?: Map<K, V>) {
   }
   return new Map();
 }
+
+function shallowCopyAsyncWorkCache<K, V>(from?: AsyncWorkCache<K, V>) {
```
justinfagnani: This function seems unnecessary, AsyncWorkCache's constructor already does this.

src/core/analyzer-cache-context.ts
```diff
@@ -139,9 +131,9 @@ export class AnalyzerCacheContext {
       const document = this._makeDocument(scannedDocument);
       doneTiming();
       return document;
-    })();
-    this._cache.analyzedDocumentPromises.set(resolvedUrl, promise);
-    return promise;
+    };
+    return this._cache.analyzedDocumentPromises.getOrCompute(
+        resolvedUrl, computeDocument);
```
justinfagnani: I like the style of inlining the compute function like you do below

src/core/async-work-cache.ts
```diff
@@ -0,0 +1,68 @@
...
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+/**
+ * A map from keys to promises of values. Used for caching asynchronous work.
+ */
+export class AsyncWorkCache<K, V> {
+  private _keyToResultMap: Map<K, Promise<V>>;
```
justinfagnani: Given that for two of the uses we also keep the synchronous result in a map, why don't we fold that in here so we ensure we keep them in sync properly.

src/core/async-work-cache.ts
```diff
@@ -0,0 +1,68 @@
...
+   * This method ensures that we will only try to compute the value for `key`
+   * once, no matter how often or with what timing getOrCompute is called, even
+   * recursively.
+   */
+  async getOrCompute(key: K, compute: () => Promise<V>) {
+    const cachedResult = this._keyToResultMap.get(key);
+    if (cachedResult) {
+      return cachedResult;
+    }
+    const promise = (async() => {
```
justinfagnani: Could this be simply: `const promise = Promise.resolve().then(() => compute());`

rictic: Yes, but I prefer the more explicit style here, as I also want to leave in the comment about why we have the `Promise.resolve()`

src/core/analyzer-cache-context.ts
rictic: The indentation bothers me a bit, but it is easier to read. Done.

src/core/async-work-cache.ts
rictic: I can see that. There are some fuzzy bits that I want to double check, but I left a TODO.

PR #411, opened 2016-12-14
Teach AsyncWorkQueue how to cancel

rictic:
- [x] CHANGELOG.md has not been updated, this is a noop change in preparation for adding CancelTokens.

After this (provided it's correct) we would get cancellation just by threading a CancelToken through to the AsyncWorkQueue#getOrCompute calls.


FredKSchott:
Closing this PR as abandoned.

src/core/async-work-cache.ts
```diff
@@ -31,11 +35,34 @@ export class AsyncWorkCache<K, V> {
...
    * If not, compute it with the given function.
    *
-   * This method ensures that we will only try to compute the value for `key`
-   * once, no matter how often or with what timing getOrCompute is called, even
-   * recursively.
+   * This method ensures that, in the absence of cancellations, we will only try
+   * to compute the value for `key` once, no matter how often or with what
+   * timing getOrCompute is called, even recursively.
+   *
+   * If a compute() throws a Cancel that should not result in a Cancel for
```
justinfagnani: I don't quite understand the first sentence here.

src/core/async-work-cache.ts
```diff
@@ -31,11 +35,34 @@ export class AsyncWorkCache<K, V> {
...
+   * non-cancelled operations. So long as the given cancelToken is not
+   * cancelled, the caller will not receive a Cancel.
    */
-  async getOrCompute(key: K, compute: () => Promise<V>) {
+  async getOrCompute(
+      key: K, compute: () => Promise<V>, cancelToken?: CancelToken) {
+    cancelToken = cancelToken || neverCancels;
```
justinfagnani: put the default in the parameter definition

src/core/async-work-cache.ts
```diff
@@ -31,11 +35,34 @@ export class AsyncWorkCache<K, V> {
...
+    cancelToken.throwIfRequested();
+    while (true) {
+      try {
+        const result = await this._getOrCompute(key, compute);
+        cancelToken.throwIfRequested();
+        return result;
+      } catch (err) {
+        cancelToken.throwIfRequested();
```
justinfagnani: Should `cancelToken.throwIfRequested()` come last? In a case where there's both a cancel and a program exception, should the exception be thrown so we can see it?

package.json
```diff
@@ -80,6 +80,7 @@
     "babel-traverse": "^6.26.0",
     "babel-types": "^6.26.0",
     "babylon": "^6.18.0",
+    "cancel-token": "^0.1.1",
```
justinfagnani: I'd like to see this library moved to PolymerLabs, or maybe even Polymer, since we're depending on it from non-labs code.

justinfagnani:
Needs a rebase

src/core/async-work-cache.ts
rictic: Rephrased, PTAL

src/core/async-work-cache.ts
rictic: Ah, yeah, that's something we can do now that we've dropped node v4.

src/core/async-work-cache.ts
rictic: Hm. If you've cancelled an operation and then that operation fails, do you care? IMO, when you cancel an operation you no longer care about either success or failure.

package.json
rictic: Yeah, seems like things have cooled down a bit. Looks like the fetch cancellation API is basically this one, though with fewer methods on the token: https://developer.mozilla.org/en-US/docs/Web/API/AbortController

justinfagnani: FYI, I think it's fine to not exactly emulate any other API for the time being. Maybe just don't call it a polyfill.

rictic: Moved to PolymerLabs, happy to move it from there to Polymer, maybe after a 1.0?

Updated a few links, README, and added a LICENSE file here: https://github.com/PolymerLabs/cancel-token/pull/2

PR #415, opened 2016-12-18
Fix an issue with parallel analysis of mutually importing documents.

rictic:

We do take a ~100% hit to the benchmark, but that is a worst case scenario for this change (modifying the root of a very large import graph), and analysis is still only taking ~40ms.

It would still be nice if we could figure out a way to add something like this cache. There's no correctness issue with scanning dependencies multiple times, just a performance one, so we don't need the full power of getOrCompute. Hm. I think I have an idea for a followup PR that restores this performance win.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
    this comment.
 -->
 
+* Fix a deadlock when there are concurrent analysis runs of cyclic graphs.
+
 ## [2.0.0-alpha.20] - 2016-12-19
 
 ### Added
@@ -20,8 +22,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 * [Polymer] Extract pseudo elements from HTML comments
 
 ### Fixed
-* Fixed a class of race conditions and cache invalidation errors that can occur when there are parallel analysis runs and edits to files.
-* Fixed a deadlock when there are parallel analysis runs of cyclic graphs.
+* Fix a class of race conditions and cache invalidation errors that can occur when there are concurrent analysis runs and edits to files.
+
 
 ## [2.0.0-alpha.19] - 2016-12-12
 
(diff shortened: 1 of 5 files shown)


rictic:
I changed the dependency scanning "cache" to just be a set of urls that we can check synchronously. If we've definitely finished scanning a url's dependencies, then great! we can skip further work. If not, don't worry about repeating work.

CHANGELOG.md
```diff
@@ -20,8 +20,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 * [Polymer] Extract pseudo elements from HTML comments
 
 ### Fixed
-* Fixed a class of race conditions and cache invalidation errors that can occur when there are concurrent analysis runs and edits to files.
-
+* Fixed a class of race conditions and cache invalidation errors that can occur when there are parallel analysis runs and edits to files.
```
justinfagnani: pedantic: I think "concurrent" is the right term here. _If_ anyone intends there to be a difference, it's usually that concurrent operations are not necessarily parallel, and parallel actually means truly at the same time.

justinfagnani:
Seems ok for now to fix the bug, but it still seems to me like the dependenciesScannedOf Set isn't quite the best way to track loading. Let's chat next time we can.

CHANGELOG.md
rictic: Arg, you're right, this gets me every time. The technical definitions of these words just seem so intuitively backwards to me.

`<rant please-ignore>`

In the nontechnical dictionary sense, concurrent means multiple things happening simultaneously, and parallel means two things being side by side. When there are two threads of execution they're naturally represented as parallel lines, and even when they run on one core they're still separate linear sequences of operations that just contend for the core.

`</rant>`

rictic:
Sounds good. I'm interested in a better way of doing this, we've had a few bugs in here lately, and it seems like we should represent what's going on here better.

PR #416, opened 2016-12-18
Add Analyzer#analyzePackage()

rictic:

This is something of a strawman proposal, as I'm confident that I want something very much like this, and that it should live on the analyzer, but I'm much less confident about the details.

More specifically, goals that this satisfies:

* Discovering all files in the project
* Excluding unimported files from dependencies
* Exposing a single, queryable object that represents all of this.

Use cases:

* Running the linter on a project-wide basis
* Running `polymer analyze` on a project-wide basis
* Generating elements.json for a project in polyserve
* Finding all references of an element in an editor

The thing that I'm most unhappy with, from a design standpoint, is of course the `listAllFilesInProject` method on UrlLoader. It's a pretty complex thing to add to a fairly simple interface, and there's not a clear implementation path for implementing it in the FetchUrlLoader (though it's doable with server-side cooperation, e.g. in the designer use case). It reminds me a bit of the other method I wanted to add to UrlLoader, where you could ask it for files in a directory, for use as part of autocompletion of paths in the editor.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 -->
 
 ### Added
-* Add an analyzeProject() method, for getting a queryable representation of everything in a project.
+* Add an analyzePackage() method, for getting a queryable representation of everything in a package.
 
 ### Fixed
 * Fix a deadlock when there are concurrent analysis runs of cyclic graphs.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [2.0.0-alpha.20] - 2016-12-19
 
 ### Added
+* Add an analyzeProject() method, for getting a queryable representation of everything in a project.
 * [Polymer] Extract 'listeners' from 1.0-style declarations.
 * [Polymer] Extract pseudo elements from HTML comments
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,12 +13,15 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
    this comment.
 -->
 
+### Added
+* Add an analyzeProject() method, for getting a queryable representation of everything in a project.
+
+### Fixed
 * Fix a deadlock when there are concurrent analysis runs of cyclic graphs.
 
 ## [2.0.0-alpha.20] - 2016-12-19
 
 ### Added
-* Add an analyzeProject() method, for getting a queryable representation of everything in a project.
 * [Polymer] Extract 'listeners' from 1.0-style declarations.
 * [Polymer] Extract pseudo elements from HTML comments
 
--- a/src/model/project.ts
+++ b/src/model/project.ts
@@ -18,9 +18,11 @@ import {Feature} from './feature';
 
 export class Project {
   private _rootDocuments: Set<Document>;
+  private _toplevelWarnings: Warning[];
 
-  constructor(rootDocuments: Iterable<Document>) {
+  constructor(rootDocuments: Iterable<Document>, warnings: Warning[]) {
     this._rootDocuments = new Set(rootDocuments);
+    this._toplevelWarnings = warnings;
 
     // This is a performance optimization. We only need a set of documents such
     // that all other documents we're interested in can be reached from them.
@@ -84,7 +86,7 @@ export class Project {
    * Get all warnings in the project.
    */
   getWarnings(): Warning[] {
-    const result = new Set();
+    const result = new Set(this._toplevelWarnings);
     for (const doc of this._rootDocuments) {
       addAll(result, new Set(doc.getWarnings(true)));
     }
--- a/src/model/project.ts
+++ b/src/model/project.ts
@@ -1,0 +1,100 @@
+/**
+ * @license
+ * Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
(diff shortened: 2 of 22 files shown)


src/url-loader/url-loader.ts
```diff
@@ -14,4 +14,9 @@ export interface UrlLoader {
    * This should only be called if `canLoad` returns `true` for `url`.
    */
   load(url: string): Promise<string>;
+
+  /**
+   * Lists the files in the current project.
+   */
+  listFilesInProject(): Promise<string[]>;
```
justinfagnani: as discussed, let go with: `readDirectory(path: string, deep?: boolean): Promise<string[]>`

src/url-loader/fetch-url-loader.ts
```diff
@@ -46,4 +46,8 @@ export class FetchUrlLoader implements UrlLoader {
       }
     });
   }
+
+  listFilesInProject(): Promise<string[]> {
+    throw new Error('Not implementable.');
```
justinfagnani: let's do: "Not supported", since this could be implemented in a subclass with a known server protocol.

src/url-loader/fs-url-loader.ts
```diff
@@ -64,4 +66,41 @@ export class FSUrlLoader implements UrlLoader {
...
   }
-};
+
+  async listFilesInProject(): Promise<string[]> {
+    return this._listFilesInDir('');
+  }
+
+  async _listFilesInDir(pathFromRoot: string): Promise<string[]> {
+    const files = await new Promise<string[]>((resolve, reject) => {
+      fs.readdir(
```
justinfagnani: I discovered `mz` and `fs-promise` the other day. `mz` is really nice. Since we use fs-extra, I think we should pull in fs-promise: https://www.npmjs.com/package/fs-promise

src/model/project.ts
```diff
@@ -0,0 +1,102 @@
...
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+import {Warning} from '../warning/warning';
+
+import {Document, FeatureKinds} from './document';
+import {Feature} from './feature';
+
+export class Project {
```
justinfagnani: docs

src/model/project.ts
```diff
@@ -0,0 +1,102 @@
...
+  private _rootDocuments: Set<Document>;
+  private _toplevelWarnings: Warning[];
+
+  constructor(rootDocuments: Iterable<Document>, warnings: Warning[]) {
```
justinfagnani: docs, particularly defining `rootDocuments`

justinfagnani: Actually, given what you do next, I wouldn't call this root documents, just documents.

src/model/project.ts
```diff
@@ -0,0 +1,102 @@
...
+    this._rootDocuments = new Set(rootDocuments);
```
justinfagnani: consider not assigning this until you've filtered it

src/model/project.ts
```diff
@@ -0,0 +1,102 @@
...
+    this._toplevelWarnings = warnings;
+
+    // This is a performance optimization. We only need a set of documents such
```
justinfagnani: How much of an optimization is this vs just passing all documents to `analyze()`? the analysis cache should take care of avoiding duplicate effort.

src/model/project.ts
```diff
@@ -0,0 +1,102 @@
...
+        // When there's cycles we can keep any element of the cycle, so why not
+        // this one.
+        if (imprt.document !== doc) {
+          this._rootDocuments.delete(imprt.document);
+        }
+      }
+    }
+  }
+
+  getByKind<K extends keyof FeatureKinds>(kind: K): Set<FeatureKinds[K]>;
```
justinfagnani: Share a common interface with Document?

rictic: I'm cool with that, if just to enforce that they be mutually compatible. What would we call such a thing?

src/model/project.ts
rictic: The optimization is in the expense of duplicate work when querying, not in analyzing.

src/model/project.ts
rictic: I went with Queryable.

src/url-loader/fetch-url-loader.ts
rictic: Instead I made the method optional. Lets us do better error messages anyways.

src/url-loader/fs-url-loader.ts
rictic: There are only two non-test files where we use fs. This one, and generate-elements, and generate-elements could use require, as it's loading a json file from disk that's packaged with the analyzer's code. I don't think that's worth the overhead of another dependency (yet), even if it is just 200KiB

src/url-loader/url-loader.ts
rictic: Done

src/model/project.ts
rictic: Done

CHANGELOG.md
```diff
@@ -13,6 +13,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
    this comment.
 -->
 
+### Added
+* Add an analyzeProject() method, for getting a queryable representation of everything in a project.
```
justinfagnani: analyzePackage()

src/model/project.ts
justinfagnani: +1

PR #425, opened 2017-01-09
Refactor scanning to track dependency scanning with PromiseGroups

justinfagnani:


src/core/dependency-graph.ts
```diff
@@ -12,25 +12,73 @@
...
+import {Deferred} from '../utils';
+
+class DocumentRecord {
+  url: string;
+  dependencies: Set<string>;
+  dependants: Set<string>;
+
+  // Represents the Promise that the dependencies are known, but nothing
+  // about their state (loaded, scanned, etc)
+  _dependenciesDeferred: Deferred<Set<string>>;
```
rictic: `private`

src/core/dependency-graph.ts
```diff
@@ -39,20 +87,51 @@ export class DependencyGraph {
...
+      for (const dep of deps) {
+        try {
+          await this._whenReady(dep, visited);
+        } catch (ee) {
+          // ignore rejected dependencies
+        }
       }
-      dependants.add(path);
+    } catch (e) {
+      // turn rejection into resolution
```
rictic: Is this catch needed? It's pretty broad

src/core/analyzer-cache-context.ts
```diff
@@ -258,37 +252,74 @@ export class AnalyzerCacheContext {
...
-    const scannedDocument =
-        await this._cache.scannedDocumentPromises.getOrCompute(
-            resolvedUrl, async() => {
-              const parsedDoc = await this._parse(resolvedUrl, contents);
-              return this._scanDocument(parsedDoc, actualVisited);
+  private async _scan(resolvedUrl: string, contents?: string):
+      Promise<ScannedDocument> {
+    return this._cache.dependenciesScannedPromises.getOrCompute(
+        resolvedUrl, async() => {
+          const scannedDocument = await this._scanLocal(resolvedUrl, contents);
```
rictic: Optional: what if we returned both the scannedDocument and the list of non-lazy resolved import urls from _scanLocal?

Pro: less duplicated code

Cons: somewhat tightly coupled here?

src/core/analyzer-cache-context.ts
```diff
@@ -258,37 +252,74 @@ export class AnalyzerCacheContext {
...
+          // Find all non-lazy imports
+          const imports = scannedDocument.getNestedFeatures().filter(
+              (e) => e instanceof ScannedImport &&
+                  e.type !== 'lazy-html-import') as ScannedImport[];
+
+          // Scan imports
+          for (const scannedImport of imports) {
+            const importUrl = this._resolveUrl(scannedImport.url);
+            this._scan(importUrl).catch((error) => {
```
rictic: Worth adding a comment here that we're intentionally not awaiting on the result here to avoid the deadlock, and we're getting the benefit of that await with the call to `dependencyGraph.whenReady` below.

src/core/analyzer-cache-context.ts
```diff
@@ -258,37 +252,74 @@ export class AnalyzerCacheContext {
     return copy;
   }
 
+  private async _scanLocal(resolvedUrl: string, contents?: string):
```
rictic: Mention what local means in a comment here?

src/test/core/analysis-cache_test.ts
```diff
@@ -26,10 +28,12 @@ suite('AnalysisCache', () => {
     cache.parsedDocumentPromises.set(path, `parsed ${path} promise` as any);
     cache.scannedDocumentPromises.set(path, `scanned ${path} promise` as any);
     cache.analyzedDocumentPromises.set(path, `analyzed ${path} promise` as any);
-    cache.dependenciesScannedOf.add(path);
+    const dependenciesScanned = new PromiseGroup<ScannedDocument>();
+    dependenciesScanned.close();
+    // cache.dependenciesScannedOf.set(path, dependenciesScanned);
```
rictic: ?

rictic:
I think just rebase to master and drop the fix tests commit

src/core/async-work-cache.ts
```diff
@@ -50,6 +50,10 @@ export class AsyncWorkCache<K, V> {
     return promise;
   }
 
+  get(key: K) {
```
justinfagnani: done

src/core/dependency-graph.ts
justinfagnani: this whole class is private, for the moment. Renamed without `_` to match other fields.

src/core/dependency-graph.ts
justinfagnani: The outer try was needed to catch per-file rejections, but the inner wasn't since this function only called itself. I redid this to use a catch and added some intermediate vars to make it more clear.

src/core/analyzer-cache-context.ts
justinfagnani: I bristled at the duplicated code but decided to defer until a bit later, mainly because this line is asking a simple question, and it should be possible to phrase it more simply than this. I think either we should improve the ergonomics of the of the feature query API, add direct support for imports to ScannedDocument, or maybe both.

Also, I need to add a TODO, but for the purposes of scanning I think we want to include even the lazy imports.

src/core/analyzer-cache-context.ts
justinfagnani: +1 done

src/test/core/analysis-cache_test.ts
justinfagnani: done

justinfagnani:
Comments addressed, removed PromiseGroup, and rebased the PR onto `master`, so the commits are different. PTAL.

I haven't looked at the memory usage yet. There's a small amount of per-file additional memory to be expected, but not a leak unless we're retaining contexts somewhow.

src/core/analyzer-cache-context.ts
```diff
@@ -259,35 +253,99 @@ export class AnalyzerCacheContext {
...
+    return this._cache.dependenciesScannedPromises.getOrCompute(
+        resolvedUrl, async() => {
+          const scannedDocument = await this._scanLocal(resolvedUrl, contents);
+          // Find all non-lazy imports
+          // TODO(justinfagnani): I think we should scan lazily imported
+          // documents
+          // since we know about them, we should load them. Their features
+          // should
+          // possibly be separated out at export time via better definition of
+          // scopes
```
rictic: Style: reflow comments here

src/core/analyzer-cache-context.ts
rictic: rgr

src/core/analyzer-cache-context.ts
```diff
@@ -259,35 +253,99 @@ export class AnalyzerCacheContext {
...
+              (e) => e instanceof ScannedImport &&
+                  e.type !== 'lazy-html-import') as ScannedImport[];
+
+          // Scan imports
+          for (const scannedImport of imports) {
+            const importUrl = this._resolveUrl(scannedImport.url);
+            // Request a scan of `importUrl` but do not wait for the results to
+            // avoid deadlock in the case of cycles. Later we use the
+            // DependencyGraph
+            // to wait for all transitive dependencies to load.
```
rictic: Style: reflow this comment so that DependencyGraph isn't alone on a line like that?

src/core/analyzer-cache-context.ts
```diff
@@ -322,59 +379,7 @@ export class AnalyzerCacheContext {
     return [];
   }
 
-  /**
-   * Scans all of the transitive dependencies of the given document.
-   *
-   * Uses the `visited` set to break cycles.
-   */
-  private async _scanImports(
```
rictic: Does inlining this into _scan help legibility? I kinda like it more with them separate.

src/core/analyzer-cache-context.ts
justinfagnani: done

src/core/analyzer-cache-context.ts
justinfagnani: done

src/core/analyzer-cache-context.ts
justinfagnani: _scan is now basically _scanImports anyway

PR #428, opened 2017-01-13
Improve ES6 module support.

rictic:
This ensures that we don't throw parse errors for ES6 modules. We can't, in general, know at parse time whether a given bit of JS should be parsed as a module or a script.

I think, given that, it is better to parse robustly and just note whether a file must be a module, or that it is legal syntax as either a module or a script.

Fixes https://github.com/Polymer/polymer-editor-service/issues/12 and https://github.com/Polymer/vscode-plugin/issues/44

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-analyzer/428)

### DIFF
--- a/src/javascript/javascript-parser.ts
+++ b/src/javascript/javascript-parser.ts
@@ -28,43 +28,42 @@ declare class SyntaxError {
 }
 
 export class JavaScriptParser implements Parser<JavaScriptDocument> {
-  sourceType: 'module'|'script';
-
-  constructor(options: {sourceType: 'module' | 'script'}) {
-    console.assert(options != null);
-    console.assert(options.sourceType != null);
-    this.sourceType = options.sourceType;
-  }
-
   parse(contents: string, url: string, inlineInfo?: InlineDocInfo<any>):
       JavaScriptDocument {
     const isInline = !!inlineInfo;
     inlineInfo = inlineInfo || {};
     let ast: Program;
+    const options = {
+      ecmaVersion: 8,
+      attachComment: true,
+      comment: true,
+      loc: true,
+      sourceType: 'script' as ('script' | 'module'),
+    };
+
     try {
-      ast = <Program>espree.parse(contents, {
-        ecmaVersion: 8,
-        attachComment: true,
-        comment: true,
-        loc: true,
-        sourceType: this.sourceType,
-      });
-    } catch (err) {
-      if (err instanceof SyntaxError) {
-        throw new WarningCarryingException({
-          message: err.message.split('\n')[0],
-          severity: Severity.ERROR,
-          code: 'parse-error',
-          sourceRange: correctSourceRange(
-              {
-                file: url,
-                start: {line: err.lineNumber - 1, column: err.column - 1},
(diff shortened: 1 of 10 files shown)


justinfagnani:
I was thinking about this, and I think ultimately we want to keep some metadata with a file about how it was imported. The metadata can be a key into the work caches and inform parsing, etc. meaning that we could support a file imported as a script and module, even if we don't really want to. This might also come in handy for handling different ways HTML could be imported.

rictic:
I disagree. I think that we want to keep parsing separate from later stages, as squirrelling metadata away at scan time (or later) into a parse-time data structure is going to make cache consistency a nightmare.

Instead, we can just store all of the relevant parse-time information at parse time (like whether a file must be parsed as a module), and then do whatever lint checks or cross-file-aware processing at the link or resolve phases.

src/javascript/javascript-parser.ts
```diff
@@ -28,43 +28,42 @@ declare class SyntaxError {
...
-                file: url,
-                start: {line: err.lineNumber - 1, column: err.column - 1},
-                end: {line: err.lineNumber - 1, column: err.column - 1}
-              },
-              inlineInfo.locationOffset)!
-        });
+      ast = <Program>espree.parse(contents, options);
+    } catch (_ignored) {
+      try {
+        options.sourceType = 'module';
```
justinfagnani: is there anyway to tell that the error was from an import or export?

rictic: After we parse it again as a module we could walk the AST looking for imports and exports.

justinfagnani:
I don't see how cache consistency is any harder if metadata is part of the file key. Instead of requesting `'foo.js'` we could request `{url: 'foo.js', type: 'module'}`. Also, we'll eventually get to projects with a significant number of modules, exceptioning and re-parsing on all of them when we could avoid that.

src/javascript/javascript-parser.ts
justinfagnani: I'm wondering if we can skip reparsing if it's not a module-related syntax error.

rictic:
Oh, I misunderstood, I thought it involved mutating the ParsedDocument.

We could instead always parse as modules, and just have lint warnings if you reference a file with imports or exports as a script.

src/javascript/javascript-parser.ts
justinfagnani: eh, forget it for now

justinfagnani:
Well, the problem there, and this is only because of JS, is the syntax differences. Modules are always strict is one that _might_ affect parsing in that direction?

justinfagnani:
But yeah, for HTML parsing wouldn't change, though imports vs modules might affect scanning.

rictic:
True, but does strict vs loose parsing ever produce two different ASTs for a given input, or does strict parsing just reduce the language?

rictic:
I was recently reading this part of the spec, but didn't get far enough to have a confident answer to this question.

justinfagnani:
I don't remember either. I usually see lists of difference that include syntax and semantics. It's not super-high priority, we can ask our V8 buddies.

PR #429, opened 2017-01-13
Fix the build, option #1

rictic:
It looks like the node v4 typings don't include definitions for `console` anymore? Seems like a bug, but moving to node v6 typings fixes.

### DIFF
--- a/src/url-loader/fs-url-loader.ts
+++ b/src/url-loader/fs-url-loader.ts
@@ -27,7 +27,7 @@ export class FSUrlLoader implements UrlLoader {
   root: string;
 
   constructor(root?: string) {
-    this.root = root || process.cwd();
+    this.root = root || '';
   }
 
   canLoad(url: string): boolean {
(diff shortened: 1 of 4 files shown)


src/url-loader/fs-url-loader.ts
```diff
@@ -24,9 +24,9 @@ import {UrlLoader} from './url-loader';
  * Resolves requests via the file system.
  */
 export class FSUrlLoader implements UrlLoader {
-  root: string|undefined;
+  root: string;
 
-  constructor(root?: string) {
+  constructor(root: string) {
```
justinfagnani: how about `root: string = ''`?

justinfagnani: Also, we might want to pick between #L65 and #L71 for how to handle undefined roots.

rictic: WFM. Went with the closest node v4 compatible version of that

PR #437, opened 2017-01-30
Typescript 1.1

justinfagnani:


justinfagnani:
@rictic getting closer on how things fit together and using the TS API. I was able to avoid double parsing, and have a test that shows we can tell if a class extends `HTMLElement`, and get its properties.

src/core/analysis-context.ts
```diff
@@ -58,10 +61,15 @@ export class AnalysisContext {
   private _parsers = new Map<string, Parser<ParsedDocument<any, any>>>([
     ['html', new HtmlParser()],
     ['js', new JavaScriptParser()],
+    ['ts', new TypeScriptPreparser()],
     ['css', new CssParser()],
     ['json', new JsonParser()],
   ]);
 
+  private _languageAnalyzers = new Map<string, LanguageAnalyzer<any>>([
+    ['ts', new TypeScriptAnalyzer(this)],
```
rictic: Should we use the same object to analyze both ts and js files? I suspect the answer is yes

src/test/typescript/typescript-analyzer_test.ts
```diff
@@ -0,0 +1,105 @@
...
+
+import {assert} from 'chai';
+import * as ts from 'typescript';
+
+import {AnalysisContext} from '../../core/analysis-context';
+import {TypeScriptAnalyzer} from '../../typescript/typescript-analyzer';
+import {UrlLoader} from '../../url-loader/url-loader';
+import {UrlResolver} from '../../url-loader/url-resolver';
+
+class TestUrlResolver implements UrlResolver {
```
rictic: Recommend dropping the special test url resolvers and loaders and just put your test data in the fixtures dir. The tests will be more authentic, and there will be less boilerplate obscuring what they're doing.

src/test/typescript/typescript-analyzer_test.ts
```diff
@@ -0,0 +1,105 @@
...
+      });
+
+      // Get class A and assert that it extends HTMLElement
+      const sourceFile = program.getSourceFile(fileName);
+      ts.forEachChild(sourceFile, (node) => {
+        if (node.kind === ts.SyntaxKind.ClassDeclaration) {
+          const clazz = node as ts.ClassDeclaration;
+          if (clazz.name && clazz.name.getText() === 'A') {
+            const type = checker.getTypeAtLocation(clazz);
+            const baseTypes = checker.getBaseTypes(type as ts.InterfaceType);
```
rictic: Unfortunate that there's so much casting needed.

src/typescript/typescript-analyzer.ts
```diff
@@ -0,0 +1,196 @@
...
+    // not found
+  }
+  libraryCache.set(filePath, source);
+  return source;
+}
+
+/**
+ * A TypeScript CompilerHost that only supports a single file, so that we can
+ * find imports in TypeScript files before we have loaded a whole program.
+ */
```
rictic: Is this comment correct?

src/typescript/typescript-analyzer.ts
```diff
@@ -0,0 +1,196 @@
...
+    return defaultLib;
+  }
+
+  get writeFile(): ts.WriteFileCallback {
+    return (_fileName: string,
+            _data: string,
+            _writeByteOrderMark: boolean,
+            _onError?: (message: string) => void,
+            _sourceFiles?: ts.SourceFile[]): void => {
+      // throw new Error('unsupported operation');
```
rictic: can we uncomment this?

src/core/analysis-context.ts
justinfagnani: Oh definitely. The only driving reason for these additions is better JS analysis. We're using the TypeScript compiler for that because it seems to have the better API + inference capabilities than the alternatives. I'm only wiring it in for `ts` files for now so that we can get scanners ported and test coverage up before switching `js` to use them.

src/test/typescript/typescript-analyzer_test.ts
justinfagnani: I really like being able to provide small test cases as strings. I'm inclined to add an `InMemoryUrlLoader` that takes a map of files to make this even easier.

src/test/typescript/typescript-analyzer_test.ts
justinfagnani: I'm not sure how the API could easily reduce that. Maybe if there were enough overrides of `getTypeAtLocation`...

src/typescript/typescript-analyzer.ts
justinfagnani: yes

src/test/typescript/typescript-analyzer_test.ts
rictic: Well there's the cast to ClassDecoration that immediately follows a comparison for same, which is just a discrepancy in the type checker (can't do type discrimination on enum instances).

The type type should be able to do the same (even if we might not bother with the assertion here).

justinfagnani: Yeah. Oh, and speaking of the casts, I should really put it on the var declaration, not the use.

src/test/typescript/typescript-analyzer_test.ts
rictic: +1, that could make tests like this more legible. Something like this perhaps?

```typescript
export function getStaticTestAnalyzer(contentsMap: {[path: string]: string}): Analyzer;
```

justinfagnani: done. That'll be really nice for more tests.

justinfagnani:
@rictic PTAL

PR #444, opened 2017-02-06
Fix several compilation errors with Typescript 2.2

TimvdLippe:
Installing the analyzer with `yarn` pulls in Typescript 2.2. This resulted in some compiler errors which I fixed all but one.

In TS 2.2 operators are more thoroughly checked, most notably arithmetic expressions do not allow undefined or null. I fixed these errors.

The last compiler error is because `lib.es2017.d.ts` in `typescript-analyzer.ts` pulls in `lib.es2016.d.ts` which pulls in `lib.es2015.d.ts` which pulls in `lib.es2015.collection.d.ts` and that file has a definition of `Map<K,V>`. This definition forces the second argument `value` of to be non-null per [this commit](https://github.com/Microsoft/TypeScript/commit/adedc1953b9a1136b5067f5224a3b4e7cd5b8956#diff-f7219547c512c7ed79bb415579ceb5afR27). I am not sure why the Typescript maintainers introduced this change, since on master `value` is allowed to be `undefined`. Therefore I pinned the version of Typescript to 2.1.4 for now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,6 +83,6 @@
     "shady-css-parser": "0.0.8",
     "split": "^1.0.0",
     "strip-indent": "^2.0.0",
-    "typescript": "^2.1.4"
+    "typescript": "2.1.4"
   }
 }
--- a/package.json
+++ b/package.json
@@ -83,6 +83,6 @@
     "shady-css-parser": "0.0.8",
     "split": "^1.0.0",
     "strip-indent": "^2.0.0",
-    "typescript": "2.1.4"
+    "typescript": "^2.1.4"
   }
 }
--- a/src/javascript/ast-value.ts
+++ b/src/javascript/ast-value.ts
@@ -28,6 +28,9 @@ function literalToValue(literal: estree.Literal): LiteralValue {
  */
 function unaryToValue(unary: estree.UnaryExpression): LiteralValue {
   const operand = expressionToValue(unary.argument);
+  if (operand === null || operand === undefined) {
+    return operand;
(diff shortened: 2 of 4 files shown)


src/javascript/ast-value.ts
```diff
@@ -28,6 +28,9 @@ function literalToValue(literal: estree.Literal): LiteralValue {
  */
 function unaryToValue(unary: estree.UnaryExpression): LiteralValue {
   const operand = expressionToValue(unary.argument);
+  if (operand === null || operand === undefined) {
+    return operand;
```
rictic: Actually we want to match javascript semantics here, even when it looks like a mistake. `+null === 0` so when we encounter that code the value we should produce is `0`.

Recommend changing those three expressions on operator to be like:

`+(operator as any)`

TimvdLippe:
The corresponding 2.2 PR which has stricter operand checks is https://github.com/Microsoft/TypeScript/pull/13483

package.json
```diff
@@ -83,6 +83,6 @@
     "shady-css-parser": "0.0.8",
     "split": "^1.0.0",
     "strip-indent": "^2.0.0",
-    "typescript": "^2.1.4"
+    "typescript": "2.1.4"
```
rictic: I'd prefer to just change libraryCache's type to `Map<string, string|undefined>`, because we want Typescript 2.2's ability to analyze mixins.

rictic:
LGTM!

PR #445, opened 2017-02-06
Add a simple test class for testing source ranges.

rictic:
Previously this was awkward, but we want to make it as easy as possible to test source ranges and warnings, as they're key to the user experience of the analyzer and downstream tools.

 - [x] CHANGELOG.md has not been updated, internal change.

### DIFF
--- a/src/test/test-utils.ts
+++ b/src/test/test-utils.ts
@@ -12,7 +12,11 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
+import {Analyzer} from '../analyzer';
+import {SourceRange} from '../model/model';
 import {UrlLoader} from '../url-loader/url-loader';
+import {Warning} from '../warning/warning';
+import {WarningPrinter} from '../warning/warning-printer';
 
 export class UnexpectedResolutionError extends Error {
   resolvedValue: any;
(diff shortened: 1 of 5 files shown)


src/test/test-utils.ts
```diff
@@ -50,3 +54,49 @@ export class TestUrlLoader implements UrlLoader {
...
+
+  /**
+   * Converts one or more warnings/source ranges into underlined text.
+   *                                                  ~~~~~~~~~~ ~~~~
+   *
+   * This has a loose set of types that it will accept in order to make
+   * writing tests simple and legible.
+   */
+  async underline(reference: Reference): Promise<string>;
+  async underline(references: Array<Reference>): Promise<string[]>;
```
justinfagnani: why `Array<Reference>` over `Reference[]` or even `Iterable<Reference>`?

src/test/test-utils.ts
```diff
@@ -50,3 +54,49 @@ export class TestUrlLoader implements UrlLoader {
...
+                  Array<Reference>): Promise<string|string[]> {
+    if (!Array.isArray(references)) {
+      if (references === undefined) {
+        return 'No source range given.';
+      }
+      const sourceRange =
+          isWarning(references) ? references.sourceRange : references;
+      return '\n' + await this.warningPrinter.getUnderlinedText(sourceRange);
+    }
+    return Promise.all(references.map(wOrS => this.underline(wOrS)));
```
justinfagnani: parens around params. Also, consider naming `ref`.

src/test/test-utils.ts
```diff
@@ -50,3 +54,49 @@ export class TestUrlLoader implements UrlLoader {
...
+  }
+}
+
+function isWarning(wOrS: Warning|SourceRange): wOrS is Warning {
```
justinfagnani: type this as accepting a `Reference`?

src/test/test-utils.ts
rictic: Switched to `Reference[]`. Previous version didn't have the declared type.

src/test/test-utils.ts
rictic: done

src/test/test-utils.ts
rictic: It doesn't accept undefined.

PR #446, opened 2017-02-07
Fix caching of warnings for failed scanning of imports

TimvdLippe:
This fixes a TODO in the code. I am not 100% this is the correct fix, but given the comment regarding `resolve` I think this code should live in the `resolve` of `Import`. Note that since `html-script-tag` contains duplicate code, the same code is added twice. This can be fixed later I suppose.

### DIFF
--- a/src/html/html-script-tag.ts
+++ b/src/html/html-script-tag.ts
@@ -13,6 +13,7 @@
  */
 
 import {Document, Import, ScannedImport} from '../model/model';
+import {Severity} from '../warning/warning';
 
 /**
  * <script> tags are represented in two different ways: as inline documents,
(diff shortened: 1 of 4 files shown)


src/html/html-script-tag.ts
```diff
@@ -49,6 +50,12 @@ export class ScannedScriptTagImport extends ScannedImport {
           this.warnings);
     } else {
       // not found or syntax error
+      document.warnings.push({
+        code: 'could-not-load',
+        message: `Unable to load import: ${this.error!.message || this.error}`,
```
rictic: This throws if `this.error` is undefined, probably better to handle that case?

src/model/import.ts
```diff
@@ -58,14 +60,23 @@ export class ScannedImport implements Resolvable {
...
-                                   this.type,
-                                   importedDocument,
-                                   this.sourceRange,
-                                   this.urlSourceRange,
-                                   this.astNode,
-                                   this.warnings);
+    if (!importedDocument) {
+      document.warnings.push({
+        code: 'could-not-load',
+        message: `Unable to load import: ${this.error!.message || this.error}`,
```
rictic: Ditto

src/html/html-script-tag.ts
TimvdLippe: I was under the impression when the scanned document was not found, it failed to analyze. Nonetheless will add an if check for it 😄

TimvdLippe:
Rebased and fixed handling of undefined error.

rictic:
LGTM!

PR #447, opened 2017-02-07
Pull out telemetry tracking experiment.

rictic:
It added overhead but didn't propertly account for where async time was going. We'd need something like zones.js to do this properly.

Also bumped up memory threshold in the benchmark for now.


rictic:
+1

PR #448, opened 2017-02-07
ReferenceScanner should descend into <template>s

### LINKED ISSUE #419: HtmlCustomElementReferenceScanner does not traverse past template
### Reproduction steps
1. The following script `element-reference.js`

```js
'use strict';

const Analyzer = require('polymer-analyzer').Analyzer;
const FSUrlLoader = require('polymer-analyzer/lib/url-loader/fs-url-loader').FSUrlLoader;
const PackageUrlResolver = require('polymer-analyzer/lib/url-loader/package-url-resolver').PackageUrlResolver;
const HtmlCustomElementReferenceScanner = require('polymer-analyzer/lib/html/html-element-reference-scanner').HtmlCustomElementReferenceScanner;

let analyzer = new Analyzer({
  urlLoader: new FSUrlLoader('./'),
  urlResolver: new PackageUrlResolver(),
  scanners: new Map([['html', [new HtmlCustomElementReferenceScanner()]]])
});

module.exports = () => {
  analyzer.analyze('content.html')
  .then((document) => {
    console.log(document.getByKind('element-reference'));
  });
};
```
2. The following file named `content.html`:
```html
<dom-module>
  <template>
    <my-custom-element></my-custom-element>
  </template>
</dom-module>
```

3. Run the script with `node element-reference.js`

### Expected results
```js
Set {
  ElementReference {
  attributes: [],
  warnings: [],
  kinds: Set { 'element-reference' },
  tagName: 'my-custom-element',
  sourceRange: { file: 'content.html', start: [Object], end: [Object] },
  astNode: 
   { nodeName: 'my-custom-element',
     tagName: 'my-custom-element',
     attrs: [],
     namespaceURI: 'http://www.w3.org/1999/xhtml',
     childNodes: [],
     parentNode: [Object],
     __location: [Object] } } }
```

### Actual results
```js
Set {}
```

Note that removing the template does show the correct result. E.g. this `content.html` is correctly analyzed:
```html
<dom-module>
  <my-custom-element></my-custom-element>
</dom-module>
```

rictic:

Fixes #419

### DIFF
--- a/src/html/html-element-reference-scanner.ts
+++ b/src/html/html-element-reference-scanner.ts
@@ -37,7 +37,7 @@ export class HtmlElementReferenceScanner implements HtmlScanner {
       Promise<ScannedElementReference[]> {
     const elements: ScannedElementReference[] = [];
 
-    await visit((node) => {
+    const visitor = (node: ASTNode) => {
       if (node.tagName && this.matches(node)) {
         const element = new ScannedElementReference(
             node.tagName, document.sourceRangeForNode(node)!, node);
@@ -58,7 +58,18 @@ export class HtmlElementReferenceScanner implements HtmlScanner {
 
         elements.push(element);
       }
-    });
+
+      // Descend into templates.
+      if (node.tagName === 'template' && node['content']) {
+        const content = node['content'] as ASTNode;
+        dom5.nodeWalk(content, (n) => {
+          visitor(n);
+          return false;
+        });
+      }
+    };
+
+    await visit(visitor);
 
     return elements;
   }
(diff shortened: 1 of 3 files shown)


src/html/html-element-reference-scanner.ts
```diff
@@ -58,7 +58,18 @@ export class HtmlElementReferenceScanner implements HtmlScanner {
 
         elements.push(element);
       }
-    });
+
+      // Descend into templates.
+      if (node.tagName === 'template' && node['content']) {
```
justinfagnani: I'm debating with myself right now on whether or not `visit` should be visiting template contents automatically. From an "it's an AST" perspective, it's odd that it doesn't. From a "it's DOM, and traverses with .childNodes" perspective, it's probably ok that it doesn't.

rictic: Yeah, I mean, looking at the other HTML scanners, I think this is the only one that actually does want to descend into templates. Most of the others wouldn't really ever see any use inside of templates so it doesn't matter much, but it does hint at which way the default should point.

PR #449, opened 2017-02-07
Add a query options object for all Queryable methods.

### LINKED ISSUE #372: getWarnings(deep=false) should return warnings for inline documents

rictic:
By default we do not look in dependencies, and we do look in inline documents.

Ironically, it is currently much faster to look in dependencies than not to.

Lays the foundation for more query options, on documents and on packages.

Fixes #372

### DIFF
--- a/src/demo/polymer-lint.ts
+++ b/src/demo/polymer-lint.ts
@@ -40,7 +40,7 @@ async function getWarnings(analyzer: Analyzer, localPath: string):
     Promise<Warning[]> {
       try {
         const document = await analyzer.analyze(localPath);
-        return document.getWarnings({lookInDependencies: false});
+        return document.getWarnings({imported: false});
       } catch (e) {
         if (e instanceof WarningCarryingException) {
           return [e.warning];
--- a/src/demo/polymer-lint.ts
+++ b/src/demo/polymer-lint.ts
@@ -40,7 +40,7 @@ async function getWarnings(analyzer: Analyzer, localPath: string):
     Promise<Warning[]> {
       try {
         const document = await analyzer.analyze(localPath);
-        return document.getWarnings();
+        return document.getWarnings({lookInDependencies: false});
       } catch (e) {
         if (e instanceof WarningCarryingException) {
           return [e.warning];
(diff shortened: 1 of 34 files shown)


src/demo/polymer-lint.ts
```diff
@@ -40,7 +40,7 @@ async function getWarnings(analyzer: Analyzer, localPath: string):
     Promise<Warning[]> {
       try {
         const document = await analyzer.analyze(localPath);
-        return document.getWarnings();
+        return document.getWarnings({lookInDependencies: false});
```
justinfagnani: bikeshed time: that's kind of a long option name. `followImports`, `imported`, `transitive`, `deep`? `deep` might be too short, but on the other hand we can document the difference between `deep` and `nested`.

src/editor-service/local-editor-service.ts
```diff
@@ -65,7 +65,8 @@ export class LocalEditorService extends EditorService {
     }
     if (location.kind === 'tagName' || location.kind === 'text') {
       const elements =
-          Array.from(document.getByKind('element')).filter(e => e.tagName);
+          Array.from(document.getByKind('element', {lookInDependencies: true}))
+              .filter(e => e.tagName);
```
justinfagnani: nit: parens around params

src/model/document.ts
```diff
@@ -97,6 +97,14 @@ export interface FeatureKinds {
   'js-import': Import;
 }
 
+export type QueryOptions = object & {
```
justinfagnani: this reads a little funny vs an interface. It's a shame an interface can't extend `object`, and that non-objects satisfy the implicit interface of empty classes.

src/demo/polymer-lint.ts
rictic: `deep` suggests that it might look inside `<template>`s.

`imported` is punchy and clear. I like it.

src/editor-service/local-editor-service.ts
rictic: Added tslint rule, ran `tslint --fix` to fix here and elsewhere.

src/model/document.ts
rictic: Yep, agreed.

Better options wanted, but this is the best I've come up with. ¯\\\_(ツ)\_/¯

PR #450, opened 2017-02-07
Allow filtering out features from code external to the current package.

rictic:
Builds on top of #449 

### DIFF
--- a/src/model/package.ts
+++ b/src/model/package.ts
@@ -17,6 +17,18 @@ import {Document, FeatureKinds} from './document';
 import {Feature} from './feature';
 import {Queryable} from './queryable';
 
+export type QueryOptions = object & {
+  /**
+   * If true then results will include features from outside the package, e.g.
+   * from files in bower_components or node_modules directories.
+   */
+  includeExternalCode?: boolean;
+};
+
+// A regexp that matches paths to external code.
+// TODO(rictic): Make this extensible (polymer.json?).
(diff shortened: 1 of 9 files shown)


src/model/package.ts
```diff
@@ -17,6 +17,18 @@ import {Document, FeatureKinds} from './document';
 import {Feature} from './feature';
 import {Queryable} from './queryable';
 
+export type QueryOptions = object & {
+  /**
+   * If true then results will include features from outside the package, e.g.
+   * from files in bower_components or node_modules directories.
+   */
+  includeExternalCode?: boolean;
```
justinfagnani: I'd suggest `includeExternalLibs`, `externalLibs`, `includePackages`, or just `packages`. It'd be nice to align this with the `imports` option name, even if they're not now used together, or never will.

rictic: What about `external`?

rictic: Agreed on `externalPackages` on slack.

rictic:
Made `externalPackages` an option on Document queries too.

justinfagnani:
Cool.

I wonder if ultimately we want to simplify the get & query API to just:

```typescript
interface query {
  id?: string,
  kind?: string,
  imports?: boolean,
  externalPackages?: boolean,
  inlineDocuments?: boolean,
}

getFeature(query: Query): Set<Feature>;
```

PR #451, opened 2017-02-08
More consistent behavior reference determination.

### LINKED ISSUE #440: failed to resolve Behavior as function()
If a behavior is defined like this

`behaviors: [ 
     MyBehavior({ prop : "value"})
],`

analyzer failed with

> [unknown-polymer-behavior] - Unable to resolve behavior `UNKNOWN`

and polymer build failed

rictic:
Fixes #440

### DIFF
--- a/src/polymer/declaration-property-handlers.ts
+++ b/src/polymer/declaration-property-handlers.ts
@@ -16,14 +16,47 @@ import * as estree from 'estree';
 
 import * as astValue from '../javascript/ast-value';
 import {JavaScriptDocument} from '../javascript/javascript-document';
-import {Severity} from '../warning/warning';
+import {ScannedBehaviorAssignment} from '../polymer/behavior';
+import {Severity, Warning} from '../warning/warning';
 
 import {analyzeProperties} from './analyze-properties';
 import {ScannedPolymerElement} from './polymer-element';
 
+export type BehaviorAssignmentOrWarning = {
+  kind: 'warning',
+  warning: Warning
+} | {kind: 'behaviorAssignment', assignment: ScannedBehaviorAssignment};
+
+export function getBehaviorAssignmentOrWarning(
+    argNode: estree.Node,
+    document: JavaScriptDocument): BehaviorAssignmentOrWarning {
+  const behaviorName = astValue.getIdentifierName(argNode);
+  if (!behaviorName) {
+    return {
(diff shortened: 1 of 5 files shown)


src/polymer/declaration-property-handlers.ts
```diff
@@ -16,15 +16,48 @@ import * as estree from 'estree';
...
+        message:
+            `Could not determine behavior name from expression of type ${argNode
+                .type}`,
+        severity: Severity.WARNING,
+        sourceRange: document.sourceRangeForNode(argNode)!
+      }
+    };
+  }
+  return {
+    kind: 'behaviorAssignment' as 'behaviorAssignment',
```
FredKSchott: is this type cast needed? Seems strange but I haven't tried running it.

src/polymer/declaration-property-handlers.ts
```diff
@@ -16,15 +16,48 @@ import * as estree from 'estree';
...
 
 import {analyzeProperties} from './analyze-properties';
 import {ScannedPolymerElement} from './polymer-element';
 
+export type BehaviorAssignmentOrWarning = {
+  kind: 'warning',
+  warning: Warning
+} | {kind: 'behaviorAssignment', assignment: ScannedBehaviorAssignment};
+
+export function getBehaviorAssignmentOrWarning(
```
FredKSchott: This design with two possible return types is interesting to see in TypeScript. I think it's fine as written, but in the future it might be more clear to only return a single behavior assignment type on success, and throw an exception on failure with the `warning` data on it.

src/polymer/declaration-property-handlers.ts
rictic: It was needed before I declared the BehaviorAssignmentOrWarning type explicitly. Not needed anymore, removed.

src/polymer/declaration-property-handlers.ts
rictic: I prefer this style when I want to be sure that the caller is deliberate about both cases. Exceptions are tricky because they're so invisible, usually undocumented, and generally easy to forget to handle.

FredKSchott: Those are good points! I wish TypeScript had the ability to describe the expected errors that may be thrown by a fn.

Issue #458, opened 2017-02-10
The linter's integration test throws warnings from trouble analyzing .ts files in hydrolysis (from bower).


rictic:
There's a good chance that this is just an issue with hydrolysis publishing .ts files to bower in a strange state or something, but in the linter's integration test we pull down a ton of code from bower and run analyzePackage over the bower_components dir itself. As of alpha.23 it prints warnings like:

```
declare module 'escodegen' {
~

hydrolysis/custom_typings/escodegen.d.ts(0,0) error [unable-to-analyze] - Unable to analyze file: Cannot read property 'length' of null

declare module 'espree' {
~

hydrolysis/custom_typings/espree.d.ts(0,0) error [unable-to-analyze] - Unable to analyze file: Cannot read property 'length' of null

declare module 'estraverse' {
~

hydrolysis/custom_typings/estraverse.d.ts(0,0) error [unable-to-analyze] - Unable to analyze file: Cannot read property 'length' of null

/**
~

hydrolysis/src/analyzer.ts(0,0) error [unable-to-analyze] - Unable to analyze file: Cannot read property 'length' of null
```

rictic:
Ok, looked into this a bit more. We're running into module resolution issues in the typescript analyzer. hydrolysis's ts has imports like:

```typescript
// from url-resolver.ts
import * as resolver from './resolver';
```

One issue is a simple bug that resolves that module reference to `url-resolver/url-resolver.ts/resolver` rather than `url-resolver/resolver`. The other, trickier issue is that even after that, that module specifier doesn't have a file extension. That's not fatal, as it looks like there's supposed to be some component that finds the right file extension to use, but I don't think we have it hooked up, so the typescript analyzer crashes because it can't determine the file extension.

justinfagnani:
Thanks for looking into that. The way I wrote the JS and TS import scanners assumes web-compatible module specifiers - that they actually resolve to a file. We don't support node style resolution or even implicit extensions. That could be made more clear somewhere in the docs.

Given that support for modules is slowly coming along and not advertized, I don't think source with JS imports should be in our integration tests yet.

rictic:
Fair. I'm happy to mark all of the hydrolysis files as ignored in the linter's integration test. (its integration "test" consists of bower installing a ton of polymer elements, then trying to lint them. it's also only advisory at this point, it won't fail continuous integration, but it's helpful when writing new lint rules to see if/when our own code fails it).

PR #459, opened 2017-02-10
Fix minor bug in module resolution.

rictic:
- [x] CHANGELOG.md has not been updated, this is a minor bugfix on an unreleased feature.

Makes progress on, but doesn't solve #458


justinfagnani:
Can we get some test coverage on this? A test with two files should be enough.

rictic:
It's a little tricky to test, as typescript doesn't want module specifiers to have a ts extension, and we don't have things hooked up yet to analyze javascript with the typescript analyzer (or do we?).

PR #460, opened 2017-02-11
Find local ids in dom of polymer elements

TimvdLippe:
Add support for the local id map of Polymer elements per https://www.polymer-project.org/1.0/docs/devguide/local-dom#node-finding

### DIFF
--- a/src/test/polymer/dom-module-scanner_test.ts
+++ b/src/test/polymer/dom-module-scanner_test.ts
@@ -1,0 +1,55 @@
+/**
+ * @license
+ * Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at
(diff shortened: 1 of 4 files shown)


src/test/polymer/dom-module-scanner_test.ts
```diff
@@ -0,0 +1,55 @@
...
+    });
+
+    test('finds local IDs', async() => {
+      const contents = `<html><head></head>
+        <body>
+          <dom-module>
+            <template>
+              <div id="foo"></div>
+              <span id="bar"></div>
+              <div id2="nope"></div>
```
rictic: Add a nested template with another id inside, as an assertion that those aren't captured (as intended!).

rictic:
I rebased and added the additional assertion. Merged as 77a199bff7670b35d916055539294a73eaa1191c

Thanks for the great PR!

Issue #432, opened 2017-01-19
Polymer.Behavior annotation needs to be placed on a strange spot in order to pass analyzer


Bubbit:
## issue
I'm experiencing some strange behavior with behaviors while using the polymer-build node module, the issue doesn't occur when using the latest polymer-cli.
The issue is that for a certain behavior I can't place the annotation at the normal spot: right before the behavior is specified.
```javascript
<script>
  window.TEST = window.TEST || /* istanbul ignore next */ {};
  (function (NS) {
    /* global Redux:false, PolymerRedux:false */
    var composeEnhancers = window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || Redux.compose;
    var store = Redux.createStore(
      NS.rootReducer,
      composeEnhancers(Redux.applyMiddleware.apply(this, NS.reduxMiddleware))
    );
    /* eslint-disable no-param-reassign */
    /** @polymerBehavior TEST.ReduxBehavior */
    NS.ReduxBehavior = PolymerRedux(store);
  }(TEST));
</script>
```
Instead I have to place it before the IIFE.
```javascript
<script>
  /** @polymerBehavior TEST.ReduxBehavior */
  window.TEST = window.TEST || /* istanbul ignore next */ {};
  (function (NS) {
    /* global Redux:false, PolymerRedux:false */
    var composeEnhancers = window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || Redux.compose;
    var store = Redux.createStore(
      NS.rootReducer,
      composeEnhancers(Redux.applyMiddleware.apply(this, NS.reduxMiddleware))
    );
    /* eslint-disable no-param-reassign */
    NS.ReduxBehavior = PolymerRedux(store);
  }(TEST));
</script>
```

## node versions:
node: 6.9.2
polymer-build: 0.6.0
polymer-cli: 0.18.0-alpha.8
polymer-analyzer: 2.0.0-alpha.22

## Example repos:
I created a small example repo with explanation:
- Working version: https://github.com/Bubbit/polymerBuildTesting/tree/behavior-works
- Broken version: https://github.com/Bubbit/polymerBuildTesting/tree/behavior

lassecapel:
I was able to narrow this issue down; the behavior declaration is not accepting callExpressions to be the right side of the assignment.

I was able to reproduce this issue in the analyzer by adding the following code to the test suite;

```js
// in: static/js-behaviors.js
function getFunctionBehavior() {
[...]

PR #461, opened 2017-02-12
Fix scanning of function calls to previously defined behaviors

TimvdLippe:
This should at least support the base case of defining behaviors with function calls. There are some edge cases:

- The function call must be the first polymer behavior declaration after the function is defined. Else it will not be able to find the function
- The function must live in the same document (this is a bummer, but the fix uses AST traversal and memoizes function declarations)
- The function must have the return statement as last statement and it must return an object.

Therefore this should fix the issue described in https://github.com/Polymer/polymer-analyzer/issues/432#issuecomment-276699368 but not the issue of the OP, since this uses a function which is globally defined.

### DIFF
--- a/src/test/static/js-behaviors.js
+++ b/src/test/static/js-behaviors.js
@@ -22,3 +22,35 @@ Really.Really.Deep.Behavior = [
 */
 CustomBehaviorList =
     [SimpleBehavior, CustomNamedBehavior, Really.Really.Deep.Behavior];
+
+function unhoistedGenerateBehaviorBefore() {
+  return {
+    properties: {
+      generatedBy: {
+        value: 'unhoistedGenerateBehavior'
+      }
+    }
+  };
+}
+
+/**
+ * Call to a function declared before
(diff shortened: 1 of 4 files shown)


src/test/static/js-behaviors.js
```diff
@@ -22,3 +22,35 @@ Really.Really.Deep.Behavior = [
...
+      generatedBy: {
+        value: 'unhoistedGenerateBehavior'
+      }
+    }
+  };
+}
+
+/**
+ * Call to a function declared before
+ * @polymerBehavior
```
rictic: It doesn't look like you're checking for `@polymerBehavior` annotations. Do you get the same test results without them?

I'm worried about false positives here. Mind adding a number of examples of function declarations and calls that should not be matched?

TimvdLippe:
Closing this per https://github.com/Polymer/polymer-analyzer/issues/432#issuecomment-279437125 Haven't had the time to work on the fix I describe there :cry:

PR #463, opened 2017-02-14
Add Polymer2MixinScanner

justinfagnani:

### DIFF
--- a/src/javascript/jsdoc.ts
+++ b/src/javascript/jsdoc.ts
@@ -74,6 +74,11 @@ function parsePolymerElement(tag: doctrine.Tag): Tag {
   return {tag: tag.title, type: null, name: tag.description, description: null};
 }
 
+// @polymerMixin [name]
+function parsePolymerMixin(tag: doctrine.Tag): Tag {
+  return {tag: tag.title, type: null, name: tag.description, description: null};
+}
+
 // @polymerBehavior [name]
 function parsePolymerBehavior(tag: doctrine.Tag): Tag {
   return {tag: tag.title, type: null, name: tag.description, description: null};
@@ -89,7 +94,8 @@ const CUSTOM_TAGS: {[name: string]: (tag: doctrine.Tag) => Tag} = {
   hero: parseHero,
   polymerBehavior: parsePolymerBehavior,
   polymerElement: parsePolymerElement,
-  pseudoElement: parsePseudoElement
+  polymerMixin: parsePolymerMixin,
+  pseudoElement: parsePseudoElement,
 };
 
 /**
--- a/src/model/element-mixin.ts
+++ b/src/model/element-mixin.ts
@@ -1,0 +1,36 @@
+/**
+ * @license
+ * Copyright (c) 2015 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at
+ * http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import {Element, ScannedElement} from './element';
+import {Document} from './model';
+
+export {Attribute} from './element';
+
+export class ScannedElementMixin extends ScannedElement {
+  get name() {
+    return this.tagName;
+  }
+
+  resolve(_document: Document): Element {
+    const element = new ElementMixin();
+    Object.assign(element, this);
+    return element;
+  }
+}
+
+export class ElementMixin extends Element {
+  get name() {
+    return this.tagName;
+  }
+}
--- a/src/polymer/polymer-element.ts
+++ b/src/polymer/polymer-element.ts
@@ -35,6 +35,7 @@ export interface BasePolymerProperty {
 
 export interface ScannedPolymerProperty extends ScannedProperty,
                                                 BasePolymerProperty {}
+
 export interface PolymerProperty extends Property, BasePolymerProperty {}
 
 export interface ScannedFunction extends ScannedPolymerProperty {
@@ -80,9 +81,7 @@ export interface Options {
   sourceRange: SourceRange|undefined;
 }
 
-export type Constructor<T extends object> = new (...args: any[]) => T;
-
-export interface ScannedPolymerExtension {
+export interface ScannedPolymerExtension extends ScannedElementBase {
   properties: ScannedPolymerProperty[];
   observers: {
     javascriptNode: estree.Expression | estree.SpreadElement,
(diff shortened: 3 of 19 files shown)


src/model/element-mixin.ts
```diff
@@ -0,0 +1,36 @@
...
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import {Element, ScannedElement} from './element';
+import {Document} from './model';
+
+export {Attribute} from './element';
+
+export class ScannedElementMixin extends ScannedElement {
+  get name() {
```
rictic: What's with this name property?

src/javascript/jsdoc.ts
```diff
@@ -74,6 +74,11 @@ function parsePolymerElement(tag: doctrine.Tag): Tag {
   return {tag: tag.title, type: null, name: tag.description, description: null};
 }
 
+// @polymerMixin [name]
+function parsePolymerMixin(tag: doctrine.Tag): Tag {
+  return {tag: tag.title, type: null, name: tag.description, description: null};
```
rictic: How does this behave with annotations that include a description, like

```javascript
/**
 * @polymerMixin MyMixin A mixin that adds random yodeling through the web audio apis.
 */
const MyMixin = function(clazz) {...};
```

or even:

```javascript
/**
 * @polymerMixin A mixin that adds random yodeling through the web audio apis.
 */
const MyMixin = function(clazz) {...};
```

src/polymer/polymer-element.ts
```diff
@@ -67,7 +67,7 @@ export interface Options {
   events?: ScannedEvent[];
 
   abstract?: boolean;
-  sourceRange: SourceRange|undefined;
+  sourceRange?: SourceRange|undefined;
```
rictic: Why loosen this?

src/polymer/polymer2-config.ts
```diff
@@ -0,0 +1,43 @@
+import * as estree from 'estree';
```
rictic: Need license

src/polymer/polymer2-config.ts
```diff
@@ -0,0 +1,43 @@
...
+
+import {getIdentifierName} from '../javascript/ast-value';
+import {getPropertyValue} from '../javascript/esutil';
+import {JavaScriptDocument} from '../javascript/javascript-document';
+
+import {analyzeProperties} from './analyze-properties';
+import {ScannedPolymerProperty} from './polymer-element';
+
+export function getConfig(node: estree.ClassDeclaration|estree.ClassExpression):
```
rictic: Document this function. What's a config in this context?

rictic: Document this function. What's a config in this context?

src/polymer/polymer2-config.ts
```diff
@@ -0,0 +1,43 @@
...
+  const configBody = config.value.body;
+  if (configBody.body.length !== 1) {
+    // not a single statement function
+    return null;
+  }
+  if (configBody.body[0].type !== 'ReturnStatement') {
+    return null;
+  }
+
+  const returnStatement = configBody.body[0] as estree.ReturnStatement;
```
rictic: Here and below with returnValue, if you do the assignment above the if statement then you don't need the type cast.

src/polymer/polymer2-mixin-scanner.ts
```diff
@@ -0,0 +1,140 @@
...
+    const docs = jsdoc.parseJsdoc(comment);
+    if (this._hasPolymerMixinDocTag(docs)) {
+      const name = node.id.name;
+      this._currentMixinFunction = node;
+      this._currentMixin = new ScannedPolymerElementMixin({
+        tagName: name,
+      });
+      this._currentMixinNode = node;
+      this._currentMixin.tagName = name;
+      this._mixins.push(this._currentMixin);
```
rictic: Here and elsewhere, we should give the best source range that we can for the mixin. When we add the mixin to the array, it seems like a good time to check to see if it has a source range yet, and if not, give it the best one available.

src/polymer/polymer2-mixin-scanner.ts
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  enterVariableDeclaration(
+      node: estree.VariableDeclaration, _parent: estree.Node) {
+    const comment = esutil.getAttachedComment(node) || '';
+    const docs = jsdoc.parseJsdoc(comment);
+    const isMixin = this._hasPolymerMixinDocTag(docs);
+    if (isMixin) {
+      this._currentMixin = new ScannedPolymerElementMixin();
+      this._currentMixinNode = node;
```
rictic: Or maybe the thing to do is to set the source range here when you initially create the mixin, and then if we encounter something better in our walk later it'll overwrite it?

src/test/polymer/polymer2-mixin-scanner_test.ts
```diff
@@ -0,0 +1,137 @@
...
+import * as fs from 'fs';
+import * as path from 'path';
+
+import {Visitor} from '../../javascript/estree-visitor';
+import {JavaScriptParser} from '../../javascript/javascript-parser';
+import {ScannedFeature} from '../../model/model';
+import {ScannedPolymerElementMixin} from '../../polymer/polymer-element-mixin';
+import {Polymer2MixinScanner} from '../../polymer/polymer2-mixin-scanner';
+
+// function compareNames(a: {name?: string}, b: {name?: string}): number {
```
rictic: delete?

src/polymer/polymer2-mixin-scanner.ts
```diff
@@ -0,0 +1,140 @@
...
+
+  enterFunctionDeclaration(
+      node: estree.FunctionDeclaration, _parent: estree.Node) {
+    const comment = esutil.getAttachedComment(node) || '';
+    const docs = jsdoc.parseJsdoc(comment);
+    if (this._hasPolymerMixinDocTag(docs)) {
+      const name = node.id.name;
+      this._currentMixinFunction = node;
+      this._currentMixin = new ScannedPolymerElementMixin({
+        tagName: name,
```
rictic: I don't think we should overload tagName like this. Wouldn't it be better to just have a name property on ScannedPolymerElementMixin?

src/polymer/polymer2-mixin-scanner.ts
```diff
@@ -0,0 +1,140 @@
...
+  async scan(
+      document: JavaScriptDocument, visit: (visitor: Visitor) => Promise<void>):
+      Promise<ScannedPolymerElementMixin[]> {
+    let visitor = new MixinVisitor(document);
+    await visit(visitor);
+    return visitor._mixins;
+  }
+}
+
+class MixinVisitor implements Visitor {
```
rictic: I think we should be generous with warnings here for mixins that are explicitly declared but that we can't analyze. We could have an additional annotation that indicates that the analyzer won't be able to figure out a more complex mixin, and that it's the user's responsibility at that point to explain it to us.

If we don't, we're in danger of being opaque, mysterious and thus frustrating to use.

test/mocha.opts
```diff
@@ -1,3 +1,2 @@
 --ui tdd
 --require source-map-support/register
-lib/test/*_test.js lib/test/**/*_test.js
```
rictic: revert?

rictic: Or, hm, rather I think this should be put into the package.json

Though I've had good luck lately with using `.only`, it seems mocha fixed the bugs I was seeing with it at some point.

src/test/polymer/polymer2-mixin-scanner_test.ts
```diff
@@ -0,0 +1,137 @@
...
+//   const nameB = b && b.name;
+
+//   if (nameA == null)
+//     return (nameB == null) ? 0 : -1;
+//   if (nameB == null)
+//     return 1;
+//   return nameA.localeCompare(nameB);
+// };
+
+suite('Polymer2MixinScanner', () => {
```
rictic: Please use the underline printer to also test source ranges.

rictic:
Oh, and there's a couple of lint errors, fixable with `tslint --fix`

src/model/element-mixin.ts
justinfagnani: I wanted a better name than `tagName`. I'm going to pull out a common subclass.

src/javascript/jsdoc.ts
rictic: Please add a TODO to the code here, file a bug, or add a skipped test, just to help us keep track of it.

Ah, nm, found the TODO later on.

src/polymer/polymer-element.ts
```diff
@@ -80,10 +81,61 @@ export interface Options {
   sourceRange: SourceRange|undefined;
 }
 
+export interface ScannedPolymerExtension extends ScannedElementBase {
```
rictic: Arg, this is painful. Yeah, I can definitely see why we want mixins here.

Add a TODO to use native mixins ourselves here once we drop node 4?

src/polymer/polymer2-config.ts
```diff
@@ -0,0 +1,65 @@
...
+import {getIdentifierName} from '../javascript/ast-value';
+import {getPropertyValue} from '../javascript/esutil';
+import {JavaScriptDocument} from '../javascript/javascript-document';
+
+import {analyzeProperties} from './analyze-properties';
+import {ScannedPolymerProperty} from './polymer-element';
+
+/**
+ * Returns the object literal that defines a Polymer element configuration from
+ * a Plymer element class.
```
rictic: Plymer -> Polymer

justinfagnani: fixed in next PR

PR #466, opened 2017-02-15
Use upstream memory-streams typing.

rictic:


package.json
```diff
@@ -49,7 +49,7 @@
     "gulp-tslint": "^7.0.1",
     "gulp-typescript": "^2.13.4",
     "jsdoc-to-markdown": "^1.0.3",
-    "memory-streams": "^0.1.0",
+    "memory-streams": "^0.1.1",
```
justinfagnani: make sure the patch version is needed

rictic: It is, or this doesn't compile.

PR #467, opened 2017-02-17
First, basic pass at scanning databinding expressions.

rictic:
- [x] CHANGELOG.md not updated, this scanner isn't hooked up anywhere, and still needs more work.

Needed for the linter. For example: https://github.com/Polymer/polymer-linter/issues/29 https://github.com/Polymer/polymer-linter/issues/25 https://github.com/Polymer/polymer-linter/issues/9 and https://github.com/Polymer/polymer-linter/issues/8 will all depend on this work or something much like it.

### DIFF
--- a/src/javascript/javascript-parser.ts
+++ b/src/javascript/javascript-parser.ts
@@ -27,19 +27,21 @@ declare class SyntaxError {
   column: number;
 }
 
+export const baseParseOptions = {
+  ecmaVersion: 8,
+  attachComment: true,
+  comment: true,
+  loc: true,
+};
+
 export class JavaScriptParser implements Parser<JavaScriptDocument> {
   parse(contents: string, url: string, inlineInfo?: InlineDocInfo<any>):
       JavaScriptDocument {
     const isInline = !!inlineInfo;
     inlineInfo = inlineInfo || {};
     let ast: Program;
-    const options = {
-      ecmaVersion: 8,
-      attachComment: true,
-      comment: true,
-      loc: true,
-      sourceType: 'script' as ('script' | 'module'),
-    };
+    const options = Object.assign(
+        {sourceType: 'script' as ('script' | 'module')}, baseParseOptions);
 
     try {
       ast = <Program>espree.parse(contents, options);
--- a/src/polymer/expression-scanner.ts
+++ b/src/polymer/expression-scanner.ts
@@ -1,0 +1,163 @@
+/**
+ * @license
+ * Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at
+ * http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import * as dom5 from 'dom5';
+import * as espree from 'espree';
+import * as estree from 'estree';
+import * as parse5 from 'parse5';
+
+import {HtmlVisitor, ParsedHtmlDocument} from '../html/html-document';
+import {HtmlScanner} from '../html/html-scanner';
+import {baseParseOptions} from '../javascript/javascript-parser';
+import {ScannedFeature, SourceRange} from '../model/model';
+import {Warning} from '../warning/warning';
+
+const p = dom5.predicates;
+const isTemplate = p.hasTagName('template');
+const matchInnerDataBindingTemplate = p.AND(
+    isTemplate,
+    p.OR(
+        p.hasAttrValue('is', 'dom-bind'),
+        p.hasAttrValue('is', 'dom-if'),
+        p.hasAttrValue('is', 'dom-repeat'),
+        p.parentMatches(p.OR(
+            p.hasTagName('dom-bind'),
+            p.hasTagName('dom-if'),
+            p.hasTagName('dom-repeat')))));
+export const isDomBindTemplate =
+    p.AND(p.hasTagName('template'), p.hasAttrValue('is', 'dom-bind'));
+export const isDomModuleTemplate = p.AND(
+    p.parentMatches(p.hasTagName('dom-module')), p.hasTagName('template'));
+
+export interface Template extends parse5.ASTNode { content: parse5.ASTNode; }
+
+export function getAllDataBindingTemplates(node: parse5.ASTNode) {
+  const toplevelTemplates = dom5.queryAll(
+      node,
+      dom5.predicates.OR(isDomBindTemplate, isDomModuleTemplate)) as Template[];
+
+  const results = new Set<Template>();
+  for (const template of toplevelTemplates) {
+    results.add(template);
+
+    const innerTemplates = dom5.queryAll(
+        template.content,
+        matchInnerDataBindingTemplate,
+        [],
+        dom5.childNodesIncludeTemplate) as Template[];
+    for (const innerTemplate of innerTemplates) {
+      results.add(innerTemplate);
+    }
+  }
+
+  return results;
+}
+
+
+/**
+ * Databinding into an attribute/property.
+ */
+export class ScannedAttributeExpression implements ScannedFeature {
+  /** The element whose attribute is being assigned to.*/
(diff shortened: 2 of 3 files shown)


rictic:
Sending this out as is while I build on top in the interest of sending out smaller, more easily reviewed chunks.

justinfagnani:
High level comment:

Polymer's expression parser has a lot of quirks. How will we model these against a parser that's much more correct? Will we scan string literals for commas and warn? Will we ban nested expressions?

At some point, when is it easier to model the idiosyncrasies by just using Polymer's set of regexes?

TimvdLippe:
This is probably unrelated to this PR, but instead of adhering to the regex of the core library I think it would be a better idea to refactor these regexes into a small automata in the core library. The automata will be a lot easier to reason about I suppose and does not have the regex quirks.

rictic:
We may want to move to Polymer's regexes at some point. It'd certainly be nice if Polymer core's algorithm was packaged up nicely.

I've thrashed about with different implementations of this a couple times though, and this seemed like a reasonable first pass that moves in the right direction. It won't be identical with to Polymer core's parsing, but we can do fairly well by parsing with espree and then warning based on strict validation of AST shapes.

This gives us a standard and well documented format for the expression parse trees that we can extend out as necessary if more syntax is added at some later date (or if an element uses e.g. polymer-expressions). I think it does not limit us if we later switch to using Polymer core's regexes, as the syntax they match should be a subset of JS.

So in short, I propose this route because it's incremental, adds no dependencies, and doesn't limit us in the medium term.

justinfagnani:
@TimvdLippe 🦄

src/javascript/javascript-parser.ts
```diff
@@ -27,19 +27,21 @@ declare class SyntaxError {
...
     let ast: Program;
-    const options = {
-      ecmaVersion: 8,
-      attachComment: true,
-      comment: true,
-      loc: true,
-      sourceType: 'script' as ('script' | 'module'),
-    };
+    const options = Object.assign(
+        {sourceType: 'script' as ('script' | 'module')}, baseParseOptions);
```
justinfagnani: is that cast necessary? try adding a type to `options`

src/polymer/expression-scanner.ts
```diff
@@ -0,0 +1,163 @@
+/**
+ * @license
+ * Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
```
justinfagnani: 2017

src/polymer/expression-scanner.ts
```diff
@@ -0,0 +1,163 @@
...
+
+import {HtmlVisitor, ParsedHtmlDocument} from '../html/html-document';
+import {HtmlScanner} from '../html/html-scanner';
+import {baseParseOptions} from '../javascript/javascript-parser';
+import {ScannedFeature, SourceRange} from '../model/model';
+import {Warning} from '../warning/warning';
+
+const p = dom5.predicates;
+const isTemplate = p.hasTagName('template');
+const matchInnerDataBindingTemplate = p.AND(
```
justinfagnani: Add comment. I thought at first this was going to include templates in `dom-module`s.

Also, any of these can be used standalone, not sure you need the separate dom-bind predicate.

src/polymer/expression-scanner.ts
```diff
@@ -0,0 +1,163 @@
...
+        p.parentMatches(p.OR(
+            p.hasTagName('dom-bind'),
+            p.hasTagName('dom-if'),
+            p.hasTagName('dom-repeat')))));
+export const isDomBindTemplate =
+    p.AND(p.hasTagName('template'), p.hasAttrValue('is', 'dom-bind'));
+export const isDomModuleTemplate = p.AND(
+    p.parentMatches(p.hasTagName('dom-module')), p.hasTagName('template'));
+
+export interface Template extends parse5.ASTNode { content: parse5.ASTNode; }
```
justinfagnani: If you add tagName: 'template', will that aid narrowing?

src/polymer/expression-scanner.ts
```diff
@@ -0,0 +1,163 @@
...
+ * Databinding into an attribute/property.
+ */
+export class ScannedAttributeExpression implements ScannedFeature {
+  /** The element whose attribute is being assigned to.*/
+  astNode: parse5.ASTNode;
+  /** The HTML element attribute that's being assigned to.*/
+  attribute: parse5.ASTAttribute;
+  sourceRange: SourceRange;
+  warnings: Warning[] = [];
+  directionality: 'down-only'|'bidirectional';
```
justinfagnani: these are new terms to describe this. I'd go with something like bindingType: one-way/two-way or '{' and '[' (which is how this is represented in core).

src/polymer/expression-scanner.ts
```diff
@@ -0,0 +1,163 @@
...
+  expressionText: string;
+  expressionAst: estree.Program;
+
+  /**
+   * If this is a bidirectional data binding, and an event name was specified
```
justinfagnani: Seems like you should capture the event name even for one-way bindings, since it might be written that way. Lint can warn on it.

src/polymer/expression-scanner.ts
```diff
@@ -0,0 +1,163 @@
...
+    const results: ScannedAttributeExpression[] = [];
+    const dataBindingTemplates = getAllDataBindingTemplates(document.ast);
+    for (const template of dataBindingTemplates) {
+      dom5.nodeWalkAll(template.content, (node) => {
+        if (node.attrs) {
+          for (const attr of node.attrs) {
+            if (!attr.value) {
+              continue;
+            }
+            const oneWayMatch = attr.value.match(/^\s*\[\[\s*(.*?)\s*\]\]\s*$/);
```
justinfagnani: How similar is this to the core regex? how about one regex and check the value of the group?

src/polymer/expression-scanner.ts
```diff
@@ -0,0 +1,163 @@
...
+              results.push(new ScannedAttributeExpression(
+                  node,
+                  attr,
+                  document.sourceRangeForAttributeValue(node, attr.name)!,
+                  'down-only',
+                  expressionText,
+                  undefined));
+            } else if (bidiMatch) {
+              let expressionText = bidiMatch[1];
+              const match = expressionText.match(/(.*)::(.*)/);
```
justinfagnani: see comment above, but I'd always try to parse the event

src/test/polymer/expression-scanner_test.ts
```diff
@@ -0,0 +1,91 @@
+/**
+ * @license
+ * Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
```
justinfagnani: 2017

src/javascript/javascript-parser.ts
rictic: It is, otherwise sourceType is inferred as `string`. Fixed in a later commit.

src/polymer/expression-scanner.ts
rictic: Done

src/polymer/expression-scanner.ts
rictic: Oh, you can use dom-if and dom-repeat as top level? Well that simplifies things a lot!

src/polymer/expression-scanner.ts
rictic: Sadly, no, as parse5.ASTNode's tagName is `string` so it doesn't know that 'template' would imply Template, only that all Templates have 'template'.

src/polymer/expression-scanner.ts
rictic: Went with `{` and `[`

src/polymer/expression-scanner.ts
rictic: Doing it this way, you get a javascript parse error if you use `::` on anything but a bidirectional match of an entire attribute. No need for the additional lint pass.

src/polymer/expression-scanner.ts
rictic: In the next PR this is significantly refactored and made more efficient to match inline expressions.

src/polymer/expression-scanner.ts
rictic: See comment above.

src/test/polymer/expression-scanner_test.ts
rictic: done

src/polymer/expression-scanner.ts
rictic: Also went with `direction` rather than `directionality`

PR #468, opened 2017-02-17
Parse inline Polymer databinding expressions.

rictic:
- Builds on top of #467
 - [x] CHANGELOG.md not updated yet, this is incomplete and not yet hooked up.

I'm not super happy with ScannedDatabindingExpression. There's too many fields that are sometimes present, sometimes absent, etc. Considering breaking it up into a few related types but I don't quite see a super clean breakdown yet.

PR #469, opened 2017-02-18
Add source ranges to expression parsing.

rictic:
Based on #468 

 - [x] CHANGELOG.md not updated quite yet, this isn't hooked up anywhere yet.


src/javascript/javascript-parser.ts
```diff
@@ -39,42 +40,78 @@ export class JavaScriptParser implements Parser<JavaScriptDocument> {
       JavaScriptDocument {
     const isInline = !!inlineInfo;
     inlineInfo = inlineInfo || {};
-    let ast: Program;
-    const options = Object.assign(
-        {sourceType: 'script' as ('script' | 'module')}, baseParseOptions);
+    const result = parseJs(contents, url, inlineInfo.locationOffset);
+    if (result.type === 'failure') {
+      throw new WarningCarryingException(result.warning);
```
justinfagnani: TODO: remove and propagate a parse result

src/test/polymer/expression-scanner_test.ts
```diff
@@ -54,23 +47,25 @@ suite('ExpressionScanner', () => {
...
+      const results = await scanForExpressions(document);
+      const expressions = results.expressions;
+
+      assert.deepEqual(results.warnings, []);
       assert.deepEqual(
           await underliner.underline(expressions.map((e) => e.sourceRange)), [
             `
             <div id="{{foo}}"></div>
-                    ~~~~~~~~~`,
+                       ~~~`,
```
justinfagnani: I'm not really sure we want to exclude the braces from the source range. What if we want to warn on a two-way binding to a non-notifying property? We'd want to underline the curly-braces.

rictic: Because the expression is pressed up against the curly braces it's easy to construct the one from the other. This seemed like the nicer default.

src/javascript/javascript-parser.ts
rictic: TODO added.

PR #470, opened 2017-02-18
Expression parsing all hooked up and usable

rictic:
All hooked up and ready to go.

Builds on #469 

A few design notes:
  * We're currently only bothering to parse and track expressions off dom-module templates, though the code knows how to handle stuff like dom-bind as well.
  * Uses espree to parse the expressions, but that's actually a very small part of this code. It'd be fairly easy to swap out a different parser. I did discover a couple places where Polymer expressions are invalid javascript: `foo.*` and `arr.0`. Other than that and two (I think legit!) errors, this parses all of PolymerElements without a problem or noticeable performance impact.

### DIFF
--- a/src/polymer/expression-scanner.ts
+++ b/src/polymer/expression-scanner.ts
@@ -49,7 +49,7 @@ export function getAllDataBindingTemplates(node: parse5.ASTNode) {
 /**
  * A databinding expression.
  */
-export class ScannedDatabindingExpression {
+export class DatabindingExpression {
   /**
    * If databinding into an attribute this is the element whose attribute is
    * assigned to. If databinding into a text node, this is that text node.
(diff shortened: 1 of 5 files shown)


src/polymer/expression-scanner.ts
```diff
@@ -314,3 +365,25 @@ function indexesToSourceRange(
...
+      content,
+      expressionSourceRange.file,
+      expressionOffset,
+      'polymer-expression-parse-error');
+  if (parseResult.type === 'success') {
+    return parseResult;
+  }
+  // The polymer databinding expression language allows for foo.0 and foo.*
+  // formats when accessing sub properties. These aren't valid JS, but we don't
+  // want to warn for them either. So just return undefined for now.
```
justinfagnani: file an issue to track

rictic: Filed as https://github.com/Polymer/polymer-analyzer/issues/477

PR #472, opened 2017-02-21
Find mixins assigned to namespaces

justinfagnani:


src/generate-elements.ts
```diff
@@ -64,52 +80,66 @@ export function validateElements(analyzedPackage: Elements|null|undefined) {
...
+    mixin: ResolvedMixin, packagePath: string): ElementMixin {
+  const metadata: ElementMixin =
+      serializeElementLike(mixin, packagePath) as ElementMixin;
+  metadata.name = mixin.name;
+  return metadata;
+}
+
+function serializeElementLike(
+    elementOrMixin: ElementOrMixin, packagePath: string): ElementLike {
+  // if (!resolvedElement.tagName) {
```
rictic: Move this up to serializeElement?

src/model/element-base.ts
```diff
@@ -58,8 +70,26 @@ export abstract class ElementBase implements Feature {
   astNode: estree.Node|null;
   kinds: Set<string> = new Set(['element']);
   warnings: Warning[] = [];
+  slots:
```
rictic: Recommend:

```typescript
'slots': Slot[] = [];
```

src/generate-elements.ts
justinfagnani: Removing. We need to document abstract elements with no tagName

src/model/element-base.ts
justinfagnani: done

PR #474, opened 2017-02-22
Extract property names, validate expressions.

rictic:
- [x] CHANGELOG.md not updated, small improvement on top of #470

### DIFF
--- a/src/polymer/expression-scanner.ts
+++ b/src/polymer/expression-scanner.ts
@@ -116,7 +116,7 @@ export class DatabindingExpression {
       col: sourceRange.start.column
     };
 
-    this._extract();
+    this._extractPropertiesAndValidate();
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/polymer/expression-scanner.ts
```diff
@@ -128,6 +138,67 @@ export class DatabindingExpression {
...
   }
+
+  private _extract() {
+    if (this._expressionAst.body.length !== 1) {
+      this.warnings.push(this._validationWarning(
+          'Expected exactly one expression', this._expressionAst));
+      return;
+    }
+    const expressionStatement = this._expressionAst.body[0]!;
+    if (expressionStatement.type !== 'ExpressionStatement') {
```
justinfagnani: we know that we're always getting a statement as opposed to an expression?

src/polymer/expression-scanner.ts
```diff
@@ -128,6 +138,67 @@ export class DatabindingExpression {
...
+      this.warnings.push(this._validationWarning(
+          'ExpressionStatement expected', expressionStatement));
```
justinfagnani: I don't think a user will understand a `ExpressionStatement`

src/polymer/expression-scanner.ts
```diff
@@ -125,6 +135,67 @@ export class DatabindingExpression {
...
+    }
+    if (expression.type === 'CallExpression') {
+      this._handleSubExpression(expression.callee);
+      for (const arg of expression.arguments) {
+        this._handleSubExpression(arg);
+      }
+      return;
+    }
+    this.warnings.push(this._validationWarning(
+        `Unexpected node type: ${expression.type
```
justinfagnani: I don't think ast node types will be useful in warnings.

src/polymer/expression-scanner.ts
rictic: Yep! Program#body is of type `Array<Statement | ModuleDeclaration>`

src/polymer/expression-scanner.ts
```diff
@@ -128,6 +138,67 @@ export class DatabindingExpression {
...
+        this.warnings.push(this._validationWarning(
+            'Only the not (!) operator is supported.', expression));
+        return;
+      }
+      expression = expression.argument;
+    }
+    this._handleSubExpression(expression);
+  }
+
+  private _handleSubExpression(expression: estree.Node) {
```
rictic: Fair, named more consistently. Also added a flag to the method to track whether call expressions are allowed.

src/polymer/expression-scanner.ts
rictic: Fair. I changed to a two sentence warning style in a couple places here. The first gives the problem in plain language, the second gives more technical detail that will be helpful to experts (including us, when people report issues).

src/polymer/expression-scanner.ts
```diff
@@ -128,6 +138,67 @@ export class DatabindingExpression {
     return correctSourceRange(
         databindingRelativeSourceRange, this.locationOffset);
   }
+
+  private _extract() {
```
rictic: Done

src/polymer/expression-scanner.ts
rictic: Changed to a two sentence pattern. Simple first sentence, more technical second sentence.

PR #480, opened 2017-02-23
properly set sourceRange and astNode on inline documents

FredKSchott:
- Quick-ish fix to properly set `sourceRange` and `astNode` on inlined documents.
- Add TODO to document a better fix for when we have time

/cc @rictic @justinfagnani


src/model/document.ts
```diff
@@ -36,9 +37,16 @@ import {SourceRange} from './source-range';
...
   warnings: Warning[];
+  isInline = false;
+
+  // A scanned document has no `sourceRange` or `astNode` by default. However,
+  // they can be set later if the ScannedDocument represents an inline doc.
+  // TODO(fks): Inherit these two properties from ParsedDocument so that they do
+  // not have to be set awkwardly after the fact.
+  sourceRange: SourceRange|undefined = undefined;
+  astNode: dom5.Node|undefined = undefined;
+  ;
```
rictic: Nit: extra semicolon

src/test/analyzer_test.ts
```diff
@@ -85,6 +85,38 @@ suite('Analyzer', () => {
...
 
+    test('analyzes inline scripts correctly', async() => {
+      const document = await analyzer.analyze(
+          'static/inline-documents/inline-documents.html');
+      const jsDocs = document.getByKind('js-document');
+      assert.equal(jsDocs.size, 1);
+      const jsDoc = jsDocs.values().next().value;
+      assert.deepEqual(jsDoc.sourceRange, {
+        file: 'static/inline-documents/inline-documents.html',
+        start: {line: 3, column: 2},
```
rictic: Use the CodeUnderliner rather than doing equality checking against source range

src/test/analyzer_test.ts
```diff
@@ -85,6 +85,38 @@ suite('Analyzer', () => {
...
+    });
+
+    test('analyzes inline styles correctly', async() => {
+      const document = await analyzer.analyze(
+          'static/inline-documents/inline-documents.html');
+      const cssDocs = document.getByKind('css-document');
+      assert.equal(cssDocs.size, 1);
+      const cssDoc = cssDocs.values().next().value;
+      assert.deepEqual(cssDoc.sourceRange, {
+        file: 'static/inline-documents/inline-documents.html',
```
rictic: ditto

src/test/analyzer_test.ts
```diff
@@ -85,6 +85,38 @@ suite('Analyzer', () => {
...
+      assert.equal(jsDocs.size, 1);
+      const jsDoc = jsDocs.values().next().value;
+      assert.deepEqual(jsDoc.sourceRange, {
+        file: 'static/inline-documents/inline-documents.html',
+        start: {line: 3, column: 2},
+        end: {line: 5, column: 11}
+      });
+      assert.isObject(jsDoc.astNode);
+      assert.equal(jsDoc.astNode!.tagName, 'script');
+      assert.equal(jsDoc.astNode!.parentNode!.tagName, 'head');
```
rictic: Why the assertion on the parent node?

FredKSchott: Chose it as a simple proxy for checking that the astNode was accurate. But having worked with the analyzer more now I feel we can assume that if the astNode is set it is accurate (at least in a larger analyzer test like this). So I'll back this out.

PR #482, opened 2017-02-23
Add basic inheritance support in PolymerElement

justinfagnani:

### DIFF
--- a/src/polymer/polymer-element.ts
+++ b/src/polymer/polymer-element.ts
@@ -239,53 +239,152 @@ function propertyToAttributeName(propertyName: string): string|null {
 
 function resolveElement(
     scannedElement: ScannedPolymerElement, document: Document): PolymerElement {
-  // TODO: Copy over all properties better. Maybe exclude known properties not
-  //   copied?
-  const clone: PolymerElement =
-      Object.assign(new PolymerElement(), scannedElement);
+  const element = new PolymerElement();
 
-  const flatteningResult = getFlattenedAndResolvedBehaviors(
-      scannedElement.behaviorAssignments, document);
+  //
+  // Superclass
+  //
+  if (scannedElement.superClass && scannedElement.superClass != 'HTMLElement') {
+    const superElements =
+        document.getById('element', scannedElement.superClass, {
+          externalPackages: true,
+          imported: true,
+        });
+    if (superElements.size === 1) {
+      const superElement = superElements.values().next().value;
+      if (!superElement.kinds.has('polymer-element')) {
+        element.warnings.push({
+          message:
+              `A Polymer element can\'t extend from a non-Polymer element: ${scannedElement
+                  .superClass}`,
+          severity: Severity.ERROR,
+          code: 'unknown-superclass',
+          // TODO(justinfagnani): we need to know where the superclass came
+          // (annotation vs extends clause) from and track its range.
+          sourceRange: null as any as SourceRange,
+        });
+      }
+      inheritFrom(element, superElement as PolymerElement);
+    } else {
+      if (superElements.size > 0) {
+        element.warnings.push({
+          message: `Unable to resolve superclass ${scannedElement.superClass}`,
+          severity: Severity.ERROR,
+          code: 'unknown-superclass',
+          // TODO(justinfagnani): we need to know where the superclass came
+          // (annotation vs extends clause) from and track its range.
+          sourceRange: null as any as SourceRange,
+        });
+      }
+    }
+  }
+
+  // TODO(justinfagnani): Copy over all properties better, or have
+  // PolymerElement wrap ScannedPolymerElement.
+  element.abstract = scannedElement.abstract;
+  element.astNode = scannedElement.astNode;
+  scannedElement.attributes.forEach((o) => element.attributes.push(o));
+  scannedElement.behaviorAssignments.forEach(
+      (o) => element.behaviorAssignments.push(o));
+  element.className = scannedElement.className;
+  scannedElement.demos.forEach((o) => element.demos.push(o));
+  element.description = scannedElement.description;
+  element.domModule = scannedElement.domModule;
+  scannedElement.events.forEach((o) => element.events.push(o));
+  element.extends = scannedElement.extends;
+  element.jsdoc = scannedElement.jsdoc;
+  scannedElement.listeners.forEach((o) => element.listeners.push(o));
+  scannedElement.observers.forEach((o) => element.observers.push(o));
+  scannedElement.properties.forEach((o) => element.properties.push(o));
+  element.scriptElement = scannedElement.scriptElement;
+  scannedElement.slots.forEach((o) => element.slots.push(o));
+  element.sourceRange = scannedElement.sourceRange!;
+  element.superClass = scannedElement.superClass;
+  element.tagName = scannedElement.tagName;
+  scannedElement.warnings.forEach((o) => element.warnings.push(o));
+
+  //
+  // Behaviors
+  //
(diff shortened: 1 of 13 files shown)


justinfagnani:
FYI, I'm not very happy with the state here, but the current code is a jungle. I might dig in and fix things, especially to make superclass inheritance and behavior inheritance use the same code paths, but on the other hand, the potential for getting bogged down in a lot of fixup at once is pretty large here. We'll be in this code a lot this week, and can clean up incrementally.

src/polymer/polymer-element.ts
```diff
@@ -239,53 +239,152 @@ function propertyToAttributeName(propertyName: string): string|null {
...
+              `A Polymer element can\'t extend from a non-Polymer element: ${scannedElement
+                  .superClass}`,
+          severity: Severity.ERROR,
+          code: 'unknown-superclass',
+          // TODO(justinfagnani): we need to know where the superclass came
+          // (annotation vs extends clause) from and track its range.
+          sourceRange: null as any as SourceRange,
+        });
+      }
+      inheritFrom(element, superElement as PolymerElement);
```
rictic: should this be in an else block?

src/polymer/polymer-element.ts
```diff
@@ -239,53 +239,152 @@ function propertyToAttributeName(propertyName: string): string|null {
...
+      if (!superElement.kinds.has('polymer-element')) {
+        element.warnings.push({
+          message:
+              `A Polymer element can\'t extend from a non-Polymer element: ${scannedElement
+                  .superClass}`,
+          severity: Severity.ERROR,
+          code: 'unknown-superclass',
+          // TODO(justinfagnani): we need to know where the superclass came
+          // (annotation vs extends clause) from and track its range.
+          sourceRange: null as any as SourceRange,
```
rictic: This will cause a lot of unknown problems down the line. Mind modifying the superClass object to track both name and the source range of the reference?

src/polymer/polymer-element.ts
```diff
@@ -239,53 +239,152 @@ function propertyToAttributeName(propertyName: string): string|null {
...
+        });
+      }
+      inheritFrom(element, superElement as PolymerElement);
+    } else {
+      if (superElements.size > 0) {
+        element.warnings.push({
+          message: `Unable to resolve superclass ${scannedElement.superClass}`,
```
rictic: Message should be more like `found multiple resolutions for ${scannedElement.superClass}` yeah?

src/polymer/polymer-element.ts
```diff
@@ -239,53 +239,152 @@ function propertyToAttributeName(propertyName: string): string|null {
...
+          severity: Severity.ERROR,
+          code: 'unknown-superclass',
+          // TODO(justinfagnani): we need to know where the superclass came
+          // (annotation vs extends clause) from and track its range.
+          sourceRange: null as any as SourceRange,
+        });
+      }
+      inheritFrom(element, superElement as PolymerElement);
+    } else {
+      if (superElements.size > 0) {
```
rictic: Consider also warning for exactly zero superElements

src/polymer/polymer-element.ts
```diff
@@ -239,53 +239,152 @@ function propertyToAttributeName(propertyName: string): string|null {
...
-  clone.events = mergeByName(
-      scannedElement.events,
-      behaviors.map((b) => ({name: b.className, vals: b.events})));
+  element.warnings = element.warnings.concat(behaviorsAndWarnings.warnings);
+
+  const behaviors = Array.from(behaviorsAndWarnings.behaviors);
+
+  // for (const behavior of behaviors) {
+  //   inheritFrom(element, behavior);
+  // }
```
rictic: ?

src/polymer/polymer-element.ts
```diff
@@ -239,53 +239,152 @@ function propertyToAttributeName(propertyName: string): string|null {
...
+/**
+ * Note: mutates `element`.
+ */
+function inheritFrom(element: PolymerElement, superElement: PolymerElement) {
+  // TODO(justinfagnani): fixup and use inheritValues, but it has slightly odd
+  // semantics currently
+
+  for (const superProperty of superElement.properties) {
+    // TODO(justinfagnani): better clone/merge
+    console.log('superProperty', superProperty);
```
rictic: console

src/polymer/polymer-element.ts
```diff
@@ -239,53 +239,152 @@ function propertyToAttributeName(propertyName: string): string|null {
...
+
+/**
+ * Note: mutates `element`.
+ */
+function inheritFrom(element: PolymerElement, superElement: PolymerElement) {
+  // TODO(justinfagnani): fixup and use inheritValues, but it has slightly odd
+  // semantics currently
+
+  for (const superProperty of superElement.properties) {
+    // TODO(justinfagnani): better clone/merge
```
rictic: better how?

rictic:
Tests complaining about a lint error. (btw the tslint vscode plugin has an auto-fix option, is pretty great)

src/polymer/polymer-element.ts
justinfagnani: Sure

justinfagnani: Done, but we have some nullability inconsistencies in our interfaces wrt SourceRange. It was defined as `SourceRange|undefined` in `ScannedFeature` and `Feature`, but it's non-null in `Warning`, which makes it hard to give a source range just derived from a feature without a non-null assertion. We should clean this up a bit to be consistent.

justinfagnani: We should also use the `Reference` class I introduced for more things.

src/polymer/polymer-element.ts
justinfagnani: done

src/polymer/polymer-element.ts
justinfagnani: done

src/polymer/polymer-element.ts
justinfagnani: done

src/polymer/polymer-element.ts
justinfagnani: Added a TODO - I want to handle superclasses, behaviors and soon mixins more similarly.

src/polymer/polymer-element.ts
justinfagnani: done

src/polymer/polymer-element.ts
justinfagnani: I was referring to a deep clone/merge. But given that overrides completely shadow the overridden, it's not necessary. Behaviors do need to merge though - they currently don't :(

src/polymer/polymer-element.ts
rictic: Re the source range nullability, i agree. In almost all cases where source range can be null came from me being paranoid (what if we don't  a configured espree to attach locations to nodes? what if you pass a node from one document to getSourceRange of another?) and we should just throw in those cases.

justinfagnani:
@rictic PTAL

rictic:
Arg, rewrote the one commit. We really need to use reviewable then.

rictic:
Oh NM, github was just being eventually consistent

justinfagnani:
I didn't `push -f` :)

justinfagnani:
It's not really a "name", it's the identifier that the other side of the reference should be looked up as. It should correspond to the second arg of `Document.getById()`

rictic:
Fair. In my head I read identifier in the AST sense. Not blocking.

PR #483, opened 2017-02-23
Add bin dir to files array in package.json

rictic:
Otherwise you can't npm install this version of analyzer.

Also prepare changelog for stamp of next release.


rictic:
Merging before review, as this is so minor

PR #486, opened 2017-02-23
add @namespace annotation support

FredKSchott:
- Creates a new 'namespace' feature to represent namespaces
- Detects namespaces tagged with `@namespace`
- Warns if multiple namespaces are defined
- Next up: adding `@memberof` support for connecting members to namespaces


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 -->
 
 <!--## Unreleased-->
+* Add `@namespace` annotation support.
```
rictic: Consider mentioning that it's a jsdoc tag. Maybe:

```
* Support the `@namespace` jsdoc tag.
```

src/polymer/namespace-scanner.ts
```diff
@@ -0,0 +1,85 @@
...
+   * Look for object assignments with @namespace in the docs.
+   */
+  enterAssignmentExpression(
+      node: estree.AssignmentExpression, parent: estree.Node) {
+    this._initNamespace(parent, () => esutil.objectKeyToString(node.left)!);
+  }
+
+  private _initNamespace(node: estree.Node, getName: () => string) {
+    const comment = esutil.getAttachedComment(node) || '';
+    // Quickly filter down to potential candidates.
```
rictic: +1

src/polymer/namespace-scanner.ts
```diff
@@ -0,0 +1,85 @@
...
+    if (!comment || comment.indexOf('@namespace') === -1) {
+      return;
+    }
+
+    const docs = jsdoc.parseJsdoc(comment);
+    const name = jsdoc.getTag(docs, 'namespace', 'name') || getName();
+    if (!name) {
+      throw new Error(`Unable to determine name for @namespace: ${comment}`);
```
rictic: We don't want to throw errors in the analyzer based on unexpected source, both because it's a bad user experience, and because the code doing the analysis may not care and want the analysis result anyways.

Instead, we should generate a Warning and propagate it up. We haven't needed it yet, but I've almost written a WarningFeature (a Feature that only carries a Warning/array of Warnings) a couple of times for scanners to emit when they can't emit their usual scanned feature type but they still found a problem.

src/polymer/namespace-scanner.ts
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    const sourceRange = this.document.sourceRangeForNode(node);
+    if (!sourceRange) {
+      throw new Error(
+          `Unable to determine sourceRange for @namespace: ${comment}`);
```
rictic: This is fine though, because this really should never happen (and we haven't yet found a case when it doesn't happen).

src/polymer/namespace-scanner.ts
```diff
@@ -0,0 +1,85 @@
...
+   * Look for object declarations with @namespace in the docs.
+   */
+  enterVariableDeclaration(
+      node: estree.VariableDeclaration, _parent: estree.Node) {
+    if (node.declarations.length !== 1) {
+      return;  // Ambiguous.
+    }
+    this._initNamespace(node, () => {
+      const id = node.declarations[0].id;
+      return esutil.objectKeyToString(id)!;
```
rictic: They're very similar, but I'd recommend ast-value.getIdentifierName here instead of objectKeyToString, only because the former was designed for this purpose, and we may extend objectKeyToString in the future in ways unexpected to this code.

src/polymer/namespace.ts
```diff
@@ -0,0 +1,79 @@
...
+  name: string;
+  kinds: Set<string>;
+  identifiers: Set<string>;
+  sourceRange: SourceRange;
+  astNode: any;
+  warnings: Warning[];
+
+  constructor(scannedNamespace: ScannedNamespace) {
+    this.name = scannedNamespace.name;
+    this.kinds = new Set(['namespace']);
```
rictic: It's nice that this maps so well onto the kinds and identifiers system.

Please add the `'namespace': Namespace` mapping to the kind map in model/document.ts

src/test/analyzer_test.ts
```diff
@@ -532,6 +532,45 @@ suite('Analyzer', () => {
...
+    ]);
+    assert.deepEqual(
+        document.getWarnings({imported: true}), [{
+          message:
+              'Found more than one namespace named ExplicitlyNamedNamespace.',
+          severity: Severity.WARNING,
+          code: 'multiple-polymer-namespaces',
+          sourceRange: {
+            file: 'static/namespaces/namespace-duplicate.js',
+            start: {column: 0, line: 5},
```
rictic: Here and elsewhere in this PR, asserting on source ranges is really brittle. If we use CodeUnderliner the code is easier to read, easier to maintain, and isn't broken by adding unrelated lines to a fixture file.

You can use assertSubset to leave out the sourceRange property of the expected object.

src/polymer/namespace.ts
FredKSchott: Yea I was really impressed as well.

Already added it to `FeatureKinds`, let me know if you meant somewhere else that I've missed.

src/test/analyzer_test.ts
FredKSchott: yea just corrected that in your last PR, will do the same here

src/polymer/namespace-scanner.ts
FredKSchott: I actually couldn't think of how name could ever be falsey here. Instead of throwing, maybe just ignoring and returning nothing is a better solution until we have a way to propagate real Warnings?

A lot of this is based on behavior in Behavior scanning, so throwing doesn't introduce a new problem.

src/polymer/namespace-scanner.ts
FredKSchott: +1 I'll update in behavior scanner as well

FredKSchott: ~~+1 I'll update in behavior scanner as well~~ Actually scratch that, it looks like `getIdentifierName` doesn't support everything that `objectKeyToString` does. I'll leave behavior scanner as is for now.

justinfagnani: Either will be fine for now since this is a variable declaration, but both do some possibly surprising things and don't handle symbols, and we need to be careful in other cases.

src/polymer/namespace-scanner.ts
```diff
@@ -0,0 +1,85 @@
...
+  /**
+   * Look for object assignments with @namespace in the docs.
+   */
+  enterAssignmentExpression(
+      node: estree.AssignmentExpression, parent: estree.Node) {
+    this._initNamespace(parent, () => esutil.objectKeyToString(node.left)!);
+  }
+
+  private _initNamespace(node: estree.Node, getName: () => string) {
+    const comment = esutil.getAttachedComment(node) || '';
```
justinfagnani: I don't think you need the fallback

src/polymer/namespace-scanner.ts
justinfagnani: It introduces a new place we need to clean up to keep the analyzer from dying on normal operation. We need to only throw on unrecoverable internal errors, there's even a bug on the behavior exception.

src/polymer/namespace.ts
```diff
@@ -0,0 +1,79 @@
...
+ */
+
+import * as estree from 'estree';
+
+import * as jsdoc from '../javascript/jsdoc';
+import {Document, Feature, Resolvable, SourceRange} from '../model/model';
+import {Severity, Warning} from '../warning/warning';
+
+/**
+ * The metadata for a Polymer namespace.
```
justinfagnani: Polymer -> JavaScript

src/polymer/namespace.ts
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  resolve(document: Document) {
+    const foundNamespaces = document.getById(
+        'namespace', this.name, {imported: true, externalPackages: true});
+    if (foundNamespaces.size > 0) {
+      document.warnings.push({
+        message: `Found more than one namespace named ${this.name}.`,
+        severity: Severity.WARNING,
+        code: 'multiple-polymer-namespaces',
```
justinfagnani: polymer -> javascript

src/test/static/namespaces/import-all.html
```diff
@@ -0,0 +1,4 @@
+<script src="namespace-named.js"></script>
```
justinfagnani: I don't think we need these html files, just test on the JS files.

src/polymer/namespace-scanner.ts
FredKSchott: This is only ever an issue for nonstandard, dynamic namespaces like `DynamicNamespace['ArrayNotation'] `. 

This logic was based on Behavior scanning, so this isn't a new problem. Instead of throwing, maybe just ignoring and returning nothing is a better solution until we have a way to propagate real Warnings? I'll leave it as-is for now.

src/polymer/namespace-scanner.ts
FredKSchott: Yea exactly, `sourceRange` was `SourceRange | undefined` and I wanted to explicitly check rather than sending it into ScannedNamespace with a `!`. Hopefully we'll never hit this, but if we do at least we'll know immediately.

src/test/static/namespaces/import-all.html
FredKSchott: I test the JS files directly during scanning, but for full analysis its necessary to analyze them all together to test for duplicates across files. Will also be useful to test `memberof` across files.

src/polymer/namespace-scanner.ts
FredKSchott: good catch

src/test/static/namespaces/import-all.html
justinfagnani: ok

PR #487, opened 2017-02-23
Don't try to load imports if the url can not be resolved by the AnalysisContext.

usergenic:
- This prevents `Import` and `HtmlScriptTag` from having warnings reported when the URL encountered is outside the package (i.e. a remote URL)
 - Helps resolve issues downstream in `polymer-build` which has been pretending it can resolve the contents of unresolved imports as empty strings, which
   - Results in `polymer-bundler` replacing `<script src="https://googlestatic.com/jquery.js"></script>` with `<script></script>` https://github.com/Polymer/polymer-bundler/issues/436
   - Chokes on `<link href='https://fonts.googleapis.com/css?family=Lato:400,900italic,900,700,700italic,400italic,300italic,300,100italic,100' rel='stylesheet' type='text/css'>` https://github.com/Polymer/polymer-build/issues/133


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 * Polymer 2.0 mixin scanner
 * [polymer] Parse polymer databinding expressions.
   Give accurate warnings on parse errors.
+* Analyzer will not attempt to load or add warnings for imports which are can not be resolved by its urlResolver.
```
rictic: s/are can not/can not/

src/analyzer.ts
```diff
@@ -98,6 +98,10 @@ export class Analyzer {
     return this._context.load(resolvedUrl, providedContents);
   }
 
+  canResolveUrl(url: string): string {
```
rictic: Consider leaving this out to keep Analyzer as simple as possible (unless there's a use case for it).

Also, I'm surprised that this type checks, as canResolveUrl returns a boolean

src/model/import.ts
```diff
@@ -59,6 +59,9 @@ export class ScannedImport implements Resolvable {
   }
 
   resolve(document: Document): Import|undefined {
+    if (!document.analyzer.canResolveUrl(this.url)) {
```
rictic: Will this cause problems downstream? The unknown import won't be present at all as a Feature

What if instead Import.document was optional?

src/analyzer.ts
usergenic: Derp about `string -> boolean`.  Fixed.

The use case is that `polymer-build` will use this to inquire about whether a dependency is external to the package or not.

Added test to express the expectation of this API.

src/model/import.ts
usergenic: I suppose it depends what we actually want from the Import object.  Right now it is very convenient for me, in `polymer-bundler` for example to test whether an import's url was imported by seeing if an import with that url exists.  However, I could probably can change the logic to see if that import exists AND if it has a document.

The reason I went with this logic was that the existing behavior was that _unloadable_ documents resulted in an `undefined` resolution instead of an Import without a `document` property.  Would it be right to continue having the failure-to-load case resolve with undefined but the unexceptional unloadable case would resolve with an Import having no document?

What would you prefer?  I can roll with either.

CHANGELOG.md
usergenic: I can are writing good best.  Fixeded it.

src/model/import.ts
usergenic: To answer your question though, I don't think the current approach causes downstream problems, given this is the existing behavior for URLs that can't be loaded.

usergenic: ... What's changed is that this makes it so we don't add warnings for remote/out-of-package imports.  The absence of ERROR severity warnings enables `polymer-build` to do the right thing and not pretend to resolve unloadable files with empty strings which is what it does erroneously right now.

justinfagnani: I think this is the right thing for now, as @usergenic says, it only remove the warning. Next step would be to preserve the feature in a way that the bundler ignores, and possibly add warnings on the _imports_ for the ones that deserve it.

src/analyzer.ts
justinfagnani: Does the bundler need this? It won't even see these imports in the Documents. Or does it need this for when it walks the AST itself.

PR #488, opened 2017-02-23
Use Polymer2ElementScanner for all class-based element scanning

justinfagnani:


src/test/polymer/polymer2-element-scanner_vanilla-elements_test.ts
```diff
@@ -0,0 +1,142 @@
...
+  });
+
+  test('Extracts attributes from observedAttributes', () => {
+    const element = elements.get('vanilla-with-observed-attributes')!;
+    assert.containSubset(element.attributes, [
+      {
+        description: 'When given the element is totally inactive',
+        name: 'disabled',
+        type: 'boolean',
+        sourceRange: {
```
rictic: CodeUnderliner? It's more legible, more maintainable, and less brittle than asserting on raw sourceRanges.

rictic:
This looks like it has all the functionality of the vanilla element scanner, and supplants it. Delete the standalone vanilla scanner (for now, until we've got a composable class handler system worked out)?

src/test/polymer/polymer2-element-scanner_vanilla-elements_test.ts
justinfagnani: I wanted to modify this test as little as possible.

PR #489, opened 2017-02-24
Let generateMetadata accept a Package or array of Documents

justinfagnani:


src/generate-elements.ts
```diff
@@ -17,28 +17,50 @@ import * as jsonschema from 'jsonschema';
...
-      features.filter((f) => f.kinds.has('element')) as ResolvedElement[];
-  const mixins =
-      features.filter((f) => f.kinds.has('element-mixin')) as ResolvedMixin[];
+    input: Package|Document[]|Feature[], packagePath: string): Elements {
+  let elements: Set<ResolvedElement>;
+  let mixins: Set<ResolvedMixin>;
+
+  if (input instanceof Array) {
+    const features = input as Array<Feature>;
+    const onlyDocuments = features.every((i) => i.kinds.has('document'));
```
rictic: Arg, get unlucky in your features and end up with different behavior. This code was written before we had Document worked out. I say drop the option to pass in Features, and take `Package|Document[]` instead.

justinfagnani: done

PR #494, opened 2017-02-24
Export namespaces in elements metadata

justinfagnani:


src/generate-elements.ts
```diff
@@ -16,34 +16,59 @@ import * as fs from 'fs';
...
   if (input instanceof Array) {
     elements = new Set();
     mixins = new Set();
+    namespaces = new Set();
     for (const document of input as Document[]) {
-      document.getByKind('element').forEach((f) => elements.add(f));
-      document.getByKind('element-mixin').forEach((f) => mixins.add(f));
+      Array.from(document.getByKind('element'))
+          .filter(_filter)
+          .forEach((f) => elements.add(f));
```
usergenic: `.forEach(elements.add)`?

src/generate-elements.ts
```diff
@@ -16,34 +16,59 @@ import * as fs from 'fs';
...
 export function generateElementMetadata(
-    input: Package|Document[], packagePath: string): Elements {
+    input: Package|Document[], packagePath: string, filter?: Filter): Elements {
+  const _filter = filter || ((_: Feature) => true);
+
   let elements: Set<ResolvedElement>;
   let mixins: Set<ResolvedMixin>;
+  let namespaces: Set<ResolvedNamespace>;
 
   if (input instanceof Array) {
```
usergenic: We could get rid of the if/else and reduce a little duplication if the loop were changed to:
```
for (const document of (input instanceof Array ? input : [input])) {
```

src/elements-format.ts
```diff
@@ -92,8 +103,7 @@ export interface ElementLike extends Feature {
   events?: Event[];
 
   /** The shadow dom content slots that this element accepts. */
-  slots:
-  Slot[];  // this formatting is strange, yes
+  'slots': Slot[];  // this formatting is strange, yes
```
usergenic: Is this a clang-format workaround?

src/generate-elements.ts
usergenic: (etc for the other 2 set .add applications.)

src/generate-elements.ts
justinfagnani: Package and Documents are treated every so slightly differently though

src/generate-elements.ts
justinfagnani: I basically never do tearoffs in JS, since they're not called with the right `this`. In this case `add` actually throws because it doesn't get the Set as the receiver.

src/elements-format.ts
justinfagnani: yep :( we should try tsfmt

PR #497, opened 2017-02-24
Escape unsafe chars in urls

### LINKED ISSUE #496: Making PackageUrlResolver the default UrlResolver caused a regression with URL encoding
Because PackageUrlResolver resolves local/package urls as unencoded versions of the pathname, there are mismatches between Import.url (encoded) and Document.url (unencoded) in analyzed document graphs etc.

The downstream effect of this is that bundler is unable to bundle imports with spaces and build is likely effected, too.  The fix is simple enough: reencoding the pathname on the way out of PackageUrlResolver.

usergenic:
When its a URL, it should be URI-encoded.  Also, `(decode|encode)URIComponent()` is too aggressive to normalize pathname.  Switched to `(decode|encode)URI()`
 - Fixes #496 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 -->
 
 <!--## Unreleased-->
+### Fixed
+* PackageUrlResolver encodes URIs returned from `resolve` method.
 
 ## [2.0.0-alpha.27] - 2017-02-24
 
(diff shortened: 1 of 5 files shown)


FredKSchott:
Makes sense to me. 

My one concern when you mentioned this was introducing problems when trying to map from the file system to the URL scheme, if the URL scheme has been encoded. I don't think this change will affect anything there, but something to keep in mind.

usergenic:
THX!

PR #500, opened 2017-02-25
add support to understand the @memberof jsdoc tag

### LINKED ISSUE #495: Use @memberof annotations to set id of features
To associate features with namespaces, @memberof needs to set the id of the feature.

Example:
```javascript
/**
 * @memberof Polymer
 */
const ElementMixin = ...;
```
The id is now `Polymer.ElementMixin`.

This should also effect the name (whatever the main name field is for a feature. Element/className, Mixin/name, Function(TODO still)/name) so that users know how to reference the feature in the global namespace.

We may also want to set some sort of "public" bit on these features so we know to keep them in the metadata output, but that'd only be necessary if there are things that are showing up in the output that we want to filter out.

FredKSchott:
- Polymer 2.0 Elements: className inherits from `@memberof` value if one is given
- Polymer 2.0 Mixins: name inherits from `@memberof` value if one is given
- Behaviors: name inherits from `@memberof` value if one is given (If no explicit full name is given to `@polymerBehavior`)
- Polymer 1.0 Elements: No change / no `@memberof` support
- Polymer 1.0 Mixins: No change / no `@memberof` support

---

- Resolves #495

/cc @justinfagnani @usergenic

### DIFF
--- a/src/javascript/ast-value.ts
+++ b/src/javascript/ast-value.ts
@@ -15,7 +15,6 @@
 import * as estree from 'estree';
 
 import {LiteralObj, LiteralValue} from '../model/model';
-import * as esutil from './esutil';
 import * as jsdoc from './jsdoc';
 
 /**
--- a/src/javascript/ast-value.ts
+++ b/src/javascript/ast-value.ts
@@ -206,9 +206,10 @@ export function getNamespacedIdentifier(
   if (!docs) {
     return name;
   }
-  const namespaceName = jsdoc.getTag(docs, 'memberof', 'description');
-  if (namespaceName) {
-    return namespaceName + '.' + name;
+  const namespace = jsdoc.getTag(docs, 'memberof', 'description');
+  if (namespace) {
+    const rightMostIdentifierName = name.substring(name.lastIndexOf('.') + 1);
+    return namespace + '.' + rightMostIdentifierName;
   } else {
     return name;
   }
--- a/src/javascript/ast-value.ts
+++ b/src/javascript/ast-value.ts
@@ -202,10 +202,11 @@ export function getIdentifierName(node: estree.Node): string|undefined {
  * Formats the given identifier name under a namespace, if one is mentioned in
  * the commentedNode's comment. Otherwise, name is returned.
  */
-export function namespaceIdentifierName(
-    name: string, commentedNode: estree.Node): string {
-  const comment = esutil.getAttachedComment(commentedNode) || '';
-  const docs = jsdoc.parseJsdoc(comment);
+export function getNamespacedIdentifier(
+    name: string, docs?: jsdoc.Annotation): string {
+  if (!docs) {
+    return name;
+  }
   const namespaceName = jsdoc.getTag(docs, 'memberof', 'description');
   if (namespaceName) {
     return namespaceName + '.' + name;
--- a/src/javascript/ast-value.ts
+++ b/src/javascript/ast-value.ts
@@ -15,6 +15,8 @@
 import * as estree from 'estree';
 
 import {LiteralObj, LiteralValue} from '../model/model';
+import * as esutil from './esutil';
+import * as jsdoc from './jsdoc';
 
 /**
  * Converts an ast literal to its underlying valie.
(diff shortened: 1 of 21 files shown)


src/javascript/ast-value.ts
```diff
@@ -196,4 +198,21 @@ export function getIdentifierName(node: estree.Node): string|undefined {
   }
 }
 
+/**
+ * Formats the given identifier name under a namespace, if one is mentioned in
+ * the commentedNode's comment. Otherwise, name is returned.
+ */
+export function namespaceIdentifierName(
```
justinfagnani: How about `getNamespacedIdentifier`?

src/javascript/ast-value.ts
```diff
@@ -196,4 +198,21 @@ export function getIdentifierName(node: estree.Node): string|undefined {
...
+    name: string, commentedNode: estree.Node): string {
```
justinfagnani: Can this take the parsed jsdoc, since presumably the caller has to access some tags and/or description anyway?

src/polymer/polymer2-element-scanner.ts
```diff
@@ -65,16 +65,25 @@ class ElementVisitor implements Visitor {
...
 
   enterClassDeclaration(node: estree.ClassDeclaration) {
     const element = this._handleClass(node);
     if (element) {
-      element.className = node.id.name;
-      this._possibleElements.set(element.className, element);
+      const className = node.id.name;
+      const namespacedClassName = namespaceIdentifierName(className, node);
+      element.className = namespacedClassName;
+      // Set the element on both the namespaced & unnamespaced names so that we
```
justinfagnani: What's the reasoning here?  That an element will be referenceable under both is namespaced and un-namespaced name? I think we want only one canonical identifier.

justinfagnani:
FYI, I have this rebased and passing locally.

src/polymer/polymer2-element-scanner.ts
FredKSchott: Polymer's core does this in a few places, and I think it's a fair use-case that we should handle:

```js
  /**
   * @memberof Polymer
   */
  class CustomStyle extends HTMLElement { ... }

  window.Polymer.CustomStyle = CustomStyle;
  window.customElements.define('custom-style', CustomStyle);
```

By adding both `Polymer.CustomStyle` & `CustomStyle` to the `_possibleElements` map for this file, we can detect when the author uses either in `customElements.define()`. It has no affect on the className of the registered element.

justinfagnani: Good point, it doesn't leak the local analysis.

justinfagnani:
Running on core, it seems like some names are getting double-namespaced:

```
  /**
   * @polymerMixin
   * @memberof Polymer
   */
  Polymer.PropertyAccessors = Polymer.dedupingMixin(function(superClass) {
```

Is resulting in a name of `Polymer.Polymer.PropertyAccessors`

Meanwhile, DomModule isn't getting a classname:

```
   * @extends HTMLElement
   * @memberof Polymer
   */
  class DomModule extends HTMLElement {
```

FredKSchott:
@justinfagnani hmm, I had thought the test suite was thorough enough. I'll try to reproduce and add tests for any bugs found by analyzing core.

FredKSchott:
@justinfagnani ptal

src/javascript/ast-value.ts
```diff
@@ -196,4 +197,23 @@ export function getIdentifierName(node: estree.Node): string|undefined {
...
+ * the commentedNode's comment. Otherwise, name is returned.
+ */
+export function getNamespacedIdentifier(
+    name: string, docs?: jsdoc.Annotation): string {
+  if (!docs) {
+    return name;
+  }
+  const namespace = jsdoc.getTag(docs, 'memberof', 'description');
+  if (namespace) {
+    const rightMostIdentifierName = name.substring(name.lastIndexOf('.') + 1);
```
justinfagnani: hmm.. ultimately I think we might want to call `getIdentifierName` (or similar) in this function, so that given jsdocs and a node to extract a name from, this contains all the logic in one place to determine the fully qualified name - rather than building up a dot-separated name in `getIdentifierName`, just to split it apart here.

Don't block though... just add a TODO

justinfagnani:
LGTM! Please rebase down is easy enough

src/javascript/ast-value.ts
FredKSchott: Yea there's definitely some places where that would be nice, but it may be too restrictive for formatting things like `namespaces` that aren't actually node's.

PR #506, opened 2017-02-25
Update version of dom5 to fix undefined function

TimvdLippe:
`dom5` 2.1.0 introduced a new function `defaultChildNodes` which is used in the `polymer/dom-module-scanner` per #470.
However the dependency was not updated in `package.json`, meaning that an installation of `polymer-analyzer` might fail and resolve the incorrect version of `dom5`.


rictic:
LGTM!

PR #510, opened 2017-02-27
Support new properties/observers getters in Polymer 2.0 elements

justinfagnani:


justinfagnani:
Nope, not missing anything. We just weren't actually handling them before. We were passing the config ObjectLiteral node only to the getProperties() function.

usergenic:
Oh...maybe the reason we don't need them shown here is because we don't have a plan for incorporating Observers in our analysis yet anyways?

usergenic:
k.

justinfagnani:
Linting will probably want that info at some point, but for now it's just internal implementation detail.

PR #512, opened 2017-02-28
Fix crash in PolymerElementScanner when scanning Polymer 2.0 classes …

justinfagnani:
…w/ observers.


justinfagnani:
@FredKSchott if you get to this. I'm probably going to submit TBR since it's a crasher when analyzing Polymer core and it's a simple fix.

FredKSchott:
Good catch. Can you add an issue to add this as a warning on `this.element` instead of just ignoring? It'll be an easy fix that we shouldn't forget about.

PR #517, opened 2017-03-01
Add mixins to metadata output

justinfagnani:

### DIFF
--- a/src/generate-elements.ts
+++ b/src/generate-elements.ts
@@ -142,6 +142,11 @@ function serializeElement(
   if (element.className) {
     metadata.classname = element.className;
   }
+
+  // TODO(justinfagnani): Mixins should be able to have mixins too
+  if (element.mixins.length > 0) {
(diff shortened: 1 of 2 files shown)


src/generate-elements.ts
```diff
@@ -142,6 +142,11 @@ function serializeElement(
   if (element.className) {
     metadata.classname = element.className;
   }
+
+  // TODO(justinfagnani): Mixins should be able to have mixins too
+  if (element.mixins && element.mixins.length > 0) {
```
FredKSchott: no need to check `!!element.mixins`, it's always an array.

src/generate-elements.ts
```diff
@@ -142,6 +142,11 @@ function serializeElement(
   if (element.className) {
     metadata.classname = element.className;
   }
+
+  // TODO(justinfagnani): Mixins should be able to have mixins too
```
FredKSchott: 🐢

PR #518, opened 2017-03-01
add support for scanning methods on polymer elements and mixins

FredKSchott:
Note that this doesn't change the data returned by analysis yet. That has been addressed separately in the next PR. CHANGELOG will be updated then since there's no visible change to analysis results.

### DIFF
--- a/src/model/method.ts
+++ b/src/model/method.ts
@@ -12,36 +12,17 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
-import * as jsdoc from '../javascript/jsdoc';
-import {SourceRange} from '../model/model';
+import {Property, ScannedProperty} from './model';
 
-import {ScannedFeature} from './model';
-
-export interface ScannedProperty extends ScannedFeature {
-  name: string;
-  type?: string;
-  private?: boolean;
-  'default'?: string;
-  readOnly?: boolean;
-  changeEvent?: string;
-  function?: {
+export interface ScannedMethod extends ScannedProperty {
+  function: {
     params?: {name: string, type?: string}[];
     return?: {type?: string, desc: string};
   };
 }
 
-export interface Property {
-  name: string;
-  type?: string;
-  description?: string;
-  jsdoc?: jsdoc.Annotation;
-  private?: boolean;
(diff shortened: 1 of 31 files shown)


src/polymer/polymer-element-scanner.ts
```diff
@@ -217,9 +233,15 @@ class ElementVisitor implements Visitor {
             getters[scannedPolymerProperty.name] = scannedPolymerProperty;
           } else if (scannedPolymerProperty.setter) {
             setters[scannedPolymerProperty.name] = scannedPolymerProperty;
+          } else if (
+              prop.method === true ||
+              prop.value.type === 'ArrowFunctionExpression' ||
+              prop.value.type === 'FunctionExpression') {
```
FredKSchott: @justinfagnani is there a better way to do this?

src/model/method.ts
```diff
@@ -0,0 +1,29 @@
...
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import {Property, ScannedProperty} from './model';
+
+export interface ScannedMethod extends ScannedProperty {
+  function: {
```
justinfagnani: why have this nested inside of a function property? Wouldn't methods just have parameters and a return value?

src/model/property.ts
```diff
@@ -21,14 +20,14 @@ import {ScannedFeature} from './model';
...
   type?: string;
-  description?: string;
-  jsdoc?: jsdoc.Annotation;
   private?: boolean;
   'default'?: string;
   readOnly?: boolean;
-  sourceRange: SourceRange|undefined;
-  astNode: estree.Node|null;
   changeEvent?: string;
+  function?: {
```
justinfagnani: since you added ScannedMethod, what's this for here?

src/polymer/polymer-element-scanner.ts
justinfagnani: Not that I know of

PR #519, opened 2017-03-01
Use new method analysis in the results elements data format

FredKSchott:

This builds on the `methods` detection work to expose the new methods in the analysis results.

### DIFF
--- a/src/elements-format.ts
+++ b/src/elements-format.ts
@@ -105,6 +105,9 @@ export interface ElementLike extends Feature {
   /** The properties that this element has. */
   properties?: Property[];
 
+  /** The instance methods that this element has. */
+  methods?: Method[];
+
   /** The events that this element fires. */
   events?: Event[];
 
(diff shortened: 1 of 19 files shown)


src/elements-format.ts
```diff
@@ -239,6 +242,35 @@ export interface Property extends Feature {
...
 
+export interface Method extends Feature {
+  /** The name of the property. e.g. `value`, `icon`, `shouldCollapse`. */
+  name: string;
+
+  /**
+   * The javascript type of function. ex: 'ArrowFunctionExpression',
+   * 'FunctionExpression'.
+   */
+  type?: string;
```
justinfagnani: Is this necessary?

FredKSchott: Nope, good catch. Removing...

src/generate-elements.ts
```diff
@@ -168,12 +168,10 @@ function serializeElementLike(
   const attributes = elementOrMixin.attributes.map(
       (a) => serializeAttribute(elementOrMixin, path, a));
   const properties =
-      elementOrMixin.properties
-          .filter(
-              (p) => !p.private &&
-                  // Blacklist functions until we figure out what to do.
-                  p.type !== 'Function')
+      elementOrMixin.properties.filter((p) => !p.private)
```
justinfagnani: This is fine for now, since we already filter private members, but I think we actually want to add them to the output and filter in the client.

PR #520, opened 2017-03-01
Fix compilation error

justinfagnani:

### DIFF
--- a/src/analyzer.ts
+++ b/src/analyzer.ts
@@ -48,8 +48,12 @@ export type LazyEdgeMap = Map<string, string[]>;
  */
 export class Analyzer {
   private _context: AnalysisContext;
-  constructor(options: Options) {
-    this._context = new AnalysisContext(options);
+  constructor(options: Options|AnalysisContext) {
+    if (options instanceof AnalysisContext) {
+      this._context = options;
+    } else {
+      this._context = new AnalysisContext(options);
+    }
   }
 
   /**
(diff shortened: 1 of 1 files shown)


src/analyzer.ts
```diff
@@ -94,9 +98,7 @@ export class Analyzer {
    * Note: this feature is experimental.
    */
   _fork(): Analyzer {
-    const analyzer = new Analyzer({});
```
justinfagnani: FYI, the error was that `{}` didn't match options because it doesn't have `urlLoader`

src/analyzer.ts
```diff
@@ -48,8 +48,12 @@ export type LazyEdgeMap = Map<string, string[]>;
  */
 export class Analyzer {
   private _context: AnalysisContext;
-  constructor(options: Options) {
-    this._context = new AnalysisContext(options);
+  constructor(options: Options|AnalysisContext) {
+    if (options instanceof AnalysisContext) {
```
FredKSchott: use `options.constructor.name === 'AnalysisContext'` to work around different instances being passed around terrible npm trees.

justinfagnani: That doesn't trigger the type narrowing in TypeScript.

src/analyzer.ts
FredKSchott: ugh silent build failures. Lets prioritize getting that fixed, want me to look into it or are you already?

PR #522, opened 2017-03-02
Add inheritedFrom to metadata output

justinfagnani:


src/elements-format.ts
```diff
@@ -216,6 +216,9 @@ export interface Attribute extends Feature {
    */
```
rictic: Please update CHANGELOG.md

justinfagnani: done

PR #523, opened 2017-03-02
add namespace function scanner

FredKSchott:
Functions tagged with `@memberof` are now returned by the new function scanner. Returning these functions via analysis is still TODO.

### DIFF
--- a/src/javascript/function-scanner.ts
+++ b/src/javascript/function-scanner.ts
@@ -14,13 +14,13 @@
 
 import * as estree from 'estree';
 
-import {getIdentifierName, getNamespacedIdentifier} from
-'../javascript/ast-value';
+import {getIdentifierName, getNamespacedIdentifier} from '../javascript/ast-value';
 import {Visitor} from '../javascript/estree-visitor';
-import {getAttachedComment, objectKeyToString, isFunctionType} from '../javascript/esutil';
+import {getAttachedComment, isFunctionType, objectKeyToString} from '../javascript/esutil';
 import {JavaScriptDocument} from '../javascript/javascript-document';
 import {JavaScriptScanner} from '../javascript/javascript-scanner';
 import * as jsdoc from '../javascript/jsdoc';
+
 import {ScannedFunction} from './function';
 
 export class FunctionScanner implements JavaScriptScanner {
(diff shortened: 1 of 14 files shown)


src/javascript/function-scanner.ts
```diff
@@ -0,0 +1,157 @@
...
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import * as estree from 'estree';
+
+import {getIdentifierName, getNamespacedIdentifier} from
+'../javascript/ast-value';
```
justinfagnani: `npm run format`?

src/javascript/function-scanner.ts
```diff
@@ -0,0 +1,157 @@
...
+
+    const docs = jsdoc.parseJsdoc(comment);
+    const functionName = getNamespacedIdentifier(anaylzedName, docs);
+    const sourceRange = this.document.sourceRangeForNode(node)!;
+    const returnTag = jsdoc.getTag(docs, 'return');
+    const summaryTag = jsdoc.getTag(docs, 'summary');
+    const summary = (summaryTag && summaryTag.description) || '';
+    const description = docs.description;
+
+    let functionReturn;
```
justinfagnani: `const` and ternary?

src/javascript/function-scanner.ts
```diff
@@ -0,0 +1,157 @@
...
+      functionReturn = {
+        type: returnTag.type || undefined,
+        desc: returnTag.description || '',
+      };
+    }
+
+    let functionParams;
+    if (docs.tags) {
+      const paramTags = docs.tags.filter((tag) => tag.tag === 'param');
+      if (paramTags.length > 0) {
```
justinfagnani: Why not just map anyway and always produce an array, even if empty?

src/javascript/function-scanner.ts
FredKSchott: hmm,  not sure how this got in there

src/javascript/function-scanner.ts
FredKSchott: Woah, clang-format does not like that:

```

    const functionReturn = (returnTag) ? {
      type: returnTag.type || undefined,
      desc: returnTag.description || '',
    } :
                                         undefined;
```

I'm going to stick with let, I agree const is great but not at the expense of significant clarity.

PR #524, opened 2017-03-02
add functions to element.json metadata format

FredKSchott:
- [X] CHANGELOG Updated

### DIFF
--- a/src/generate-elements.ts
+++ b/src/generate-elements.ts
@@ -16,7 +16,8 @@ import * as fs from 'fs';
 import * as jsonschema from 'jsonschema';
 import * as pathLib from 'path';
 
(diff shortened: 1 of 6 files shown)


src/generate-elements.ts
```diff
@@ -134,6 +147,29 @@ function serializeNamespace(
   return metadata;
 }
 
+function serializeFunction(
+    fn: ResolvedFunction, packagePath: string): Function {
+  const packageRelativePath =
+      pathLib.relative(packagePath, fn.sourceRange.file);
+  const metadata: Function = {
```
justinfagnani: should be able to share code with serializeMethod

FredKSchott: Agreed, there should be a larger refactoring step to address this across the repo. Right now one is a class and the other is a model interface, and I actually went out of my way to not share code until we had a chance to address throughout.

PR #525, opened 2017-03-02
Track privacy better on properties and methods.

rictic:

### DIFF
--- a/src/polymer/js-utils.ts
+++ b/src/polymer/js-utils.ts
@@ -17,7 +17,7 @@ import * as estree from 'estree';
 
 import {closureType, getAttachedComment, objectKeyToString} from '../javascript/esutil';
 import * as jsdoc from '../javascript/jsdoc';
-import {ScannedMethod, ScannedProperty, SourceRange} from '../model/model';
+import {ScannedMethod, SourceRange} from '../model/model';
 import {Severity, Warning} from '../warning/warning';
 
 import {ScannedPolymerProperty} from './polymer-element';
@@ -25,10 +25,11 @@ import {ScannedPolymerProperty} from './polymer-element';
 /**
  * Create a ScannedProperty object from an estree Property AST node.
  */
-export function toScannedProperty(
+export function toScannedPolymerProperty(
     node: estree.Property|estree.MethodDefinition,
-    sourceRange: SourceRange): ScannedProperty {
+    sourceRange: SourceRange): ScannedPolymerProperty {
   const type = closureType(node.value, sourceRange);
+  const parsedJsdoc = jsdoc.parseJsdoc(getAttachedComment(node) || '');
   const description =
       jsdoc.removeLeadingAsterisks(getAttachedComment(node) || '').trim();
   const name = objectKeyToString(node.key);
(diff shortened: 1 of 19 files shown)


src/polymer/polymer-element.ts
```diff
@@ -98,7 +101,8 @@ export function addProperty(
     target: ScannedPolymerExtension, prop: ScannedPolymerProperty) {
   target.properties.push(prop);
   const attributeName = propertyToAttributeName(prop.name);
-  if (prop.private || !attributeName || !prop.published) {
+  if ((prop.privacy && prop.privacy !== 'public') || !attributeName ||
+      !prop.published) {
```
rictic: Don't create events or attributes for nonpublic properties.

src/polymer/js-utils.ts
```diff
@@ -51,24 +52,54 @@ export function toScannedProperty(
...
     sourceRange: sourceRange,
     astNode: node, warnings,
-    private: !!name && (name.startsWith('_') || name.endsWith('_')),
+    isConfiguration: configurationProperties.has(name || ''),
+    jsdoc: parsedJsdoc
   };
 
-  if (node.kind === 'get' || node.kind === 'set') {
-    result.type = '';
+  const explicitPrivacy = jsdoc.getPrivacy(parsedJsdoc);
```
justinfagnani: Pull this out into an exported function and use in the next PR too

src/polymer/js-utils.ts
```diff
@@ -51,24 +52,54 @@ export function toScannedProperty(
...
+  } else if (result.name.endsWith('_')) {
+    result.privacy = 'private';
+  } else {
+    result.privacy = 'public';
   }
 
   return result;
 };
 
+/** Properties on element prototypes that are purely configuration. */
```
justinfagnani: Can you expand the comment a little? At first read is hard to tell when this set would be used for.

Also, looking at what this is actually used for, rather than have an isConfiguration bit, I might just say that these are implicitly protected, which will hide them in the docs by default.

src/polymer/polymer-element.ts
```diff
@@ -29,9 +29,12 @@ export interface BasePolymerProperty {
...
   reflectToAttribute?: boolean;
-  configuration?: boolean;
-  getter?: boolean;
-  setter?: boolean;
+  /**
+   * True if the property is part of Polymer's element configuration syntax.
+   *
+   * e.g. 'properties', 'is', 'extends', etc
+   */
+  isConfiguration?: boolean;
```
justinfagnani: I commented above, but I don't like this flag so much. I'd rather just consider these protected. Is it used elsewhere?

src/polymer/polymer-element.ts
justinfagnani: Add this comment to the code :)

src/polymer/polymer-element.ts
```diff
@@ -294,9 +298,18 @@ function resolveElement(
   // Elements have their own logic to dictate when a method is private or public
   // that overrides whatever our scanner detected.
   for (const method of element.methods) {
-    const hasJsDocPrivateTag = !!JsDocGetTag(method.jsdoc, 'private');
-    method.private =
-        !method.jsdoc || isAnnotationEmpty(method.jsdoc) || hasJsDocPrivateTag;
+    const explicitPrivacy = getPrivacy(method.jsdoc);
```
justinfagnani: Use the to-be-exported function for this.

rictic: Done.

src/polymer/polymer-element.ts
rictic: Done.

PR #528, opened 2017-03-03
Set privacy on mixins.

rictic:

### DIFF
--- a/src/elements-format.ts
+++ b/src/elements-format.ts
@@ -207,6 +207,7 @@ export interface ElementMixin extends ElementLike {  //
    * e.g. `MyMixin`, `Polymer.PaperInputMixin`
    */
   name: string;
+  privacy: 'public'|'private'|'protected';
 }
 
 export interface Attribute extends Feature {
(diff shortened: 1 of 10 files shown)


src/elements-format.ts
```diff
@@ -207,6 +207,7 @@ export interface ElementMixin extends ElementLike {  //
    * e.g. `MyMixin`, `Polymer.PaperInputMixin`
    */
   name: string;
+  privacy: 'public'|'private'|'protected';
```
justinfagnani: didn't you introduce a `Privacy` type def?

src/javascript/function-scanner.ts
```diff
@@ -114,6 +114,10 @@ class FunctionVisitor implements Visitor {
     }
 
     const docs = jsdoc.parseJsdoc(comment);
+    if (jsdoc.hasTag(docs, 'polymerMixin')) {
+      // This is a mixin, not a normal function.
```
justinfagnani: thank you! just found this in core :)

src/elements-format.ts
rictic: Not in elements-format, as the old version of our json schema generator didn't work with `type` declarations it seemed. The new one does though, fixed in the next PR.

PR #529, opened 2017-03-03
Track privacy on elements

rictic:


src/polymer/behavior-scanner.ts
```diff
@@ -152,25 +146,25 @@ class BehaviorVisitor implements Visitor {
         return;
       }
     }
+    const jsDoc = jsdoc.parseJsdoc(comment || '');
```
justinfagnani: can you name this something like `docs`. Variable names that differ only in case are confusing.

PR #531, opened 2017-03-03
generateAnalysisMetadata -> generateAnalysisMetadata

justinfagnani:

### DIFF
--- a/src/generate-analysis.ts
+++ b/src/generate-analysis.ts
@@ -35,7 +35,7 @@ interface Members {
   functions: Set<ResolvedFunction>;
 }
 
-export function generateElementMetadata(
+export function generateAnalysisMetadata(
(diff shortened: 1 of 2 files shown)


src/generate-analysis.ts
```diff
@@ -35,7 +35,7 @@ interface Members {
   functions: Set<ResolvedFunction>;
 }
 
-export function generateElementMetadata(
+export function generateAnalysisMetadata(
```
rictic: Man, `metadata` is not such a great word here. Too general.

It's kinda a serialization. Or maybe we could just call it an analysis?

Like, Documents and Packages are live and queryable, but an Analysis is a dead, serializable representation of the most interesting stuff?

justinfagnani: Can we defer that part? This was just supposed to go in the last PR

rictic: Put it into #533

PR #538, opened 2017-03-04
We had a few pieces of the public API missing from index.ts

rictic:
- [x] CHANGELOG.md not updated, bugfix.

### DIFF
--- a/src/model/source-range.ts
+++ b/src/model/source-range.ts
@@ -79,6 +79,9 @@ export function correctPosition(
 /**
  * If the position is inside the range, returns 0. If it comes before the range,
  * it returns -1. If it comes after the range, it returns 1.
+ *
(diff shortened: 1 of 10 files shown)


src/model/source-range.ts
```diff
@@ -69,9 +69,65 @@ export function correctSourceRange(
...
     line: position.line + locationOffset.line,
     column: position.column + (position.line === 0 ? locationOffset.col : 0)
   };
 }
+
+/**
+ * If the position is inside the range, returns 0. If it comes before the range,
+ * it returns -1. If it comes after the range, it returns 1.
+ */
+export function comparePositionAndRange(
```
justinfagnani: TODO add test

rictic: Done

PR #540, opened 2017-03-04
Recognize elements registered with MyElem.is

rictic:
Fixes https://github.com/Polymer/polymer-analyzer/issues/537

### DIFF
--- a/src/polymer/polymer2-element-scanner.ts
+++ b/src/polymer/polymer2-element-scanner.ts
@@ -43,16 +43,6 @@ export class Polymer2ElementScanner implements JavaScriptScanner {
   }
 }
 
-type TagNameExpression = {
-  type: 'is',
-  className: string,
-  classNameSourceRange: SourceRange,
-}|{
-  type: 'string-literal',
-  value: string,
-  sourceRange: SourceRange,
-};
-
 class ElementVisitor implements Visitor {
   private _possibleElements = new Map<string, ScannedElement>();
   private _registeredButNotFound = new Map<string, TagNameExpression>();
@@ -242,8 +232,9 @@ class ElementVisitor implements Visitor {
 
     const className = node.id && node.id.name;
     const element = new ScannedPolymerElement({
+      className,
       astNode: node,
-      tagName: isValue, className,
+      tagName: isValue,
       description: (docs.description || '').trim(),
       events: esutil.getEventComments(node),
       sourceRange: this._document.sourceRangeForNode(node),
@@ -251,11 +242,12 @@ class ElementVisitor implements Visitor {
       methods: getMethods(node, this._document),
       superClass: this._getExtends(node, docs, warnings),
       mixins: jsdoc.getMixins(this._document, node, docs, warnings),
-      privacy: getOrInferPrivacy(className || '', docs, false), observers
+      privacy: getOrInferPrivacy(className || '', docs, false),  //
+      observers,
     });
 
-    // If a class defines observedAttributes, it overrides what the base classes
-    // defined.
+    // If a class defines observedAttributes, it overrides what the base
+    // classes defined.
     // TODO(justinfagnani): define and handle composition patterns.
     const observedAttributes = this._getObservedAttributes(node);
 
@@ -278,21 +270,32 @@ class ElementVisitor implements Visitor {
       return;
     }
 
-    const tagName =
+    const tagNameExpressionResult =
         node.arguments[0] && this.getTagNameExpression(node.arguments[0]);
-    if (tagName == null) {
+    if (tagNameExpressionResult.kind === 'failure') {
+      this._warnings.push(tagNameExpressionResult.warning);
+      return;
+    }
+    const tagNameExpression = tagNameExpressionResult.value;
+    if (tagNameExpression == null) {
       return;
     }
     const elementDefn = node.arguments[1];
     if (elementDefn == null) {
       return;
     }
-    const element: ScannedElement|null = this._getElement(tagName, elementDefn);
+    const element: ScannedElement|null =
+        this._getElement(tagNameExpression, elementDefn);
     if (!element) {
       return;
     }
     this._elements.add(element);
-    element.tagName = this.getTagNameFromExpression(tagName);
+    const tagNameResult = this.getTagNameFromExpression(tagNameExpression);
+    if (tagNameResult.kind === 'success') {
+      element.tagName = tagNameResult.value;
+    } else {
+      this._warnings.push(tagNameResult.warning);
+    }
   }
 
   private _getElement(tagName: TagNameExpression, elementDefn: estree.Node):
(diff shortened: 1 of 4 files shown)


justinfagnani:
I think we agreed that this shouldn't be necessary, the presence of the `is` getter should be sufficient to give an element a tag name.

rictic:
Yeah but without noticing the `customElements.define` call then every custom element has to be annotated with an `@customElement` or similar, which is kinda eh.

We do need something like `@customElement` but I think for most cases it should not be necessary.

src/polymer/polymer2-element-scanner.ts
```diff
@@ -43,11 +43,23 @@ export class Polymer2ElementScanner implements JavaScriptScanner {
   }
 }
 
+type TagNameExpression = {
```
rictic: Sure, done. Also gave the two options names.

src/polymer/polymer2-element-scanner.ts
```diff
@@ -231,8 +243,7 @@ class ElementVisitor implements Visitor {
     const className = node.id && node.id.name;
     const element = new ScannedPolymerElement({
       astNode: node,
-      tagName: isValue,
-      className,
+      tagName: isValue, className,
```
rictic: Cleaned up a bit

src/polymer/polymer2-element-scanner.ts
```diff
@@ -279,12 +291,12 @@ class ElementVisitor implements Visitor {
     if (!element) {
       return;
     }
-    element.tagName = tagName;
     this._elements.add(element);
+    element.tagName = this.getTagNameFromExpression(tagName);
```
rictic: +1

src/polymer/polymer2-element-scanner.ts
```diff
@@ -405,10 +417,65 @@ class ElementVisitor implements Visitor {
...
+      if (isPropertyNameIs && className) {
+        return {
+          type: 'is',
+          className,
+          classNameSourceRange:
+              this._document.sourceRangeForNode(expression.object)!
+        };
+      }
+    }
+    this._warnings.push({
```
rictic: +1, thanks for pushing for more functional and less side-effecty.

Introduced a new local type `OrWarning<T>` for this and getTagNameFromExpression

src/polymer/polymer2-element-scanner.ts
```diff
@@ -230,21 +232,22 @@ class ElementVisitor implements Visitor {
...
-      className,
       description: (docs.description || '').trim(),
       events: esutil.getEventComments(node),
       sourceRange: this._document.sourceRangeForNode(node),
       properties: getProperties(node, this._document),
       methods: getMethods(node, this._document),
       superClass: this._getExtends(node, docs, warnings),
       mixins: jsdoc.getMixins(this._document, node, docs, warnings),
-      privacy: getOrInferPrivacy(className || '', docs, false), observers
+      privacy: getOrInferPrivacy(className || '', docs, false),  //
```
FredKSchott: A wild `//` appears

src/polymer/polymer2-element-scanner.ts
```diff
@@ -462,20 +475,48 @@ class ElementVisitor implements Visitor {
...
+        message:
+            `Unable to evaluate this expression down to a definitive string tagname.`,
+        severity: Severity.WARNING,
+        sourceRange: this._document.sourceRangeForNode(expression)!
+      }
+    };
   }
 }
+
+type OrWarning<V> = {
```
FredKSchott: I really like this name but it is a little too clever ("This function returns an `OrWarning`... what the hells an `OrWarning`???")

How about `ValueOrFailure`? `PossibleValue`? `ValueAndTheTerribleHorribleNoGoodVeryBadWarning`? +1 for anything that can stand on its own as a type name.

src/polymer/polymer2-element-scanner.ts
```diff
@@ -462,20 +475,48 @@ class ElementVisitor implements Visitor {
...
+        code: 'cant-determine-element-tagname',
+        message:
+            `Unable to evaluate this expression down to a definitive string tagname.`,
+        severity: Severity.WARNING,
+        sourceRange: this._document.sourceRangeForNode(expression)!
+      }
+    };
   }
 }
+
```
FredKSchott: Why the type definitions at the bottom of the file? I thought we always defined at the top of the file unless they were used in a single specific place.

rictic:
I hashed this out with Justin.

The big issue is that we want to be able to distinguish between vanilla elements and polymer elements. Currently we conflate them. The best solution is to look at the inheritance graph. If a class is downstream of HTMLElement, it's a CustomElement. If a class is downstream of Polymer.BaseElement, then it's also a Polymer element. Other CE libraries that use inheritance can plugin the same way, if we design the data schema correctly (i.e. all Polymer-specific stuff goes on a metadata field on a generic CustomElement object). 

Until then, we can also use `@polymerElement` and `@customElement` to distinguish here. Filed https://github.com/Polymer/tools-team/issues/9 to track.

src/polymer/polymer2-element-scanner.ts
rictic: Well it'd never be an OrWarning, it'd be a `OrWarning<SpecificKindOfValue>` but it is a bit backwards reading in english.

Haskell would call this type `Either<V, Warning>`, Rust calls it `Result<V, Warning>`. I guess having `Warning` in the type name isn't any shorter really than having it as a type parameter. How about `Result`?

I also changed the name of the property on `failure` from `warning` to `value`. This will make v8 happier, as there's only one shape to our return wrapper.

src/polymer/polymer2-element-scanner.ts
rictic: It's preventing `observers` from being put on the same line as `privacy`

src/polymer/polymer2-element-scanner.ts
rictic: It felt front loaded, but I don't have a strong opinion. Moved to the top.

src/polymer/polymer2-element-scanner.ts
rictic: Hm, if we end up using this a lot we could get further savings by changing from string `success` and `failure` to a boolean `successful` value. It reads better too. Changed.

src/polymer/polymer2-element-scanner.ts
```diff
@@ -21,13 +21,41 @@ import * as esutil from '../javascript/esutil';
...
+import {ScannedElement, ScannedFeature, ScannedReference, Severity, SourceRange, Warning} from '../model/model';
 
 import {extractObservers} from './declaration-property-handlers';
 import {getOrInferPrivacy} from './js-utils';
 import {Observer, ScannedPolymerElement} from './polymer-element';
 import {getIsValue, getMethods, getProperties} from './polymer2-config';
 
+
+/** Represents the value of an operation that may fail. */
+type Result<V, E> = {
```
FredKSchott: +1, this is perfect

src/polymer/polymer2-element-scanner.ts
FredKSchott: Bah, at this point I would probably just surrender this and let clang do its thing. Someone will eventually find this and remove it, not knowing it's true purpose. Your call.

rictic: Eh, if future-us remove it and like the results, more power to them :)

rictic:
Thanks for the review!

PR #541, opened 2017-03-04
IncompleteFeature: unattached scan-time warnings

rictic:

Will use in a later PR.

### DIFF
--- a/src/model/warning.ts
+++ b/src/model/warning.ts
@@ -12,7 +12,11 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
+import {Document} from './document';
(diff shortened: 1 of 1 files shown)


src/model/warning.ts
```diff
@@ -34,3 +38,35 @@ export class WarningCarryingException extends Error {
     this.warning = warning;
   }
 }
+
+const emptySet = new Set();
```
FredKSchott: bah, sharing a single set like this makes my skin crawl. Wish there was a way to freeze a set so that we could guarantee it was empty. oh well.

justinfagnani:
Superceded by #642

PR #543, opened 2017-03-04
Rework how we handle laziness.

rictic:
TODO: test strict querying.

/cc @usergenic as this affects bundling.


rictic:
Part of handling https://github.com/Polymer/tools-sample-projects/issues/3

src/model/queryable.ts
```diff
@@ -25,6 +25,12 @@ export interface BaseQueryOptions {
    * that are referenced from within the package.
    */
   externalPackages?: boolean;
+
+  /**
+   * Do not include any features that are only reachable via paths that include
+   * lazy import edges.
+   */
+  strictImports?: boolean;
```
justinfagnani: can we rename this to `lazyImports` with default true?

src/core/analysis-context.ts
```diff
@@ -279,14 +279,8 @@ export class AnalysisContext {
...
 
-            // Find all non-lazy imports
-            // TODO(justinfagnani): I think we should scan lazily imported
-            // documents since we know about them, we should load them. Their
-            // features should possibly be separated out at export time via
-            // better definition of scopes
             const imports = scannedDocument.getNestedFeatures().filter(
-                (e) => e instanceof ScannedImport &&
-                    e.type !== 'lazy-html-import') as ScannedImport[];
+                (e) => e instanceof ScannedImport) as ScannedImport[];
```
justinfagnani: I thought there were more places that needed fixup...

rictic: Good catch, got the other place in `analysis-context`

src/model/queryable.ts
rictic: I'd been going with the idea that everything was default false, to match the semantics of `options.foo` without explicit defaults handling stuff.

src/html/html-import-scanner.ts
```diff
@@ -56,29 +55,31 @@ export class HtmlImportScanner implements HtmlScanner {
...
       if (isHtmlImportNode(node)) {
-        type = 'html-import';
+        lazy = false;
       } else if (isLazyImportNode(node)) {
-        type = 'lazy-html-import';
+        lazy = true;
       } else {
         return;
       }
+      const type = 'html-import';
```
FredKSchott: no longer need type variable here, only used once below

src/model/document.ts
```diff
@@ -238,14 +238,15 @@ export class Document implements Feature, Queryable {
   getByKind(kind: string, options?: QueryOptions): Set<Feature>;
   getByKind(kind: string, options?: QueryOptions): Set<Feature> {
     options = options || {};
-    if (this._featuresByKind && options.imported) {
+    if (this._featuresByKind && options.imported && !options.strictImports) {
```
FredKSchott: `!options.strictImports` is used in both checks below, can you pull it out of both and handle it separately?

MaKleSoft:
This is approved so why has it not been merged yet? Was about to submit an almost identical PR before I saw this one...

rictic:
I wanted to get feedback from Justin, but he's going on vacation, lemmy ping him, get his thoughts

src/model/document.ts
rictic: +1, this was a good clarity change

src/model/queryable.ts
justinfagnani: that's certainly a bit simpler, but the name here doesn't have a clear meaning to me. How about a convention of a "no" prefix for options we'd rather have a default of true. So `noLazyImports`?

rictic: +1

PR #544, opened 2017-03-05
Expose only a single feature query method.

rictic:

### DIFF
--- a/src/model/document.ts
+++ b/src/model/document.ts
@@ -19,7 +19,7 @@ import {ParsedDocument} from '../parser/document';
 import {Analysis} from './analysis';
 import {Feature, ScannedFeature} from './feature';
 import {Import} from './import';
-import {DocumentQuery as Query, DocumentQueryOptions as QueryOptions, DocumentQueryWithKind as QueryWithKind, FeatureKinds, Queryable} from './queryable';
+import {DocumentQuery as Query, DocumentQueryOptions as QueryOptions, DocumentQueryWithKind as QueryWithKind, FeatureKind, FeatureKindMap, Queryable} from './queryable';
 import {isResolvable} from './resolvable';
 import {SourceRange} from './source-range';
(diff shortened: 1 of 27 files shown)


src/model/document.ts
```diff
@@ -238,10 +191,45 @@ export class Document implements Feature, Queryable {
...
+   *
+   * You can also narrow by feature kind and identifier.
+   */
+  getFeatures<K extends keyof FeatureKinds>(query: QueryWithKind<K>):
       Set<FeatureKinds[K]>;
-  getByKind(kind: string, options?: QueryOptions): Set<Feature>;
-  getByKind(kind: string, options?: QueryOptions): Set<Feature> {
+  getFeatures(query: Query): Set<Feature>;
+  getFeatures(query: Query): Set<Feature> {
+    const queryOptions = {
```
justinfagnani: it seems like you're splitting the query up into separate objects, then mostly passing them together. Why not make `_getById`, `_getByKind` and `listFeatures` take a `Query` object?

src/model/document.ts
```diff
@@ -332,7 +297,7 @@ export class Document implements Feature, Queryable {
     return result;
   }
 
-  private _getFeatures(
+  private _listFeatures(
```
justinfagnani: docs

src/model/document.ts
```diff
@@ -371,9 +336,9 @@ export class Document implements Feature, Queryable {
   /**
    * Get warnings for the document and all matched features.
    */
-  getWarnings(options?: QueryOptions): Warning[] {
+  getWarnings(query?: Query): Warning[] {
     const warnings: Set<Warning> = new Set(this.warnings);
-    for (const feature of this.getFeatures(options)) {
+    for (const feature of this.getFeatures(query || {})) {
```
justinfagnani: use a default parameter value instead

PR #545, opened 2017-03-05
Remove second argument to analyze(), add filesChanged and InMemoryOverlayLoader

rictic:

### DIFF
--- a/src/analyzer.ts
+++ b/src/analyzer.ts
@@ -65,25 +65,26 @@ export class Analyzer {
   /**
    * Loads, parses and analyzes the root document of a dependency graph and its
    * transitive dependencies.
-   *
-   * Note: The analyzer only supports analyzing a single root for now. This
-   * is because each analyzed document in the dependency graph has a single
-   * root. This mean that we can't properly analyze app-shell-style, lazy
-   * loading apps.
-   *
-   * @param contents Optional contents of the file when it is known without
-   * reading it from disk. Clears the caches so that the news contents is used
-   * and reanalyzed. Useful for editors that want to re-analyze changed files.
    */
-  async analyze(url: string, contents?: string): Promise<Document> {
-    if (contents != null) {
-      this._context = this._context.filesChanged([url]);
-    }
-    return this._context.analyze(url, contents);
+  async analyze(url: string): Promise<Document> {
+    return this._context.analyze(url);
   }
 
   async analyzePackage(): Promise<Package> {
     return this._context.analyzePackage();
+  }
+
+  /**
+   * Clears all information about the given files from our caches, such that
+   * future calls to analyze() will reload these files if they're needed.
+   *
+   * The analyzer assumes that if this method isn't called with a file's url,
+   * then that file has not changed and does not need to be reloaded.
+   *
+   * @param urls The urls of files which may have changed.
+   */
+  filesChanged(urls: string[]) {
+    this._context = this._context.filesChanged(urls);
   }
 
   /**
@@ -106,23 +107,20 @@ export class Analyzer {
    * see a mixture of pre-fork and post-fork contents when you analyze with a
    * forked analyzer.
    *
-   * Note: this feature is experimental.
+   * Note: this feature is experimental. It may be removed without being
+   *     considered a breaking change, so check for its existence before calling
+   *     it.
    */
   _fork(options?: ForkOptions): Analyzer {
-    const context =
-        options ? this._context._fork(undefined, options) : this._context;
+    const context = this._context._fork(undefined, options);
     return new Analyzer(context);
   }
 
   /**
    * Loads the content at the provided resolved URL.
-   *
-   * Currently does no caching. If the provided contents are given then they
-   * are used instead of hitting the UrlLoader (e.g. when you have in-memory
-   * contents that should override disk).
    */
-  async load(resolvedUrl: string, providedContents?: string) {
-    return this._context.load(resolvedUrl, providedContents);
+  async load(resolvedUrl: string) {
+    return this._context.load(resolvedUrl);
   }
 
   canResolveUrl(url: string): boolean {
--- a/src/test/typescript/typescript-analyzer_test.ts
+++ b/src/test/typescript/typescript-analyzer_test.ts
@@ -17,11 +17,14 @@ import * as ts from 'typescript';
 
 import {AnalysisContext} from '../../core/analysis-context';
 import {TypeScriptAnalyzer} from '../../typescript/typescript-analyzer';
+import {InMemoryOverlayLoader} from '../../url-loader/in-memory-overlay-loader';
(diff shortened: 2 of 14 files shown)


src/url-loader/in-memory-overlay-loader.ts
```diff
@@ -0,0 +1,65 @@
...
+ * buffers at any time. For these files, the editor's in-memory buffer is
+ * canonical, so that their contents are read even when they have unsaved
+ * changes. For all other files, we can load the files using another loader,
+ * e.g. from disk.
+ */
+export class InMemoryOverlayLoader implements UrlLoader {
+  private readonly _fallbackLoader: UrlLoader;
+  private readonly _memoryMap = new Map<string, string>();
+
+  constructor(fallbackLoader: UrlLoader) {
```
justinfagnani: can you make fallbackLoader optional and replace TestUrlLoader in test-utils.js

usergenic:
I use `Analyzer#analyzer(url, newFileContents)` from inside the Bundler a bunch, so I definitely want `filesChanged()` ... but now I am going to have to gymnastics up some reforking the Analyzer to give it an InMemoryUrlLoader(whateverotherloader) somehow `O_o` ...

usergenic:
That was gripey of me.  I can totally find a way.

src/analyzer.ts
```diff
@@ -106,23 +107,20 @@ export class Analyzer {
...
+   * Note: this feature is experimental. It may be removed without being
+   *     considered a breaking change, so check for its existence before calling
+   *     it.
    */
   _fork(options?: ForkOptions): Analyzer {
-    const context =
-        options ? this._context._fork(undefined, options) : this._context;
+    const context = this._context._fork(undefined, options);
     return new Analyzer(context);
   }
```
rictic: @usergenic What do you think of this change? It seems like it's worth always calling `this._context._fork` because otherwise the contexts *are* exactly the same until the next call to `filesChanged`

usergenic: Thought the optional context `_fork` would save an unnecessary cache duplication since it would not be necessary to do until something changed?  It does mean we have to be super sure all non-idempotent analyzer methods cause a context fork, which I think that's how it is intended it to work?  I'll defer to your feelings on this though, since you have a much crisper vision of analyzer's design choices.

rictic: In my head, the semantics of `_fork` are that you want the cache from the existing analyzer, but you want to break with it, so that you won't see any new results in it and it won't see any new results from you.

Maybe there's no use case for calling it no options though. I mean, unless you're going to change the UrlLoader or something why not just use the original analyzer?

usergenic: My attempts to contrive a use case for no-options have not resulted in anything too compelling to preserve a minor optimization that is likely to only result in an eventual bug.

I hereby declare your change desirable!

usergenic: So here's my usage of `_fork` in an upcoming PR:

```
      analyzer:
          this._buildAnalyzer.analyzer
              ._fork({
                urlLoader: new FileMapUrlLoader(this.config.root, this.files)
              })
              .filesChanged(filesChanged),
```

Which is clearly going to result in a redundant analysis cache duplication, since I'm filesChanged()-ing it right after the fork.

It's minor, but I guess that's sort of a meh for me.  Do you think analyzer._fork should support a `filesChanged` option as well like `_fork({urlLoader: ..., filesChanged: [...]})`?

rictic: Compared to the other work of analyzing, `filesChanged` is very fast. It copies a few maps, then deletes a few entries. I'd expect tens of microseconds, not worth worrying about.

src/url-loader/in-memory-overlay-loader.ts
```diff
@@ -0,0 +1,74 @@
...
+ * Resolves requests first from an in-memory map of file contents, and if a
+ * file isn't found there, defers to another url loader.
+ *
+ * Useful for the editor use case. An editor will have a number of files in open
+ * buffers at any time. For these files, the editor's in-memory buffer is
+ * canonical, so that their contents are read even when they have unsaved
+ * changes. For all other files, we can load the files using another loader,
+ * e.g. from disk.
+ */
+export class InMemoryOverlayLoader implements UrlLoader {
```
justinfagnani: I like keeping `UrlLoader` in the name. How about just `OverlayUrlLoader`?

src/url-loader/in-memory-overlay-loader.ts
```diff
@@ -0,0 +1,74 @@
...
+    if (typeof contents === 'string') {
+      return contents;
+    }
+    return this._fallbackLoader.load(url);
+  }
+
+  // We have this method if our underlying loader has it.
+  readDirectory?: (pathFromRoot: string, deep?: boolean) => Promise<string[]>;
+
+  mapFile(url: string, contents: string) {
```
justinfagnani: `mapFile` reads weird to me, why not just expose the map directly so that you get the full API, including `has` and `forEach`?

src/url-loader/in-memory-overlay-loader.ts
```diff
@@ -0,0 +1,74 @@
...
+  constructor(fallbackLoader?: UrlLoader) {
+    this._fallbackLoader = fallbackLoader || new FailUrlLoader();
+    if (this._fallbackLoader.readDirectory) {
+      this.readDirectory =
+          this._fallbackLoader.readDirectory.bind(this._fallbackLoader);
+    }
+  }
+
+  canLoad(url: string): boolean {
+    if (this._memoryMap.has(url)) {
```
justinfagnani: golf: `return this._memoryMap.has(url) || this._fallbackLoader.canLoad(url);`

src/test/typescript/typescript-analyzer_test.ts
```diff
@@ -17,11 +17,14 @@ import * as ts from 'typescript';
 
 import {AnalysisContext} from '../../core/analysis-context';
 import {TypeScriptAnalyzer} from '../../typescript/typescript-analyzer';
+import {InMemoryOverlayLoader} from '../../url-loader/in-memory-overlay-loader';
 import {PackageUrlResolver} from '../../url-loader/package-url-resolver';
-import {TestUrlLoader} from '../test-utils';
 
 async function getTypeScriptAnalyzer(files: {[url: string]: string}) {
-  const urlLoader = new TestUrlLoader(files);
+  const urlLoader = new InMemoryOverlayLoader();
```
justinfagnani: maybe the constructor should take a map to initialize it?

src/url-loader/in-memory-overlay-loader.ts
rictic: Went with InMemoryOverlayUrlLoader because this is a power tool that will be used very rarely, and for rarely used power tools super long descriptive names are appropriate.

PR #546, opened 2017-03-06
Parse databindings in JS string literals.

rictic:
For stuff like the observers array, computed properties, etc.

Not hooked up anywhere yet, will do that in a followup PR.

### DIFF
--- a/src/polymer/expression-scanner.ts
+++ b/src/polymer/expression-scanner.ts
@@ -17,9 +17,11 @@ import * as estree from 'estree';
 import * as parse5 from 'parse5';
 
 import {ParsedHtmlDocument} from '../html/html-document';
+import {JavaScriptDocument} from '../javascript/javascript-document';
 import {parseJs} from '../javascript/javascript-parser';
 import {correctSourceRange, LocationOffset, SourceRange} from '../model/model';
 import {Severity, Warning} from '../warning/warning';
+
 
 const p = dom5.predicates;
 const isTemplate = p.hasTagName('template');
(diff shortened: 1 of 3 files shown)


src/test/polymer/expression-scanner_test.ts
```diff
@@ -12,16 +12,17 @@
...
 import {HtmlParser} from '../../html/html-parser';
-import {scanDocumentForExpressions} from '../../polymer/expression-scanner';
+import {JavaScriptParser} from '../../javascript/javascript-parser';
+import {AttributeDatabindingExpression, parseExpressionInJsStringLiteral, scanDocumentForExpressions, TextNodeDatabindingExpression} from '../../polymer/expression-scanner';
 import {CodeUnderliner} from '../test-utils';
 
 suite('ExpressionScanner', () => {
 
-  suite('scan()', () => {
+  suite('scanning html for expressions', () => {
```
usergenic: Is there a test that demonstrates the expression is not scanned/returned unless its in a template?

usergenic: Or is that responsibility in a different scope?

rictic: I added a few more negative examples to the scanned source, WDYT?

usergenic: Cool

src/polymer/expression-scanner.ts
```diff
@@ -201,6 +170,85 @@ export class DatabindingExpression {
...
+   */
+  readonly eventName: string|undefined;
+
+  /** The attribute we're databinding into. */
+  readonly attribute: parse5.ASTAttribute;
+
+  constructor(
+      astNode: parse5.ASTNode, isCompleteBinding: boolean, direction: '{'|'[',
+      eventName: string|undefined, attribute: parse5.ASTAttribute,
+      sourceRange: SourceRange, expressionText: string, ast: estree.Program) {
```
usergenic: That's some serious arity!  Not recommending a change per se, especially since this constructor is not really used by the outside world.  Definitely would be too much without type-checking! :)

rictic: Yeah, maybe an options object is called for? We do actually want to track all of these properties, but it is a p. serious bag of values yeah.

Issue #547, opened 2017-03-06
Extend the `Analyzer#_fork()` API to support replacing the `urlLoader` option (possibly replace any of the initialization options?)


usergenic:
The primary impetus for adding `_fork` method to Analyzer was to support use of a different `UrlLoader` downstream during `polymer-build` pipeline so the bundle operation would read from the stream at that point instead of from the original analysis documents (to support bundling of minified or other transpiled script/sass sources etc.).

Unfortunately, the private `urlLoader` property is on a private AnalysisContext within the Analyzer, so there's no means to replace it after the fork.  It makes sense to replace during the fork operation itself though, so if `_fork()` could take similar/identical arguments similar to the way Analyzer constructor does, it would be a predictable interface and solve this.

rictic:
It might be valuable to be able to pass in new initialization options to `_fork`. e.g. someone might want to add new scanners. The Analyzer can determine what caches need to be emptied based on the changed options. Related, but it would also be useful in these scenarios to be able to read the configuration off of an existing Analyzer.

e.g. a lint rule could necessitate additional scanners. By forking rather than creating a new analyzer we could reuse the other configuration options from the analyzer passed into the linter, as well as reusing the pre-scan caches (i.e. the loading and parsing caches).

usergenic:
Done.

PR #551, opened 2017-03-06
Added options to the `_fork` method and also made `_fork` public and experimental on AnalysisContext.

usergenic:
- Addresses #547 supporting line-item property forking by letting `_fork()` take options.

### DIFF
--- a/src/core/analysis-context.ts
+++ b/src/core/analysis-context.ts
@@ -260,8 +260,8 @@ export class AnalysisContext {
       urlLoader: this._loader,
       urlResolver: this._resolver,
     };
-    if (options) {
-      Object.assign(contextOptions, options);
+    if (options && options.urlLoader) {
+      contextOptions.urlLoader = options.urlLoader;
     }
     const copy = new AnalysisContext(contextOptions);
     if (!cache) {
--- a/src/core/analysis-context.ts
+++ b/src/core/analysis-context.ts
@@ -14,7 +14,7 @@
 
 import * as path from 'path';
 
-import {LazyEdgeMap, NoKnownParserError, Options, ScannerTable} from '../analyzer';
+import {ForkOptions, LazyEdgeMap, NoKnownParserError, Options, ScannerTable} from '../analyzer';
 import {CssParser} from '../css/css-parser';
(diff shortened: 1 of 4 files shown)


src/core/analysis-context.ts
```diff
@@ -247,16 +247,26 @@ export class AnalysisContext {
...
+      urlResolver: this._resolver,
+    };
+    if (options) {
+      Object.assign(contextOptions, options);
+    }
+    const copy = new AnalysisContext(contextOptions);
+    if (!cache) {
+      cache = this._cache.invalidate([]);
+    }
     copy._cache = cache;
```
rictic: If `options` has any keys besides urlLoader, I think we shouldn't copy over the cache. e.g. if the user adds a new scanner, then we'll need to at least re-scan, resolve, etc all documents. Copying over the right parts of the cache is an optimization we can get to later as needed.

I suppose you could make this argument for urlLoader too. Hm. Actually, for the bundler case, you do want to reload files from the incoming stream, don't you. You don't want to load from the cache from the previous urlLoader, do you?

I suppose you'll need to check which files have changed and call filesChanged() (will be made public in https://github.com/Polymer/polymer-analyzer/pull/545).

If this makes sense and is still a useful use case, it'll need some docs to let the user know of these constraints I think. I guess _fork is an experimental method, but we should at least have a marker in the docs indicating that the semantics of forking with options are hairy and we'll need to carefully vet them before exposing the _fork method.

usergenic: Perhaps the correct thing to do now is just to limit the override-able settings to the `urlLoader` since at least in this case, the analyzer's behavior is predictable.  Dealing with the other cases seems like they need some special handling and explanation that is beyond what should be tested for here.  I've updated the types/methods/tests accordingly.

usergenic: @rictic PTAL

rictic: That's fair. Please also update the code for _fork to not use Object.assign but to only override the one known property.

Also please document on ForkOptions.urlLoader that forking does not clear caches, so you will see a mixture of pre-fork and post-fork contents when you analyze with a forked analyzer.

usergenic: Method comment updated, @rictic PTAL

PR #552, opened 2017-03-07
Parse observers and computed functions in Polymer declarations

rictic:


src/polymer/expression-scanner.ts
```diff
@@ -59,6 +59,11 @@ export function getAllDataBindingTemplates(node: parse5.ASTNode) {
 
 export type HtmlDatabindingExpression =
     TextNodeDatabindingExpression | AttributeDatabindingExpression;
+
+// Some expressions are limited. For example, in a property declaration,
```
usergenic: Are `//` style comments ignored by documentation tools?  Just wondering if that's why this form was chosen.

src/polymer/expression-scanner.ts
```diff
@@ -120,6 +126,17 @@ export abstract class DatabindingExpression {
...
     }
     let expression = expressionStatement.expression;
+    if (limitation === 'identifierOnly') {
+      if (expression.type !== 'Identifier') {
+        this.warnings.push(this._validationWarning(
+            `Expected just a name here, not an expression`, expression));
+      }
+    } else if (limitation === 'callExpression') {
+      if (expression.type !== 'CallExpression') {
+        this.warnings.push(this._validationWarning(
```
usergenic: Minor nit; maybe this could be less nesty?  Unless theres an intention to extend the cases within the limitation categories later...?
```
if (limitation === 'identifierOnly' && expression.type !== 'Identifier) {
  ...
} else if (limitation === 'callExpression' && expression.type !== 'CallExpression') {
  ...
}
```

src/polymer/polymer2-element-scanner.ts
```diff
@@ -303,20 +311,37 @@ class ElementVisitor implements Visitor {
...
+  private _getObservers(node: estree.ClassDeclaration|estree.ClassExpression) {
+    const returnedValue = this._getReturnValueOfStaticGetter(node, 'observers');
+    if (returnedValue) {
+      return extractObservers(returnedValue, this._document);
+    }
+  }
+
+  private _getReturnValueOfStaticGetter(
+      node: estree.ClassDeclaration|estree.ClassExpression,
+      methodName: string) {
```
usergenic: These getters don't have explicit return types in the function declarations.  Is that a style choice / because these are private and typescript is smart enough?

src/test/polymer/polymer2-element-scanner_test.ts
```diff
@@ -25,7 +25,7 @@ import {CodeUnderliner} from '../test-utils';
 
 chaiUse(require('chai-subset'));
 
-suite('Polymer2ElementScanner', () => {
+suite.only('Polymer2ElementScanner', () => {
```
usergenic: I think this `only` might have been left in by mistake.

src/polymer/expression-scanner.ts
rictic: Looks like they are, but I didn't have a good reason for doing so.

Switched to `/** docs */` style

src/polymer/expression-scanner.ts
rictic: Moved to a method and a switch/case implementation that ensures all limitation values are checked or it's a compile time error.

src/polymer/polymer2-element-scanner.ts
rictic: Yeah. For private methods it seems kosher to me to let type inference do the work. If it hurts legibility then we should add the type declarations though.

Added here.

src/test/polymer/polymer2-element-scanner_test.ts
rictic: erp! removed.

We should have a lint pass to catch that…

PR #553, opened 2017-03-07
Produce clean method descriptions

rictic:
- [x] CHANGELOG.md not updated, minor bugfix.

### DIFF
--- a/src/javascript/jsdoc.ts
+++ b/src/javascript/jsdoc.ts
@@ -156,7 +156,8 @@ export function parseJsdoc(docs: string): Annotation {
       docs, {unwrap: false, lineNumbers: true, preserveWhitespace: true});
   // Strip any leading and trailing newline characters in the
   // description of multiline comments for readibility.
-  const description = d.description && d.description.replace(/^\n+|\n+$/g, '');
+  const description =
+      d.description && d.description.replace(/^\n+|\n+$/g, '').trim();
   return {description: description, tags: _tagsToHydroTags(d.tags)};
 }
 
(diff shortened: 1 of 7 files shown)


src/javascript/jsdoc.ts
```diff
@@ -156,7 +156,8 @@ export function parseJsdoc(docs: string): Annotation {
       docs, {unwrap: false, lineNumbers: true, preserveWhitespace: true});
   // Strip any leading and trailing newline characters in the
   // description of multiline comments for readibility.
-  const description = d.description && d.description.replace(/^\n+|\n+$/g, '');
+  const description =
+      d.description && d.description.replace(/^\n+|\n+$/g, '').trim();
```
justinfagnani: trim was supposedly left out on purpose to preserve indentation on the first line

src/test/javascript/jsdoc_test.ts
```diff
@@ -76,7 +76,7 @@ suite('jsdoc', function() {
 
     test('preserves indentation for the body', function() {
       const parsed = jsdoc.parseJsdoc('*     The desc.\n*     thing');
-      assert.deepEqual(parsed.description, '    The desc.\n    thing');
+      assert.deepEqual(parsed.description, 'The desc.\n    thing');
```
justinfagnani: this was testing for that leading indentation

rictic: Yeah, I couldn't think of a reason to keep it, but you're right, we don't want to be tracking down weird errors right now. Chesterton's Fence beats Cargo Cult for now. Added a todo to look into it.

Reverted.

PR #556, opened 2017-03-10
Fix a subtle issue in getIdentifierName

rictic:
It was treating `foo[bar]` the same as `foo.bar` and `foo['bar']`.

 - [x] Minor bugfix. Not CHANGELOG.md worthy


src/test/static/namespaces/namespace-dynamic-unnamed.js
```diff
@@ -7,6 +7,7 @@ DynamicNamespace['ArrayNotation'] = {
 
 var baz = 'abc';
 /**
+ * We can't look this up yet.
```
usergenic: Some day!

rictic:
Thanks for the review!

PR #559, opened 2017-03-13
Use latest version of estraverse.

rictic:
The only breaking change was a minor one in the order that nodes were walked.


rictic:
Gonna roll this in with the prepare alpha.32 branch

PR #560, opened 2017-03-13
Prepare to release alpha.32

rictic:


rictic:
The estraverse breaking changes are small, and relate to the order of tree walking. They shouldn't affect any of our code.

PR #568, opened 2017-03-17
Fix filtering out external features.

rictic:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 
 * All Documents, ParsedDocuments, and ScannedDocuments now have correct SourceRanges. This also fixes a bug whereby `getByKind('document')` would fail to filter out package-external Documents.
+* Fix an issue where package-external features could be included in queries that dot not ask for them. Specifically we were filtering based on the text in files  like `../paper-button/paper-button.html` rather than the path to the file on disk like `bower_components/paper-button/paper-button.html` when determining externality.
 
 ## [2.0.0-alpha.33] - 2017-03-14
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -11,7 +11,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
    Unreleased section, uncommenting the header as necessary.
 -->
 
-<!--## Unreleased-->
+## Unreleased
+
+* Fix an issue where package-external features could be included in queries that explicitly asked for them not to be. Specifically imports like `../paper-button/paper-button.html` weren't considering the resolved path like `bower_components/paper-button/paper-button.html` when determining externality.
```
FredKSchott: not major, but a little confused by what you mean that "imports wren't considering their own resolved path". Shouldn't this be that the analyzer isn't considering resolvedness?

src/model/document.ts
```diff
@@ -335,7 +335,8 @@ export class Document implements Feature, Queryable {
       }
       if (feature.kinds.has('import') && options.imported) {
         const imprt = feature as Import;
-        if (options.externalPackages || !Package.isExternal(imprt.url)) {
+        if (options.externalPackages ||
```
FredKSchott: can you add parens here? They make it so much easier to read/maintain when there are multiple logical operators in a conditional.

FredKSchott: you could also pull `imprt.document && !Package.isExternal(imprt.document.url)` into a variable `isExternalDocument`

CHANGELOG.md
rictic: Yeah, this wasn't clear. Rephrased to talk about user-visible concepts like "path in file" and "path on disk"

PR #569, opened 2017-03-17
Ensure all documents have source ranges

rictic:
- [x] CHANGELOG.md has been updated

PR #570, opened 2017-03-17
Use a fast and simple method to convert between offsets and source ranges

rictic:
- [x] CHANGELOG.md not updated, no visible changes

Builds on top of #569 

This change deletes a bunch of code, but fortunately there's really good tests of the deleted code, so we can be confident that the replacement is equivalent. See the tests for html document and polymer expressions.

### DIFF
--- a/src/parser/document.ts
+++ b/src/parser/document.ts
@@ -166,7 +166,7 @@ export interface StringifyOptions {
  * The variant of binary search that returns the number of elements in the
  * array that is strictly less than the target.
  */
-export function binarySearch(target: number, arr: number[]) {
+function binarySearch(target: number, arr: number[]) {
   let lower = 0;
   let upper = arr.length - 1;
   while (true) {
--- a/src/parser/document.ts
+++ b/src/parser/document.ts
@@ -111,18 +111,18 @@ export abstract class ParsedDocument<AstNode, Visitor> {
   }
 
   sourcePositionToOffset(position: SourcePosition): number {
+    const line = Math.max(0, position.line);
     let lineOffset;
-    if (position.line === 0) {
+    if (line === 0) {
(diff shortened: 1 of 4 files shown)


src/parser/document.ts
```diff
@@ -121,3 +161,26 @@ export interface StringifyOptions {
    */
   inlineDocuments?: ParsedDocument<any, any>[];
 }
+
+/**
+ * The variant of binary search that returns the number of elements in the
+ * array that is strictly less than the target.
+ */
+export function binarySearch(target: number, arr: number[]) {
```
FredKSchott: If we're going to export this, can it go somewhere else? Seems strange to live in `document.ts`

src/test/parser/document_test.ts
```diff
@@ -0,0 +1,86 @@
...
+        }
+      }
+    }
+  });
+
+  test('sourceRangeToOffsets works for simple cases', async() => {
+    let document = new TestDocument('ab');
+    assert.deepEqual(document.offsetToSourcePosition(0), {line: 0, column: 0});
+    assert.deepEqual(document.offsetToSourcePosition(1), {line: 0, column: 1});
+    assert.deepEqual(document.offsetToSourcePosition(2), {line: 0, column: 2});
```
FredKSchott: what happens when you call `offsetToSourcePosition` with a value out of range? Would be good to add a test for that

src/parser/document.ts
rictic: Ah, I only had it exported while I was messing with it. Not proud of how long it took me to write this lol

src/test/parser/document_test.ts
rictic: +1, also tested for out of bounds sourcePositionToOffset

PR #572, opened 2017-03-17
Recover from undefined source ranges.

rictic:
I've seen this a couple times, working to track it down but it's way easier with the warning code (and we don't want to disrupt users too much in this sort of situation).

I'll be looking into this in more detail.

### DIFF
--- a/src/warning/warning-printer.ts
+++ b/src/warning/warning-printer.ts
@@ -52,6 +52,13 @@ export class WarningPrinter {
   async printWarning(warning: Warning) {
     const severity = this._severityToString(warning.severity);
     const range = warning.sourceRange;
+    if (!range) {
+      this._outStream.write(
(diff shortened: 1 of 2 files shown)


src/warning/warning-printer.ts
```diff
@@ -52,6 +52,13 @@ export class WarningPrinter {
   async printWarning(warning: Warning) {
     const severity = this._severityToString(warning.severity);
     const range = warning.sourceRange;
+    if (!range) {
+      this._outStream.write(
```
FredKSchott: I understand not wanting to bother the user too much, but I think both they and us would benefit from having the message here as well. It could include important additional info, and can always be remove after we've found the specific error you're chasing down atm.

rictic: +1, done

PR #575, opened 2017-03-19
Treat `build` dir as external too.

rictic:
Seems like a reasonable default.

Fixes https://github.com/Polymer/polymer-cli/issues/616

### DIFF
--- a/src/model/package.ts
+++ b/src/model/package.ts
@@ -21,7 +21,7 @@ export type QueryOptions = object & BaseQueryOptions;
 
 // A regexp that matches paths to external code.
 // TODO(rictic): Make this extensible (polymer.json?).
-const MATCHES_EXTERNAL = /(^|\/)(bower_components|node_modules)/;
+const MATCHES_EXTERNAL = /(^|\/)(bower_components|node_modules|build($|\/))/;
 
 /**
  * Represents a queryable interface over all documents in a package/project.
--- a/src/model/package.ts
+++ b/src/model/package.ts
@@ -21,6 +21,9 @@ export type QueryOptions = object & BaseQueryOptions;
 
 // A regexp that matches paths to external code.
 // TODO(rictic): Make this extensible (polymer.json?).
+// Note that we match directories named exactly `build`, but will match any
+// directory name prefixed by `bower_components` or `node_modules`, in order to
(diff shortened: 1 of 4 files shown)


src/model/package.ts
```diff
@@ -21,7 +21,7 @@ export type QueryOptions = object & BaseQueryOptions;
 
 // A regexp that matches paths to external code.
 // TODO(rictic): Make this extensible (polymer.json?).
```
usergenic: +1

src/model/package.ts
```diff
@@ -21,7 +21,7 @@ export type QueryOptions = object & BaseQueryOptions;
...
-const MATCHES_EXTERNAL = /(^|\/)(bower_components|node_modules)/;
+const MATCHES_EXTERNAL = /(^|\/)(bower_components|node_modules|build($|\/))/;
```
usergenic: I don't think this Regexp is right...  I think it should be `/(^|\/)(bower_components|node_modules|build)($|\/)/`

We don't *need* to add a test for it but it wouldn't hurt...

src/test/static/project-with-errors/build/output.html
```diff
@@ -0,0 +1 @@
+<link rel="import" href="built-nonexistant-import.html">
```
usergenic: I couldn't figure how this or the other html test fixture is related to the PR?

src/model/package.ts
rictic: This is deliberate, but it's subtle enough that it deserves a comment. PTAL

src/test/static/project-with-errors/build/output.html
rictic: There are existing tests that analyze the packages at test/static/project and test/static/project-with-errors and assert on the documents found within. By adding these files we're getting assertions that they're not part of those packages

usergenic: Nevermind I totes understand now.  I didn't get the naming `built-nonexistant-import.html` at first.

src/model/package.ts
usergenic: Oh!  Didn't expect that, having not worked with our variants stuff yet.

Thanks for adding the comment!

rictic:
Thanks!

PR #580, opened 2017-03-23
Analysis.mixins now includes PolymerBehavior serializations

usergenic:
 - I put your `PolymerBehaviors` in your `generateAnalysis()['mixins']`
 - You can identify them by their `metadata.isPolymerBehavior=true`

Also,

noidea

So PTAL!

### DIFF
--- a/src/generate-analysis.ts
+++ b/src/generate-analysis.ts
@@ -21,7 +21,7 @@ import {Function as ResolvedFunction} from './javascript/function';
 import {Namespace as ResolvedNamespace} from './javascript/namespace';
 import {Document} from './model/document';
 import {Feature} from './model/feature';
-import {Attribute as ResolvedAttribute, Element as ResolvedElement, ElementMixin as ResolvedMixin, Event as ResolvedEvent, Method as ResolvedMethod, Property as ResolvedProperty, SourceRange as ResolvedSourceRange} from './model/model';
+import {Attribute as ResolvedAttribute, Element as ResolvedElement, ElementMixin as ResolvedMixin, Event as ResolvedEvent, Method as ResolvedMethod, PolymerBehavior as ResolvedPolymerBehavior, Property as ResolvedProperty, SourceRange as ResolvedSourceRange} from './model/model';
 import {Package} from './model/package';
 
 export type ElementOrMixin = ResolvedElement | ResolvedMixin;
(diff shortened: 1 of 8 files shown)


usergenic:
Note, this fixes https://github.com/Polymer/polymer-analyzer/issues/577

rictic:
Note that we'll have to manually mark #577 as fixed after it lands, as github only cares about text in commit messages when autoclosing.

rictic:
(I have made that mistake like two dozen times)

src/test/generate_analysis_test.ts
```diff
@@ -33,7 +33,7 @@ suite('generate-elements', () => {
 
   suite('generateAnalysisMetadata', () => {
 
-    suite('generatates for Document array from fixtures', () => {
+    suite('generates for Document array from fixtures', () => {
```
rictic: actually laughing out loud at "generatates"

src/generate-analysis.ts
```diff
@@ -235,6 +250,18 @@ function serializeElementMixin(
   return metadata;
 }
 
+function serializePolymerBehaviorAsElementMixin(
+    behavior: ResolvedPolymerBehavior, packagePath: string): ElementMixin {
+  const metadata = serializeElementLike(behavior, packagePath) as ElementMixin;
+  metadata.name = behavior.className;
+  metadata.privacy = behavior.privacy;
+  metadata.metadata.isPolymerBehavior = true;
```
rictic: prefer:

```javascript
metadata.metadata = {polymer: {isBehavior: true}};
```

The idea is that an analysis object should be able to host multiple metadata sections contributed by multiple frameworks/libraries

src/test/generate_analysis_test.ts
usergenic: generatates

justinfagnani:
Hey @usergenic and @rictic what was the reasoning here? Behaviors aren't mixins and we already had a field for behaviors?

usergenic:
- The motivation was to get all the details of Behaviors into the Analysis to support docs.
 - We decided on the `mixins` property as a way to get them in there:
   - Behaviors are akin to mixins (PRO), though not interchangeable (CON).
   - The goal of reusing `mixins` was to avoid polluting the Analysis format with top-level property for a framework-specific thing (PRO).

TimvdLippe:
> Note that we'll have to manually mark #577 as fixed after it lands, as github only cares about text in commit messages when autoclosing.

Quick note: It also works if you put the issue number in the PR title or description. Easier to fix 😄

PR #581, opened 2017-03-24
Moved the polymer behaviors from Analysis.mixins => Analysis.metadata.polymer.behaviors

usergenic:
 - Prior PR (merged) had PolymerBehaviors serialized as ElementMixins and put into Analysis.mixins.
 - That was iffy, so we decided to move them into Analysis.metadata.polymer.behaviors which feels better.
 - Required addition of Analysis.metadata?: any in the analysis-format.
 - Still serializing them as ElementMixin.  That's open to negotiation, but I had no opinion due to lack of deep familiarity with Behaviors.

### DIFF
--- a/src/generate-analysis.ts
+++ b/src/generate-analysis.ts
@@ -21,8 +21,9 @@ import {Function as ResolvedFunction} from './javascript/function';
 import {Namespace as ResolvedNamespace} from './javascript/namespace';
 import {Document} from './model/document';
 import {Feature} from './model/feature';
-import {Attribute as ResolvedAttribute, Element as ResolvedElement, ElementMixin as ResolvedMixin, Event as ResolvedEvent, Method as ResolvedMethod, PolymerBehavior as ResolvedPolymerBehavior, Property as ResolvedProperty, SourceRange as ResolvedSourceRange} from './model/model';
+import {Attribute as ResolvedAttribute, Element as ResolvedElement, ElementMixin as ResolvedMixin, Event as ResolvedEvent, Method as ResolvedMethod, Property as ResolvedProperty, SourceRange as ResolvedSourceRange} from './model/model';
 import {Package} from './model/package';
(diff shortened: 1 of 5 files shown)


src/generate-analysis.ts
```diff
@@ -124,11 +126,12 @@ function buildAnalysis(members: Members, packagePath: string): Analysis {
     namespace.functions.push(serializeFunction(_function, packagePath));
   }
 
+  // TODO(usergenic): Consider moving framework-specific code to separate file.
   for (const behavior of members.polymerBehaviors) {
     const namespaceName = getNamespaceName(behavior.name);
     const namespace = namespaces.get(namespaceName) || analysis;
-    namespace.mixins = namespace.mixins || [];
-    namespace.mixins.push(
+    setDefaults(namespace, {metadata: {polymer: {behaviors: []}}});
```
rictic: I tend to prefer wordy explicit code when reading this sort of thing, as it's unclear how such functions handle edge cases (is undefined overwritten? is null?). If we need to use it several times then the abstraction pays more rent. If `metadata` was typed I'd also object on those grounds too.

All that said though, this is a fairly minor objection as there aren't any nulls or undefineds here, tests will cover us pretty well, and metadata is untyped, so nonblocking.

usergenic:
@rictic what do we have to do to get type errors reported by `npm test`?  This is the second time I pushed type errors through on analyzer.  Sorry!

usergenic:
FWIW the type errors are fixed in 75228786d34d3f53e9eb11520ff2a6a20c2c070e

usergenic:
Oh and also 4e0a78c

usergenic:
@rictic PTAL at my Type fixes etc.

rictic:
How to make it fail? The fault is in our gulp. I don't recall the details though. The simplest thing is to just put a call to `tsc` in `npm test`.

PR #588, opened 2017-03-29
Extract method params and return types from jsdoc

### LINKED ISSUE #585: Extract method params and return types from jsdoc

aomarks:
Fixes #585 

### DIFF
--- a/src/polymer/js-utils.ts
+++ b/src/polymer/js-utils.ts
@@ -17,7 +17,7 @@ import * as estree from 'estree';
 
 import {closureType, getAttachedComment, objectKeyToString} from '../javascript/esutil';
 import * as jsdoc from '../javascript/jsdoc';
-import {Privacy, ScannedMethod, Severity, SourceRange, Warning} from '../model/model';
+import {MethodParam, Privacy, ScannedMethod, Severity, SourceRange, Warning} from '../model/model';
 
 import {ScannedPolymerProperty} from './polymer-element';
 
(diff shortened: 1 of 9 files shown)


src/polymer/js-utils.ts
```diff
@@ -101,10 +101,38 @@ export function toScannedMethod(
...
-      // formatting to escodegen.
-      return {name: escodegen.generate(param)};
+
+    const paramTags = {};
+    if (scannedMethod.jsdoc) {
+      for (const tag of (scannedMethod.jsdoc.tags || [])) {
+        if (tag.tag === 'param' && tag.name) {
+          paramTags[tag.name] = tag;
+
+        } else if (tag.tag === 'returns') {
```
azakus: According to https://github.com/google/closure-compiler/wiki/Types-in-the-Closure-Type-System#typedefs, `@return` is the correct tag, though both work.

rictic: +1 to also matching `@return`

azakus:
Hmm, there's some oddities in the reified closure types
Somehow `@param {string | !Array<number|string>} path` is turning into
`@param {(string|!Array.<(number|string)>)} path` which is invalid syntax

src/polymer/js-utils.ts
```diff
@@ -101,10 +101,38 @@ export function toScannedMethod(
...
+            type: tag.type || undefined,
+            desc: tag.description || undefined,
+          };
+        }
+      }
+    }
+
+    scannedMethod.params = (value.params || []).map((nodeParam) => {
+      const param: MethodParam = {
+        // With ES6 we can have a lot of param patterns.Best to leave the
```
rictic: space after the `.` in this comment plz

azakus:
@justinfagnani points out that the `@param {(string|!Array.<(number|string)>)}` may actually be valid in closure, so I'm giving it a shot with a full compile pass

rictic:
@azakus `Array.<string>` was valid closure at one point, they're trying to move everyone to `Array<string>` though. Where is the dot coming from? Is the jsdoc node library adding it or is it in the source?

src/polymer/js-utils.ts
aomarks: Done.

src/polymer/js-utils.ts
aomarks: Done.

Funny that http://usejsdoc.org/tags-returns.html likes `@returns` as canonical.

azakus:
@rictic nope, certainly not in the source: https://github.com/Polymer/polymer/blob/master/lib/mixins/property-effects.html#L1391

PR #589, opened 2017-03-29
Make Polymer Analyzer (and tests!) work on Windows

usergenic:
- Lots of subtle issues with Analyzer prevented passing tests on Windows and highlighted actual Windows-related bugs.
 - Mostly these were EOL differences (CRLF vs LF), path separator issues and npm script issues.
 - I learned also that if you change npm script from `x` to `sh -c "x"` you can pretty much keep `x` what it was.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,21 @@
+# Test against this version of Node.js
+environment:
+  nodejs_version: "4.8.0"
+
+# Install scripts. (runs after repo cloning)
+install:
+  # Get the latest stable version of Node.js or io.js
+  - ps: Install-Product node $env:nodejs_version
+  # install modules
+  - npm install
+
+# Post-install test scripts.
+test_script:
+  # Output useful info for debugging.
+  - node --version
+  - npm --version
+  # run tests
+  - npm run test
+
+# Don't actually build.
+build: off
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,6 +1,6 @@
 # Test against this version of Node.js
 environment:
-  nodejs_version: "6.10.0"
+  nodejs_version: "6"
 
 # Install scripts. (runs after repo cloning)
 install:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,6 +1,6 @@
 # Test against this version of Node.js
 environment:
-  nodejs_version: "6.10.2"
+  nodejs_version: "6.10.0"
 
 # Install scripts. (runs after repo cloning)
 install:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -16,6 +16,7 @@ test_script:
   - npm --version
   # run tests
   - npm run test
+  - npm run benchmark
 
 # Don't actually build.
 build: off
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,10 +1,10 @@
 # Test against this version of Node.js
 environment:
-  nodejs_version: "4.8.0"
+  nodejs_version: "6.10.2"
 
 # Install scripts. (runs after repo cloning)
 install:
-  # Get the latest stable version of Node.js or io.js
+  # install node
   - ps: Install-Product node $env:nodejs_version
   # install modules
   - npm install
@@ -15,7 +15,7 @@ test_script:
   - node --version
   - npm --version
   # run tests
-  - npm run test
+  - npm test
   - npm run benchmark
 
 # Don't actually build.
--- a/src/test/generate_analysis_test.ts
+++ b/src/test/generate_analysis_test.ts
@@ -45,7 +45,8 @@ suite('generate-elements', () => {
           }
         } else if (typeof target === 'object') {
           for (const key in target) {
-            if (target.hasOwnProperty(key) && ['file', 'path'].includes(key)) {
+            if (target.hasOwnProperty(key) &&
+                ['file', 'path'].indexOf(key) > -1) {
               target[key] = target[key].replace(/\/|\\/g, path.sep);
             }
           }
--- a/src/test/generate_analysis_test.ts
+++ b/src/test/generate_analysis_test.ts
@@ -21,6 +21,7 @@ import {generateAnalysis, validateAnalysis, ValidationError} from '../generate-a
 import {Document} from '../model/document';
 import {FSUrlLoader} from '../url-loader/fs-url-loader';
 import {PackageUrlResolver} from '../url-loader/package-url-resolver';
+import {usePathSep} from '../utils';
 
 const onlyTests = new Set<string>([]);  // Should be empty when not debugging.
 
@@ -38,19 +39,19 @@ suite('generate-elements', () => {
       // Utility function to convert 'file' and 'path' values to use
       // the current platform separator when comparing golden file to
       // actual file.
-      function usePlatformSep(target: any): any {
+      function transformPathsDeep(target: any): any {
         if (Array.isArray(target)) {
           for (let i = 0; i < target.length; ++i) {
-            usePlatformSep(target[i]);
+            transformPathsDeep(target[i]);
           }
         } else if (typeof target === 'object') {
           for (const key in target) {
             if (target.hasOwnProperty(key)) {
               const value = target[key];
               if (['file', 'path'].indexOf(key) > -1) {
-                target[key] = value.replace(/\/|\\/g, path.sep);
+                target[key] = usePathSep(value);
               } else {
-                usePlatformSep(value);
+                transformPathsDeep(value);
               }
             }
           }
@@ -97,7 +98,7 @@ suite('generate-elements', () => {
 
             const goldenAnalysis =
                 JSON.parse(fs.readFileSync(pathToGolden, 'utf-8'));
-            usePlatformSep(goldenAnalysis);
+            transformPathsDeep(goldenAnalysis);
 
             try {
               assert.deepEqual(
--- a/src/test/generate_analysis_test.ts
+++ b/src/test/generate_analysis_test.ts
@@ -36,28 +36,6 @@ suite('generate-elements', () => {
 
     suite('generates for Document array from fixtures', () => {
 
-      // Utility function to convert 'file' and 'path' values to use
-      // the current platform separator when comparing golden file to
-      // actual file.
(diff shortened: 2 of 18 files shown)


usergenic:
Green!

src/warning/warning-printer.ts
```diff
@@ -120,7 +121,7 @@ export class WarningPrinter {
   private async _getLinesOfText(
       startLine: number, endLine: number, localPath: string) {
     const contents = await this._options.analyzer.load(localPath);
-    return contents.split('\n').slice(startLine, endLine + 1);
+    return contents.split(EOL).slice(startLine, endLine + 1);
```
usergenic: I've dropped standalone `\r` from EOL case (since its archaic: Mac OS 9 and earlier)

Would you still disagree then?

src/url-loader/fs-url-loader.ts
```diff
@@ -29,13 +29,13 @@ export class FSUrlLoader implements UrlLoader {
   root: string;
 
   constructor(root?: string) {
-    this.root = root || '';
+    this.root = usePathSep(root || '');
```
usergenic: The root is just a string and it could come from a file that had path in a relative 'path/to/here' form.

usergenic: Not saying that relative/unresolved paths are recommended, but people do that stuff all day long.

src/utils.ts
```diff
@@ -11,10 +11,21 @@
...
  */
+import * as path from 'path';
 import {parse as parseUrl_, Url} from 'url';
 
 const unspecifiedProtocol = '-:';
 
+export const EOL = /\r\n|\r|\n/g;
+export const LF = '\n';
+
+export function lfify(text: string|string[]): string {
```
usergenic: I went with normalizeNewlines and normalizePathSeparators and added jsdocs.

src/utils.ts
```diff
@@ -11,10 +11,21 @@
  * subject to an additional IP rights grant found at
  * http://polymer.github.io/PATENTS.txt
  */
+import * as path from 'path';
 import {parse as parseUrl_, Url} from 'url';
 
 const unspecifiedProtocol = '-:';
 
+export const EOL = /\r\n|\r|\n/g;
```
usergenic: Okay I'm *old*.  `\r` by itself ended at Mac OS 9.  And nobody writes on Commodores now so I have removed `\r`.

src/utils.ts
```diff
@@ -25,6 +36,14 @@ export function parseUrl(url: string): Url {
   return urlObject;
 }
 
+export function posixify(pathname: string): string {
+  return pathname.replace(/\\/g, '/');
```
usergenic: Apparently they are valid in posix-land but not Windows-land.  For this reason, I've made the path transform functions no-ops in posix-land.

src/utils.ts
```diff
@@ -25,6 +36,14 @@ export function parseUrl(url: string): Url {
...
+}
+
+export function usePathSep(pathname: string): string {
```
usergenic: I went with `normalizePathSeparators` -if that's too vague I can change to your suggestion.

src/test/generate_analysis_test.ts
```diff
@@ -34,6 +35,29 @@ suite('generate-elements', () => {
   suite('generateAnalysisMetadata', () => {
 
     suite('generates for Document array from fixtures', () => {
+
+      // Utility function to convert 'file' and 'path' values to use
```
usergenic: Changed the behavior of generateAnalysis to convert paths to posix.  That makes total sense because they're all package relative.

src/utils.ts
```diff
@@ -11,10 +11,40 @@
  * subject to an additional IP rights grant found at
  * http://polymer.github.io/PATENTS.txt
  */
+import * as path from 'path';
 import {parse as parseUrl_, Url} from 'url';
 
 const unspecifiedProtocol = '-:';
 
+export const EOL = /\r\n|\n/g;
```
justinfagnani: Add a comment so we can see it in hover-over docs

src/utils.ts
```diff
@@ -11,10 +11,40 @@
...
+export const LF = '\n';
```
justinfagnani: same

src/utils.ts
```diff
@@ -11,10 +11,40 @@
...
+}
+
+/**
+ * Ensure path separators used match the platform's separator.
+ * Backslash *is* a valid filename character in a posix
+ * environment, so this function is a no-op there.
+ *
+ * @param pathname - file path to transform.
+ */
+export function normalizePathSeparators(pathname: string): string {
```
justinfagnani: I wonder if normalize is the right term here - it implies to me a canonical format, where this is converting to the platform format. How about `toNativePathFormat`?

cc @rictic @FredKSchott 

Also, TODO this for inclusion in a common package for dealing with paths and URLs in a typesafe way.

src/utils.ts
```diff
@@ -25,6 +55,21 @@ export function parseUrl(url: string): Url {
...
 
+/**
+ * Ensure path separators of the posix forward-slash form when
+ * running on environment (Windows) where path separator is a
+ * backslash.  Backslash *is* a valid filename character in a
+ * posix environment, so this function is a no-op there.
+ *
+ * @param pathname - file path to transform.
+ */
+export function posixify(pathname: string): string {
```
justinfagnani: Is it a problem that a round-trip between `posixify` and `normalizePathSeparators` is potentially lossy? Should we throw in cases where the input contains a `/` on windows?

appveyor.yml
```diff
@@ -0,0 +1,22 @@
+# Test against this version of Node.js
+environment:
+  nodejs_version: "4.8.0"
```
usergenic: Done.

appveyor.yml
```diff
@@ -0,0 +1,22 @@
...
+
+# Install scripts. (runs after repo cloning)
+install:
+  # Get the latest stable version of Node.js or io.js
```
usergenic: Fixed.

appveyor.yml
```diff
@@ -0,0 +1,22 @@
...
+  # install modules
+  - npm install
+
+# Post-install test scripts.
+test_script:
+  # Output useful info for debugging.
+  - node --version
+  - npm --version
+  # run tests
+  - npm run test
```
usergenic: Trimmed.

src/utils.ts
usergenic: Added.

src/utils.ts
usergenic: Done.

src/utils.ts
usergenic: @rictic had previously suggested "convertToPlatformsPathSeparators" which I think is the least ambiguous.  I'll go with that.  Will come up with something snappier when we make a common package.

Added TODO.

src/utils.ts
usergenic: The thing about windows is that it is presently agnostic about '/' and '\' and generally treats them the same, so we fully expect that they can come in already as '/'.

I don't perceive it as lossy for us to flip them around.

src/warning/warning-printer.ts
rictic: I can't spot any bugs. It feels a bit iffy to define source ranges by `\n` but to handle both `\r\n` and `\n` here. OTOH, printing a standalone `\r` is a great way to get super weird unexpected console output, so let's try this and see what happens. Worst case I is that we make some off-by-one mistakes when printing squiggles on windows.

Hm, it's probably worth adding a test of printing a warning in a document with `\r\n`s

appveyor.yml
```diff
@@ -0,0 +1,22 @@
+# Test against this version of Node.js
+environment:
+  nodejs_version: "6.10.2"
```
rictic: I wonder if asking for just "6" would work

rictic:
Looking good!

usergenic:
> I can't spot any bugs. It feels a bit iffy to define source ranges by \n but to handle both \r\n and \n here. OTOH, printing a standalone \r is a great way to get super weird unexpected console output, so let's try this and see what happens. Worst case I is that we make some off-by-one mistakes when printing squiggles on windows.

We have A LOT of tests that use the warning printer squiggles and they ALL fail when run on windows when there's one extra squiggle on each line because string-comparison.

> Hm, it's probably worth adding a test of printing a warning in a document with \r\ns

How would you feel if we do a conversion of `\r\n` to `\n` every time we load contents into the analyzer and warnings printer so it only ever has to deal with `\n`?

I don't know the ways in which this would bite us down the road, but it doesn't feel terrible...?

rictic:
> We have A LOT of tests that use the warning printer squiggles and they ALL fail when run on windows when there's one extra squiggle on each line because string-comparison.

Can you clarify? Like, without changing this `\n` to `\r\n` then we get a bunch of test failures? In that case, it sounds like this change is great and I'll shut up :)

> How would you feel if we do a conversion of `\r\n` to `\n` every time we load contents into the analyzer and warnings printer so it only ever has to deal with `\n`?

The source ranges are generated by a mixture of our code and by the parsers that we use. The only place that we could end-run around this is to modify the string contents immediately after loading, but that seems like it's too aggressive. The `\r` characters could be semantically meaningful. If warning printer tests were failing on Windows without this change and they're passing now then I'm happy.

usergenic:
> The \r characters could be semantically meaningful. If warning printer tests were failing on Windows without this change and they're passing now then I'm happy.

Okay.  Yeah the issue is that the `\r` characters are rendered spaces by warning printer and underlined when the source range spans lines to or across the end of a line.

rictic:
It looks like appveyor isn't happy?

rictic:
This may be the fix of some issues `bahrus` from slack is seeing on windows, where we're not properly stripping whitespace or `*`s from jsdoc comments.

FredKSchott:
Closing this PR as abandoned.

@usergenic please take a look and rebase/reopen if this is something you'd still like to get merged in.

usergenic:
Yeah this is dead dead dead.  I tried *really* hard to get it working on windows.  Will take a stab at this after 3.0 rolls around.  The new resolved urls might help a lot.

PR #601, opened 2017-04-04
Fix dependants in DependencyRecord

rictic:

### DIFF
--- a/src/core/dependency-graph.ts
+++ b/src/core/dependency-graph.ts
@@ -153,7 +153,9 @@ export class DependencyGraph {
         }
       }
       fork._documents.delete(path);
-      // We won't be told again about our dependants again later
+      // If there are dependents on this record, we must preserve them,
+      // as they're only added with an addDocument() call, and there are
+      // never repeated addDocument() calls for the same path.
       if (record.dependants.size > 0) {
         const newRecord = fork._getRecordFor(record.url);
         for (const dependant of record.dependants) {
--- a/src/core/dependency-graph.ts
+++ b/src/core/dependency-graph.ts
@@ -195,7 +195,7 @@ export class DependencyGraph {
 
   toString() {
     return Array.from(this._documents.values())
-        .map(dr => dr.toString())
+        .map((dr) => dr.toString())
         .join('\n');
   }
 }
--- a/src/core/dependency-graph.ts
+++ b/src/core/dependency-graph.ts
@@ -54,6 +54,17 @@ class DocumentRecord {
   get dependenciesKnown(): Promise<Set<string>> {
     return this.dependenciesDeferred.promise;
   }
+
+  toString(): string {
+    let s = `${this.url}\n`;
+    for (const dependency of this.dependencies) {
+      s += `  -> ${dependency}\n`;
+    }
+    for (const dependant of this.dependants) {
+      s += `  <- ${dependant}\n`;
+    }
+    return s;
+  }
 }
 
 /**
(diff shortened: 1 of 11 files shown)


src/test/analyzer_test.ts
```diff
@@ -779,6 +779,41 @@ var DuplicateNamespace = {};
...
+        'Unable to resolve superclass BaseElement');
+
+    const b3Doc = await analyzer.analyze('base.js', `
+        class BaseElement extends HTMLElement {}
+        customElements.define('base-elem', BaseElement);
+    `);
+    assert.deepEqual(b3Doc.getWarnings(), []);
+    const u3Doc = await analyzer.analyze('user.html');
+    assert.equal(u3Doc.getWarnings().length, 0);
+  });
```
rictic: This is the test of the specific invalid behavior that inspired this work.

I added more specific tests for the issue in dependency-graph_test as well below.

src/core/dependency-graph.ts
```diff
@@ -84,21 +95,17 @@ export class DependencyGraph {
...
    */
   addDocument(url: string, dependencies: Iterable<string>) {
     const record = this._getRecordFor(url);
     for (const dependency of dependencies) {
       record.dependencies.add(dependency);
-      const dependantRecord = this._getRecordFor(dependency);
-      dependantRecord.dependants.add(url);
-    }
-    try {
-      record.dependenciesDeferred.resolve(record.dependencies);
```
justinfagnani: what was this guarding against?

src/core/dependency-graph.ts
```diff
@@ -146,6 +153,13 @@ export class DependencyGraph {
         }
       }
       fork._documents.delete(path);
+      // We won't be told again about our dependants again later
```
justinfagnani: not sure what the comment means really, what are we doing because of that?

src/test/analyzer_test.ts
```diff
@@ -779,6 +779,41 @@ var DuplicateNamespace = {};
     });
   });
 
+  test('maintain caches across multiple edits', async() => {
+    const b1Doc = await analyzer.analyze('base.js', `
```
justinfagnani: can you describe briefly in prose the scenario that you're testing?

src/test/core/dependency-graph_test.ts
```diff
@@ -149,8 +166,49 @@ suite('DependencyGraph', () => {
...
+          false,
+          true,
+          `found unsettled record for url '${
+                                             record.url
+                                           }' in graph that should be settled`);
+    }
+  }
+}
+
+function assertIsValidGraph(graph: DependencyGraph) {
```
justinfagnani: add doc describing the invariants it checks?

src/test/core/dependency-graph_test.ts
```diff
@@ -149,8 +166,49 @@ suite('DependencyGraph', () => {
...
       await Promise.all([aReady, bReady]);
+      assertGraphIsSettled(graph);
+      assertIsValidGraph(graph);
     });
 
   });
 
 });
+
+function assertGraphIsSettled(graph: DependencyGraph) {
```
justinfagnani: docs?

src/core/dependency-graph.ts
rictic: Earlier this had a console.log here because of a worry that this would fail in a difficult to debug way. That doesn't seem to be happening, and we removed the console log but not the silly catch and throw

src/test/analyzer_test.ts
rictic: done

src/test/analyzer_test.ts
rictic: done

src/test/core/dependency-graph_test.ts
rictic: done

src/test/core/dependency-graph_test.ts
rictic: done

PR #603, opened 2017-04-05
Handle tab prefix in Warning Printer getSquiggleUnderline.

coreyfarrell:
This change deals with output of warnings for lines that are indented by tabs.  Previously tabs were counted as a single character causing the squiggle underline to be indented by a single space.  Now the prefix uses tabs to prefix the squiggle where tabs were used on the source line.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
    Unreleased section, uncommenting the header as necessary.
 -->
 
-<!--## Unreleased-->
+## Unreleased
+
+* The warning printer displays the squiggle underline in the correct place on lines indented by tabs.
 
 ## [2.0.0-alpha.35] - 2017-04-05
 
--- a/src/warning/warning-printer.ts
+++ b/src/warning/warning-printer.ts
@@ -137,7 +137,7 @@ function getSquiggleUnderline(
     const endColumn = sourceRange.end.line === sourceRange.start.line ?
         sourceRange.end.column :
         lineText.length;
-    const prefix = ' '.repeat(startColumn);
+    const prefix = lineText.slice(0, startColumn).replace(/[^\t]/g, ' ');
     if (startColumn === endColumn) {
       return prefix + '~';  // always draw at least one squiggle
     }


src/warning/warning-printer.ts
```diff
@@ -137,7 +137,7 @@ function getSquiggleUnderline(
     const endColumn = sourceRange.end.line === sourceRange.start.line ?
         sourceRange.end.column :
         lineText.length;
-    const prefix = ' '.repeat(startColumn);
+    const prefix = lineText.replace(/[^\t]/g, ' ').substr(0, startColumn);
```
rictic: This is close, but I don't think this is quite right. prefix must be only whitespace, and `startColumn` might be after non-whitespace characters. E.g. suppose we want to underline MyElem in this line:

```javascript
class MyElem extends HTMLElement {
```

I think the simplest correct solution would be to match tab character for tab character of the original input. Something like:

```javascript
const prefix = Array.from(
    lineText.slice(0, startColumn)).map((c) => c === '\t' ? '\t' : ' ').join('');
```

coreyfarrell: I think your code does the same thing as mine in a different way.  My patch was texted against a tab indented instance of:
``````

`importPath` was underlined due to https://github.com/Polymer/polymer-linter/issues/71.  This patch puts the underline in the correct place, the lineText.replace('/[^\t]/g', ' ') replaces all characters that are not tabs, the substr trims to the correct number of characters for the prefix.

File: CHANGELOG.md [hunk 2]
```diff
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 * Jsdoc `@demo` annotations are now added to `demos` collection for `Element`, `ElementMixin`, `PolymerElement` and `Behavior`.
 * Types and descriptions are now extracted from method @param and @returns jsdoc annotations.
 * Fixed caching issue such that Documents would not always have information from the latest versions of their dependencies.
+* The warning printer displays the squiggle underline in the correct place on lines indented by tabs.
```
rictic: We stamped alpha.35 just now. Please uncomment Unreleased above and put this bullet point under it.

src/warning/warning-printer.ts
rictic: Oh, I see, yes! I misread the `[^\t]` part of your regex, easy to forget that `^` means to invert the character set. It should have been a clue that all the tests pass! :)

The only recommendations left would be to prefer the `slice` method to the `substr` method, and to do the slice before the replace, as a microoptimization.

coreyfarrell:
I've made the requested changes.  May I ask what is the reason to prefer `slice` over `substr`?

rictic:
All three methods (slice, substr, and substring) have idiosyncracies, but slice is present in the same form on both strings and arrays, so it's less weirdness to try and keep in one's head. e.g. http://stackoverflow.com/a/2243835

rictic:
Thanks for the PR!

PR #613, opened 2017-04-10
Added a canLoad method to Analyzer

usergenic:
- `Analyzer` and `AnalysisContext` have public `canLoad` methods now.
- `AnalysisContext` uses its own `canLoad` method now.
- Opportunistically added missing comments and type annotations.
- Fixes https://github.com/Polymer/polymer-analyzer/issues/612

### DIFF
--- a/src/analyzer.ts
+++ b/src/analyzer.ts
@@ -117,14 +117,18 @@ export class Analyzer {
   }
 
   /**
-   * Returns `true` if the provided resolved URL can be loaded.
+   * Returns `true` if the provided resolved URL can be loaded.  Obeys the
+   * semantics defined by `UrlLoader` and should only be used to check
+   * resolved URLs.
    */
   canLoad(resolvedUrl: string): boolean {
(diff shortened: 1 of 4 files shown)


src/analyzer.ts
```diff
@@ -117,16 +117,29 @@ export class Analyzer {
   }
 
   /**
+   * Returns `true` if the provided resolved URL can be loaded.
```
rictic: Thanks for documenting these.

Please also mention on each of them that the semantics are determined by the configured url loader/resolver, and that only resolved urls should ever be loading.

usergenic: `Analyzer implements UrlLoader` would say as much?  But that feels so wrong.  I'll just put it in the comments for now.

usergenic: Although, I'm looking forward to updating `polymer-bundler` to say `new InMemoryOverlayUrlLoader(opts.analyzer)`

PR #616, opened 2017-04-11
Moved removeFakeNodes to the html-parser so nobody gets fake nodes in their resolved document ASTs.

usergenic:
As a users of the `ast` property of parsed html documents, I want a tree that represents what's *really* in the document, not what parse5 is pretending is there.  I moved the removeFakeNodes code to the parser and all the tests are passing.  WDYT?

### DIFF
--- a/src/html/html-document.ts
+++ b/src/html/html-document.ts
@@ -197,22 +197,7 @@ export class ParsedHtmlDocument extends ParsedDocument<ASTNode, HtmlVisitor> {
       }) + '  '.repeat(expectedIndentation - 1));
     }
 
-    removeFakeNodes(selfClone.ast);
     return parse5.serialize(selfClone.ast);
-  }
-}
-
-const injectedTagNames = new Set(['html', 'head', 'body']);
-function removeFakeNodes(ast: dom5.Node) {
-  const children = (ast.childNodes || []).slice();
-  if (ast.parentNode && !ast.__location && injectedTagNames.has(ast.nodeName)) {
-    for (const child of children) {
-      dom5.insertBefore(ast.parentNode, ast, child);
-    }
-    dom5.remove(ast);
-  }
-  for (const child of children) {
-    removeFakeNodes(child);
   }
 }
 
--- a/src/html/html-parser.ts
+++ b/src/html/html-parser.ts
@@ -12,7 +12,7 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
-import {getAttribute, predicates as p, query} from 'dom5';
+import {getAttribute, insertBefore, Node, predicates as p, query, remove} from 'dom5';
 import {parse as parseHtml} from 'parse5';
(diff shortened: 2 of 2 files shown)


rictic:
The downside of this is that the ast is less like it is in the browser. It kinda seems to me that this is a serialization-time concern.

I suppose the advantage of this approach is that we don't need to do a full clone whenever we do a minimal serialization of a document. Hm.

rictic:
Do we have a sense of the performance cost of the extra clone?

usergenic:
If anything is going to need to be lighter-weight it should be parsing, I think.  This PR puts an extra burden on the parse step since it now has to walk nodes to find potentially location-less nodes to remove.

On the plus side, the ast on the document can as-is be `parse5.serialize()`'d and otherwise treated as an authoritative representation of the contents by parse5.serialize() etc.  So, for example, a scanner that looks for a body tag could trust in the absence of a body tag in the ast instead of having to take extra pains to verify the `__location` data on the node.  In that sense I think it makes sense for the ast to be "pure".

usergenic:
@rictic also there's a selfish reason for this change which is that I won't have to copy the `removeFakeNodes` function into Bundler to generate bundled html to re-analyze this way too.

rictic:
If we don't remove fake nodes at parse time, the bundler will have to clone html asts in order to modify them. I'd like to have a sense of the performance impact. If it's significant, that argues for doing it at parse time.

As far as the duplicated code argument goes, we could export the function.

usergenic:
@rictic the bundler ought to be cloning them anyways because its mutating trees, both the imported AST when modifying it for inlining and the bundle document's AST itself when being inlined into prior to re-analysis.  In that sense, the behavior of the Analyzer to remove the fake nodes or not defines no additional cost in the bundler in terms of cloning.

The primary issue is whether we want the ParsedDocument.ast to represent the nodes actually present in the document.  My personal opinion is that the analyzer should prune the tree because it means users don't have to be aware of or special case handling to check for the presence of `__location` on a node.

usergenic:
> The downside of this is that the ast is less like it is in the browser.

I feel like I'm missing something important if this is true.  Can you help me understand this point?

rictic:
parse5 is doing the right thing here. The HTML5 parse algorithm is specced to do a bunch of error recovery work, including hallucinating nodes. So in both parse5 and the browser, the document `<div>hello</div>` has a body element.

I can't immediately point to a case where this is what we want during static analysis, but it does seem like the right thing.

usergenic:
Ok.  I don't *like* it, but ok.  Considering the following code...

```html
<style>
h1 { color: blue; }
body h1 { color: red; }
</style>

<h1>Cash me outside</h1>

<script>
document.querySelector('body h1').innerText = 'How bow dat?';
</script>
```

Does produce this...
howbowdat

I will accept it.  Bundler has to do a bunch of fake node removal now tho *sigh*.  Will close this PR.

PR #617, opened 2017-04-12
Export the removeFakeNodes function from `html-document`.

usergenic:
Want to deduplicate the function currently being copied into Build and Bundler to deal with AST serialization problems in https://github.com/Polymer/polymer-build/issues/173


justinfagnani:
Shouldn't this go into dom5?

rictic:
Plan is to instead pull this in from dom5 I think

PR #620, opened 2017-04-13
Fix compile to work with better chai typings.

rictic:
Chai equality typings now verify that the type on the left hand side matches the type on the right hand side. This caught a few places in tests that we were playing fast and loose, as well as one place where we had null where our types said undefined.

### DIFF
--- a/src/polymer/polymer2-element-scanner.ts
+++ b/src/polymer/polymer2-element-scanner.ts
@@ -380,7 +380,7 @@ class ClassFinder implements Visitor {
       return;
     }
 
-    const name = node.id && astValue.getIdentifierName(node.id);
+    const name = (node.id && astValue.getIdentifierName(node.id)) || undefined;
     const comment = esutil.getAttachedComment(node) ||
         esutil.getAttachedComment(parent) || '';
     this._classFound(name, jsdoc.parseJsdoc(comment), node);
--- a/src/polymer/polymer2-element-scanner.ts
+++ b/src/polymer/polymer2-element-scanner.ts
@@ -380,7 +380,7 @@ class ClassFinder implements Visitor {
       return;
     }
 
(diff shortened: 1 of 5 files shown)


src/polymer/polymer2-element-scanner.ts
```diff
@@ -380,7 +380,7 @@ class ClassFinder implements Visitor {
       return;
     }
 
-    const name = node.id && astValue.getIdentifierName(node.id);
+    const name = (node.id && astValue.getIdentifierName(node.id)) || undefined;
```
justinfagnani: `|| undefined` is unusual. Why is it necessary, and why for an update to Chai types?

justinfagnani: I think I'd prefer `const name = node.id ? astValue.getIdentifierName(node.id) : undefined;` as more explicit, and shorter than the bool operator version.

src/test/scanning/scan_test.ts
```diff
@@ -23,7 +23,7 @@ import {invertPromise} from '../test-utils';
 suite('scan()', () => {
 
   test('calls Scanner.scan', async() => {
-    const feature = Symbol('feature');
+    const feature = Symbol('feature') as any;
```
justinfagnani: er :(

rictic: Yep. Same with the dependency graph tests, where we treat strings as ParsedDocuments for easy comparison. I'm fine with doing a more comprehensive fix later but this is just to unblock

src/polymer/polymer2-element-scanner.ts
rictic: +1, done

rictic:
As this is largely a simple noop change I'm gonna merge and release to unblock Brendan's bundler work. Happy to address other issues in follow up

PR #624, opened 2017-04-14
Fix a couple issues that came up in the linter's integration test

rictic:

Works around https://github.com/Polymer/polymer-analyzer/issues/623 and also fixes an issue with PackageUrlResolver

### DIFF
--- a/src/core/analysis-context.ts
+++ b/src/core/analysis-context.ts
@@ -68,8 +68,10 @@ export class AnalysisContext {
     ['json', new JsonParser()],
   ]);
 
-  private readonly _languageAnalyzers =
-      new Map<string, LanguageAnalyzer<any>>([]);
+  private readonly _languageAnalyzers = new Map<string, LanguageAnalyzer<any>>([
+    // TODO(rictic): add typescript language analyzer back after investigating
+    //     https://github.com/Polymer/polymer-analyzer/issues/623
+  ]);
 
   /** A map from import url to urls that document lazily depends on. */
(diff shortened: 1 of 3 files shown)


src/core/analysis-context.ts
```diff
@@ -69,9 +68,8 @@ export class AnalysisContext {
     ['json', new JsonParser()],
   ]);
 
-  private readonly _languageAnalyzers = new Map<string, LanguageAnalyzer<any>>([
-    ['ts', new TypeScriptAnalyzer(this)],
-  ]);
+  private readonly _languageAnalyzers =
```
justinfagnani: Can you add a TODO to add back and reference to the issue?

src/url-loader/package-url-resolver.ts
```diff
@@ -39,7 +39,14 @@ export class PackageUrlResolver implements UrlResolver {
 
   canResolve(url: string): boolean {
     const urlObject = parseUrl(url);
-    const pathname = pathlib.normalize(decodeURI(urlObject.pathname || ''));
+    let decodedUrl;
+    try {
+      decodedUrl = decodeURI(urlObject.pathname || '');
+    } catch (e) {
+      return false;
```
justinfagnani: this is the kind of thing that makes me like checked exceptions, against all current attitudes...

rictic: `Result<string, Error>` is Good and Right, IMO

src/core/analysis-context.ts
rictic: Done

PR #625, opened 2017-04-16
Add a general concept of a Class, emit them when no other feature matches

rictic:

This also further cleans up the resolved element prototype tree, eliminating a lot of duplication. Our scanned element types are still crazytown but this is progress.


src/model/class.ts
```diff
@@ -0,0 +1,246 @@
...
+            code: 'unknown-superclass',
+            sourceRange: scannedElement.superClass.sourceRange!,
+          });
+        }
+      }
+    }
+    return undefined;
+  }
+
+  private _applyMixins(scannedClass: ScannedClass, document: Document) {
```
justinfagnani: I neglected to do this with `jsdoc.getMixins`, and `PolymerElement.inheritFrom`, but I think we need to reinforce the order that mixins apply in every function that deals with lists of mixins, because even I forget this. We also need to make sure this gets into docs.

The way we apply mixins @ master assumes we go from innermost (higher on the prototype chain, closer to Object) to outermost (lower, closer to concrete class). However I noticed in Polymer core that in the cases where there's more than one mixin they list from outermost to innermost.

I think the order that master assumes is the right one, because you can apply mixins as you read down the file - the mnemonic is "apply this mixin, then this mixin"...

rictic:
I'm rebasing this on top of #630 because it's a much cleaner base that will enable us to do the right thing here.

CHANGELOG.md
```diff
@@ -11,7 +11,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
    Unreleased section, uncommenting the header as necessary.
 -->
 
-<!--## Unreleased-->
+## Unreleased
+
+* Add a Class feature kind. We emit these when we encounter classes that aren't anything more specific (i.e. not an element, mixin, ...).
```
rictic: Clarified

src/generate-analysis.ts
```diff
@@ -35,6 +36,7 @@ interface Members {
   namespaces: Set<ResolvedNamespace>;
   functions: Set<ResolvedFunction>;
   polymerBehaviors: Set<ResolvedPolymerBehavior>;
+  classes: Set<ResolvedClass>;
 }
 
 export function generateAnalysis(
```
rictic: done

src/generate-analysis.ts
```diff
@@ -120,11 +127,11 @@ function buildAnalysis(members: Members, packagePath: string): Analysis {
     namespace.mixins.push(serializeElementMixin(mixin, packagePath));
   }
 
-  for (const _function of members.functions) {
-    const namespaceName = getNamespaceName(_function.name);
+  for (const fnction of members.functions) {
```
rictic: Underscore-prefixed names don't warn if they're unused. Trailing underscore?

src/model/class.ts
```diff
@@ -0,0 +1,246 @@
...
+    this.privacy = scannedClass.privacy;
+
+    if (scannedClass.superClass) {
+      this.superClass = scannedClass.superClass.resolve(document);
+    }
+    this.methods.push(...scannedClass.methods);
+    this.properties.push(...scannedClass.properties);
+
+    this._applySuperClass(scannedClass, document);
+    this._applyMixins(scannedClass, document);
```
rictic: Done

src/model/class.ts
rictic: Done

src/analysis-format.ts
```diff
@@ -117,9 +118,10 @@ export interface Namespace extends Feature {
   functions?: Function[];
   mixins?: ElementMixin[];
   namespaces?: Namespace[];
+  classes?: Class[];
```
rictic: Clarified

src/generate-analysis.ts
justinfagnani: +1

src/model/class.ts
justinfagnani: thanks. we still need to doc this everywhere...

src/model/class.ts
```diff
@@ -0,0 +1,344 @@
...
+   * members override those before it in the prototype chain.
+   */
+  readonly mixins: Reference[] = [];
+  readonly abstract: boolean;
+  readonly privacy: Privacy;
+  readonly demos: Demo[];
+
+  constructor(init: ClassInit, document: Document) {
+    this.sourceRange = init.sourceRange;
+    this.warnings = init.warnings ? Array.from(init.warnings) : [];
```
justinfagnani: this.warnings = init.warnings === undefined ? [] : Array.from(init.warnings);

src/model/class.ts
```diff
@@ -0,0 +1,344 @@
...
+    if (this.name) {
+      this.identifiers.add(this.name);
+    }
+    this.jsdoc = init.jsdoc;
+    this.description = init.description;
+    this.summary = init.summary;
+    this.abstract = init.abstract;
+    this.privacy = init.privacy;
+    this.demos = init.demos || [];
+    this.warnings = init.warnings ? Array.from(init.warnings) : [];
```
justinfagnani: double-initializing warnings

src/model/class.ts
```diff
@@ -0,0 +1,344 @@
...
+    this.name = init.name || init.className;
+    if (this.name) {
+      this.identifiers.add(this.name);
+    }
+    this.jsdoc = init.jsdoc;
+    this.description = init.description;
+    this.summary = init.summary;
+    this.abstract = init.abstract;
+    this.privacy = init.privacy;
+    this.demos = init.demos || [];
```
justinfagnani: this is an example where we could use all of the features of destructuring assignment:

```ts
{
  jsdoc: this.jsdoc,
  description: this.description,
  demos: this.demos = [],
}
```

src/model/class.ts
rictic: done

src/model/class.ts
rictic: that clarified the code a bit, though typescript thinks that that assignment to `this.demos` is invalid, it doesn't realize that that pattern ensures it's not undefined

src/model/class.ts
rictic: fixed

Issue #564, opened 2017-03-14
Mix mixins into mixins


rictic:
Currently ElementMixin has its own array of mixins, but those mixins aren't mixed into the ElementMixin. i.e. if A and B are mixins, and A includes B, A won't directly include B's properties, methods etc, unlike Elements. This is counterintuitive, and makes it harder to work with mixins. We should make them more like Elements in this respect.

arthurevans:
This also means that almost none of the methods and properties in Polymer.Element are actually documented, since most of them come in through multiple levels of mixins. 

I'd be inclined to bump the priority up on this one since it's a regression compared to 1.x in the amount of coverage we have on the most-used API reference.

PR #628, opened 2017-04-17
Mix mixins into mixins.

rictic:
Fixes https://github.com/Polymer/polymer-analyzer/issues/564

Also removed some use of `Object.assign` in constructors

### DIFF
--- a/src/polymer/polymer-element-mixin.ts
+++ b/src/polymer/polymer-element-mixin.ts
@@ -19,7 +19,7 @@ import {Document, ElementMixin, Method, Privacy, ScannedElementMixin, ScannedMet
 
 import {ScannedBehaviorAssignment} from './behavior';
 import {getOrInferPrivacy} from './js-utils';
-import {addMethod, addProperty, LocalId, Observer, PolymerExtension, PolymerProperty, ScannedPolymerExtension, ScannedPolymerProperty} from './polymer-element';
+import {addMethod, addProperty, applyMixins, LocalId, Observer, PolymerExtension, PolymerProperty, ScannedPolymerExtension, ScannedPolymerProperty} from './polymer-element';
 
 export interface Options {
   name: string;
@@ -34,24 +34,38 @@ export interface Options {
 
 export class ScannedPolymerElementMixin extends ScannedElementMixin implements
     ScannedPolymerExtension {
-  properties: ScannedPolymerProperty[] = [];
-  methods: ScannedMethod[] = [];
-  observers: Observer[] = [];
-  listeners: {event: string, handler: string}[] = [];
-  behaviorAssignments: ScannedBehaviorAssignment[] = [];
+  readonly properties: ScannedPolymerProperty[] = [];
+  readonly methods: ScannedMethod[] = [];
+  readonly observers: Observer[] = [];
+  readonly listeners: {event: string, handler: string}[] = [];
+  readonly behaviorAssignments: ScannedBehaviorAssignment[] = [];
   // FIXME(rictic): domModule and scriptElement aren't known at a file local
   //     level. Remove them here, they should only exist on PolymerElement.
   domModule: dom5.Node|undefined = undefined;
   scriptElement: dom5.Node|undefined = undefined;
   pseudo: boolean = false;
-  abstract: boolean = false;
+  readonly abstract: boolean = false;
+  readonly sourceRange: SourceRange;
 
-  constructor(options?: Options) {
+  constructor({
+    name,
+    jsdoc,
+    description,
+    summary,
+    privacy,
+    sourceRange,
+    mixins,
+    astNode
+  }: Options) {
     super();
-    // TODO(justinfagnani): fix this constructor to not be crazy, or remove
-    // class altogether.
-    const optionsCopy = Object.assign({}, options) as Options;
-    Object.assign(this, optionsCopy);
+    this.name = name;
+    this.jsdoc = jsdoc;
+    this.description = description;
+    this.summary = summary;
+    this.privacy = privacy;
+    this.sourceRange = sourceRange;
+    this.mixins = mixins;
+    this.astNode = astNode;
   }
 
   addProperty(prop: ScannedPolymerProperty) {
@@ -59,25 +73,13 @@ export class ScannedPolymerElementMixin extends ScannedElementMixin implements
   }
 
   addMethod(method: ScannedMethod) {
+    // methods are only public by default if they're documented.
+    method.privacy = getOrInferPrivacy(method.name, method.jsdoc, true);
     addMethod(this, method);
   }
 
   resolve(document: Document): PolymerElementMixin {
-    const element = new PolymerElementMixin();
-    Object.assign(element, this);
-
-    for (const method of element.methods) {
-      // methods are only public by default if they're documented.
-      method.privacy = getOrInferPrivacy(method.name, method.jsdoc, true);
-    }
-    element.mixins = [];
-    for (const mixin of this.mixins) {
-      // TODO(rictic): we should mix these mixins into `this`. See
-      // PolymerElement's logic for this.
-      element.mixins.push(mixin.resolve(document));
-    }
-
-    return element;
+    return new PolymerElementMixin(this, document);
   }
 }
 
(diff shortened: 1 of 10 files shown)


src/polymer/polymer-element-mixin.ts
```diff
@@ -34,50 +34,52 @@ export interface Options {
...
+  constructor({
+    name,
+    jsdoc,
+    description,
+    summary,
+    privacy,
+    sourceRange,
+    mixins,
+    astNode
+  }: Options) {
```
usergenic: This pattern is my new favorite thing.

src/test/polymer/polymer-element_test.ts
```diff
@@ -57,12 +57,15 @@ suite('PolymerElement', () => {
       attributes: element.attributes.map((a) => ({
                                            name: a.name,
                                          })),
-      methods: element.methods.map(
-          (m) => ({name: m.name, params: m.params, return: m.return })),
+      methods: element.methods.map((m) => ({
+                                     name: m.name,
+                                     params: m.params, return: m.return,
```
usergenic: (I look forward to upgrading analyzer to clang-format: 1.0.49 -- fixes this)

src/polymer/polymer-element.ts
```diff
@@ -468,11 +480,12 @@ function applySuperClass(
...
 }
 
-function applyMixins(
+export function applyMixins(
     element: PolymerElement,
     scannedElement: ScannedElement,
     document: Document) {
-  for (const scannedMixinReference of scannedElement.mixins) {
+  for (let i = scannedElement.mixins.length - 1; i >= 0; i--) {
+    const scannedMixinReference = scannedElement.mixins[i];
```
usergenic:
```ts
for (const scannedMixinReference of scannedElement.mixins.reverse())
```
How bow dat?

rictic: sadly, Array#reverse mutates the array and ScannedFeatures must be immutable

src/polymer/polymer-element.ts
```diff
@@ -262,9 +262,9 @@ function resolveElement(
     scannedElement: ScannedPolymerElement, document: Document): PolymerElement {
   const element = new PolymerElement();
   element.privacy = scannedElement.privacy;
-  applySuperClass(element, scannedElement, document);
-  applyMixins(element, scannedElement, document);
   applySelf(element, scannedElement, document);
+  applyMixins(element, scannedElement, document);
+  applySuperClass(element, scannedElement, document);
```
justinfagnani: Why are you reversing the order?

src/polymer/polymer-element.ts
```diff
@@ -334,52 +334,64 @@ function inheritFrom(element: PolymerElement, superElement: PolymerExtension) {
...
+ */
+function _inheritVals<P extends PropLike>(
+    existing: P[],
+    superVals: P[],
+    superName: string | undefined,
+    warnings: Warning[]) {
+  const existingVals = new Map(existing.map((e) => [e.name, e] as [string, P]));
+  for (const superVal of superVals) {
+    if (existingVals.has(superVal.name)) {
+      if (superVal.privacy === 'private') {
```
justinfagnani: I'm not so sure about this warning. I think at this point we're inventing our own semantics around private. Who's to say you can't override a private method? You can certainly define a method with the same name as a private method in Java, and you can make a protected method public in a subclass.

src/polymer/polymer-element.ts
justinfagnani: I see, with how _inheritVals works below. I think that it's a little easier to follow by sticking to modeling the prototype chain and building a class from top-down. This also leaves room to cache prototypes. The only performance hit is due to overwriting overloaded members, but I think that's going to be a very small percentage of members.

src/polymer/polymer-element.ts
```diff
@@ -334,52 +334,64 @@ function inheritFrom(element: PolymerElement, superElement: PolymerExtension) {
...
+            code: 'overriding-private',
+            message: `Overriding private member '${superVal.name}' ` +
+                `on ${superName || 'parent'}`,
+            sourceRange: existingVal.sourceRange,
+            severity: Severity.WARNING
+          });
+        }
       }
+      /**
+       * TODO(rictic): if superVal.privacy is protected, does that require
```
justinfagnani: It shouldn't. It doesn't in languages that actually have visibility, and I'm wary of creating too many rules around this since JS doesn't have any to begin with.

src/polymer/polymer-element.ts
rictic: I think there are three arguments in favor of this approach, in descending order of desirability:

With the current implementation of properties and friends as simple arrays, the implementation of overriding is tricky to do in a way that isn't O(N^2). Doing it this way is simple to read, simple to write, and IMO simple to understand once you accept that we go in leaf-down order rather than root-up. I think we're still in a good position to take advantage of caching of intermediate values, as we're doing here by mixing mixins into mixins.

Another issue is that I'd like to move as much of this code as possible up to Element and even to Class (once that lands), but if we go from top of the prototype chain down then that leaves us in a somewhat awkward position with behaviors. Maybe with the unification of behaviors and mixins in #630 we could do something there, but does seem a bit tricky.

And a yet smaller issue, the ordering of properties in this implementation is more legible, as local properties come first, in declaration order, followed by inherited properties. This is minor, and we could modify the code in generateAnalysis to do this sorting, the editor service does it.

src/polymer/polymer-element.ts
rictic: Java is different than Javascript though because it does name resolution at compile time. When you declare a new field with the same name as a private field in Java they're totally separate. In Javascript code in the superclass that was expecting its private field will suddenly start getting the subclass's field, which if done on accident is a recipe for super weird bugs, pun intended.

src/polymer/polymer-element.ts
rictic: Yeah, I don't have a strong intuition here, just thought I'd leave the question here.

src/polymer/polymer-element-mixin.ts
rictic: IMO it's a bit awkward, but for classes with tons of members I don't know of a better option that preserves type safety.

src/polymer/polymer-element.ts
justinfagnani: Can you explain how applying inheritance in order would be O(n^2)?

I disagree that applying in the order is easy to understand, it's backwards from how the language we're modeling does it, so I have to think more about whether it gives the same result.

Behaviors shouldn't be a problem, since Polymer 1 doesn't support inheritance.

rictic: Basically the naive method would be to walk the array for each inherited property to see if it should overwrite any entries in the array. To do O(N) we'd need to maintain an additional side-datastructure for doing this quickly. If we only append and overwrite to the array this could just be a Map<string, number>, so maybe this won't be so bad.

rictic: I'll give it a shot.

justinfagnani: We could also change each members array to a Map

src/polymer/polymer-element.ts
justinfagnani: I had a different response here where I was arguing both sides, but ended up convincing myself that this warning should be on by default :)

Basically a private field ideally shouldn't induce a warning on a subclass since it shouldn't be visible, but JavaScript's not ideal, and we should warn when this situation does occur to avoid mistakes. Someone can turn it off it necessary.

src/polymer/polymer-element.ts
justinfagnani: I think on this one, the answer is no. Protected means you do expect subclasses to see a member. We could inherit visibility by default though.

src/polymer/polymer-element.ts
rictic: Yeah, I think that's the best solution. I've got enough big changes swinging around right now though. I'll land these three and then take a look at converting them to maps. Tons of usages of these right now involve trying to awkwardly treat these arrays as maps.

src/polymer/polymer-element.ts
rictic: I guess this is the argument for symbol-named private members.

src/polymer/polymer-element.ts
```diff
@@ -334,52 +326,72 @@ function inheritFrom(element: PolymerElement, superElement: PolymerExtension) {
...
+ *
+ * TODO(rictic): unify this with inheritValues in this file.
+ */
+function _overwriteInherited<P extends PropertyLike>(
+    existing: Array<P>,
+    overriding: P[],
+    overridingName: string | undefined,
+    warnings: Warning[],
+    ignorePrivate = true) {
+  const existingIdxByName =
```
justinfagnani: what's `Idx` stand for?

src/polymer/polymer-element.ts
```diff
@@ -334,52 +326,72 @@ function inheritFrom(element: PolymerElement, superElement: PolymerExtension) {
...
+          });
+        }
       }
+      /**
+       * TODO(rictic): if existingVal.privacy is protected, newVal should be
+       *    protected unless an explicit privacy was specified.
+       *    https://github.com/Polymer/polymer-analyzer/issues/631
+       */
+
+      existing[existingIdxByName.get(overridingVal.name)!] = newVal;
```
justinfagnani: style nit: pull out index expression into a named variable

src/polymer/polymer-element.ts
```diff
@@ -334,52 +326,72 @@ function inheritFrom(element: PolymerElement, superElement: PolymerExtension) {
...
+ * Mutates `existing`
+ *
+ * TODO(rictic): unify this with inheritValues in this file.
+ */
+function _overwriteInherited<P extends PropertyLike>(
+    existing: Array<P>,
+    overriding: P[],
+    overridingName: string | undefined,
+    warnings: Warning[],
+    ignorePrivate = true) {
```
justinfagnani: document ignorePrivate

src/polymer/polymer-element.ts
```diff
@@ -334,52 +326,72 @@ function inheritFrom(element: PolymerElement, superElement: PolymerExtension) {
...
+      /**
+       * TODO(rictic): if existingVal.privacy is protected, newVal should be
+       *    protected unless an explicit privacy was specified.
+       *    https://github.com/Polymer/polymer-analyzer/issues/631
+       */
+
+      existing[existingIdxByName.get(overridingVal.name)!] = newVal;
+      continue;
+    }
+    if (ignorePrivate && overridingVal.privacy === 'private') {
```
justinfagnani: explain why

rictic:
All issues addressed. One complexity that came up is that the issue with private members overriding each other might only actually show up at mixing time, and thus there wouldn't be a super obvious place to put the source range. That complicated the API a bit, though once this is a method on the class I think it will get more reasonable again

justinfagnani:
@rictic the problem with overriding private and mixins is a specific instance of a general problem with checking mixin applications. The error is on the _application_ which only exists at the `@mixes` annotation (for now). This is ok, we'll just need to use that as the source range (once we can get source ranges for jsdocs).

rictic:
+1 to that

src/polymer/polymer-element.ts
```diff
@@ -331,55 +323,104 @@ function resolveElement(
...
+ *   applyingSelf is true
+ * @param . warnings Place to put generated warnings.
+ * @param . sourceRange A source range to use for warnings when the inheritance
+ *   goes wrong but there's no more specific source range.
+ * @param . applyingSelf True on the last call to this method, when we're
+ *   applying the class's own local members.
+ */
+export function _overwriteInherited<P extends PropertyLike>(
+    existing: Array<P>,
+    overriding: P[],
```
justinfagnani: why the different syntaxes for array types?

rictic: Made sense at the time, makes no sense now, fixed.

PR #630, opened 2017-04-18
Mark many fields readonly/immutable. Simplify Element class construction.

rictic:
Builds on top of #628 

Mark many fields readonly/immutable. Simplify element construction.

This is merely a descriptive change of how we already think of these fields.

More fields are still logically readonly/immutable, but we've got code that
initializes them in various places in a gradual way.

There is one functionality change in this PR, and that's that we treat behaviors
more like elements and more like mixins. We now mix behaviors into behaviors
and apply the same inheritance logic as mixins to behaviors.

This also makes element construction much more sane, by centralizing and documenting
how we deal with inheritance.

### DIFF
--- a/src/generate-analysis.ts
+++ b/src/generate-analysis.ts
@@ -16,7 +16,7 @@ import * as fs from 'fs';
 import * as jsonschema from 'jsonschema';
 import * as pathLib from 'path';
 
-import {Analysis, Attribute, Element, ElementLike, ElementMixin, Event, Function, Method, Namespace, Property, SourceRange} from './analysis-format';
+import {Analysis, Attribute, Element, ElementLike, ElementMixin, Event, Function, Method, Namespace, Parameter, Property, SourceRange} from './analysis-format';
 import {Function as ResolvedFunction} from './javascript/function';
 import {Namespace as ResolvedNamespace} from './javascript/namespace';
 import {Analysis as AnalysisResult} from './model/analysis';
@@ -271,9 +271,9 @@ function serializePolymerBehaviorAsElementMixin(
 
 function serializeElementLike(
     elementOrMixin: ElementOrMixin, packagePath: string): ElementLike {
-  const path = elementOrMixin.sourceRange.file;
+  const path = elementOrMixin.sourceRange!.file;
   const packageRelativePath =
-      pathLib.relative(packagePath, elementOrMixin.sourceRange.file);
+      pathLib.relative(packagePath, elementOrMixin.sourceRange!.file);
 
   const attributes = elementOrMixin.attributes.map(
       (a) => serializeAttribute(elementOrMixin, path, a));
@@ -361,7 +361,16 @@ function serializeMethod(
     metadata: resolvedElement.emitMethodMetadata(resolvedMethod),
   };
   if (resolvedMethod.params) {
-    method.params = resolvedMethod.params;
+    method.params = resolvedMethod.params.map(({name, description, type}) => {
+      const param: Parameter = {name: name};
+      if (description) {
+        param.description = description;
+      }
+      if (type) {
+        param.type = type;
+      }
+      return param;
+    });
   }
   if (resolvedMethod.return ) {
     method.return = resolvedMethod.return;
--- a/src/model/document.ts
+++ b/src/model/document.ts
@@ -18,7 +18,7 @@ import {ParsedDocument} from '../parser/document';
 
 import {Analysis} from './analysis';
 import {Feature, ScannedFeature} from './feature';
-import {ImmutableSet} from './immutable';
+import {ImmutableSet, unsafeAsMutable} from './immutable';
 import {Import} from './import';
 import {DocumentQuery as Query, DocumentQueryWithKind as QueryWithKind, FeatureKind, FeatureKindMap, Queryable} from './queryable';
 import {isResolvable} from './resolvable';
@@ -131,9 +131,9 @@ export class Document implements Feature, Queryable {
     this.languageAnalysis = languageAnalysis;
 
     if (!base.isInline) {
-      (this.identifiers as Set<string>).add(this.url);
+      unsafeAsMutable(this.identifiers).add(this.url);
     }
-    (this.kinds as Set<string>).add(`${this.parsedDocument.type}-document`);
+    unsafeAsMutable(this.kinds).add(`${this.parsedDocument.type}-document`);
     this.warnings = Array.from(base.warnings);
   }
 
--- a/src/model/document.ts
+++ b/src/model/document.ts
@@ -18,6 +18,7 @@ import {ParsedDocument} from '../parser/document';
 
 import {Analysis} from './analysis';
 import {Feature, ScannedFeature} from './feature';
+import {ImmutableSet} from './immutable';
 import {Import} from './import';
 import {DocumentQuery as Query, DocumentQueryWithKind as QueryWithKind, FeatureKind, FeatureKindMap, Queryable} from './queryable';
 import {isResolvable} from './resolvable';
@@ -89,8 +90,8 @@ declare module './queryable' {
   }
 }
 export class Document implements Feature, Queryable {
-  kinds: Set<string> = new Set(['document']);
-  identifiers: Set<string> = new Set();
+  readonly kinds: ImmutableSet<string> = new Set(['document']);
+  readonly identifiers: ImmutableSet<string> = new Set();
 
   /**
    * AnalysisContext is a private type. Only internal analyzer code should touch
@@ -100,8 +101,8 @@ export class Document implements Feature, Queryable {
   warnings: Warning[];
   languageAnalysis?: any;
 
-  private _localFeatures = new Set<Feature>();
-  private _scannedDocument: ScannedDocument;
+  private readonly _localFeatures = new Set<Feature>();
+  private readonly _scannedDocument: ScannedDocument;
 
 
   /**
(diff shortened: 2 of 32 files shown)


src/model/document.ts
```diff
@@ -130,9 +131,9 @@ export class Document implements Feature, Queryable {
     this.languageAnalysis = languageAnalysis;
 
     if (!base.isInline) {
-      this.identifiers.add(this.url);
+      (this.identifiers as Set<string>).add(this.url);
```
justinfagnani: could we make a function that turns all of our immutable types into mutable types?

src/model/element-base.ts
```diff
@@ -45,13 +54,13 @@ export abstract class ScannedElementBase implements Resolvable {
...
     }
-    this.jsdoc.tags.filter((tag) => tag.tag === 'demo' && tag.name)
-        .forEach((tag) => {
-          this.demos.push({
-            desc: tag.description || undefined,
-            path: url.resolve(baseUrl, tag.name!)
-          });
-        });
+    this.demos = this.jsdoc.tags.filter((tag) => tag.tag === 'demo' && tag.name)
+                     .map((tag) => {
```
justinfagnani: `.map((tag) => ({` and remove the `return`

src/model/element-base.ts
```diff
@@ -69,23 +78,50 @@ export class Slot {
...
-  abstract kinds: Set<string>;
-  warnings: Warning[] = [];
+  readonly properties: Property[] = [];
+  readonly attributes: Attribute[] = [];
+  readonly methods: Method[] = [];
+  description: string;
+  readonly summary: string;
+  readonly demos: Demo[] = [];
+  readonly events: Event[] = [];
+  readonly sourceRange: SourceRange|undefined;
```
justinfagnani: `sourceRange?: SourceRange`?

src/model/element-base.ts
```diff
@@ -95,7 +131,180 @@ export abstract class ElementBase implements Feature {
...
    * members override those before it in the prototype chain.
    */
-  mixins: Reference[] = [];
+  readonly mixins: ImmutableArray<Reference>;
+
+  constructor(init: ElementBaseInit, document: Document) {
+    const {
+      description,
+      summary,
+      demos,
```
justinfagnani: `demos = []`?

src/model/element-base.ts
```diff
@@ -95,7 +131,180 @@ export abstract class ElementBase implements Feature {
...
+      properties,
+      attributes,
+      methods,
+      astNode,
+      warnings,
+      slots,
+      privacy,
+      jsdoc,
+      superClass,
+      mixins,
```
justinfagnani: `mixins = []`

src/model/element-base.ts
```diff
@@ -95,7 +131,180 @@ export abstract class ElementBase implements Feature {
    * single list. A mixin can be applied more than once, each time its
    * members override those before it in the prototype chain.
    */
-  mixins: Reference[] = [];
+  readonly mixins: ImmutableArray<Reference>;
+
+  constructor(init: ElementBaseInit, document: Document) {
+    const {
```
justinfagnani: note, you can put this in the parameter list, but the way you have it now may be more legible

src/model/element-base.ts
```diff
@@ -95,7 +131,180 @@ export abstract class ElementBase implements Feature {
...
+    this.warnings = warnings ? Array.from(warnings) : [];
+    this.slots = slots ? Array.from(slots) : [];
+    this.privacy = privacy;
+    this.jsdoc = jsdoc;
+    this.superClass = superClass ? superClass.resolve(document) : undefined;
+    this.mixins = (mixins || []).map((m) => m.resolve(document));
+
+    const prototypeChain = this._getPrototypeChain(document, init);
+    for (const superClass of prototypeChain) {
+      this.inheritFrom(superClass);
```
justinfagnani: we only need to inherit from the direct superclass. It should already have everything it inherits.

src/model/element-base.ts
```diff
@@ -95,7 +131,180 @@ export abstract class ElementBase implements Feature {
...
+    this.sourceRange = sourceRange;
+    this.astNode = astNode;
+    this.warnings = warnings ? Array.from(warnings) : [];
+    this.slots = slots ? Array.from(slots) : [];
+    this.privacy = privacy;
+    this.jsdoc = jsdoc;
+    this.superClass = superClass ? superClass.resolve(document) : undefined;
+    this.mixins = (mixins || []).map((m) => m.resolve(document));
+
+    const prototypeChain = this._getPrototypeChain(document, init);
```
justinfagnani: made this after the comment below: this is a little of a confusing name, because you're not getting the whole prototype chain, you're combining the superclasses and mixins. It might be more obvious to just inline this function, or move all inheritance into this function under a different name

justinfagnani: Also, I wouldn't pass the init object. It makes me wonder if you've missed any of the default setup you do in the property initialization above.

src/model/element-base.ts
```diff
@@ -95,7 +131,180 @@ export abstract class ElementBase implements Feature {
...
+    }
+
+    this._overwriteInherited(
+        this.properties, properties || [], undefined, true);
+    this._overwriteInherited(
+        this.attributes, attributes || [], undefined, true);
+    this._overwriteInherited(this.methods, methods || [], undefined, true);
+    this._overwriteInherited(this.events, events || [], undefined, true);
+
+    for (const method of this.methods) {
```
justinfagnani: you sure you want to do this for `this.methods` and not `methods`? inherited methods should already have their privacy inferred

src/model/element-base.ts
```diff
@@ -95,7 +131,180 @@ export abstract class ElementBase implements Feature {
...
+    for (const mixin of mixins) {
+      if (mixin) {
+        prototypeChain.push(mixin);
+      }
+    }
+
+    return prototypeChain;
+  }
+
+  protected _resolveReferenceToSuperClass(
```
justinfagnani: should this be a method on `Reference`?

src/model/document.ts
rictic: Added function named `unsafeAsMutable`

src/model/element-base.ts
rictic: eh, ok. IMO multi-line arrow functions should use blocks

saving a couple lines isn't worth the added difficulty in reading

src/model/element-base.ts
rictic: I feel like `foo?: bar` is often worse than `foo: bar|undefined` in these sorts of cases, because we really want to be sure that we haven't forgotten about the field. It also means that we'll have more consistent object shapes. 

For class members though it's currently exactly the same, but in the future we may be able to get either typescript or tslint to warn about unassigned readonly properties.

src/model/element-base.ts
rictic: +1, done

src/model/element-base.ts
rictic: done, here and with other fields

src/model/element-base.ts
rictic: Good point, this name is misleading. Changed to `_getSuperclassAndMixins` and improved the docs a bit. I don't want to inline it because I override it in polymer element.

I agree that it's awkward to pass in `init`, I really didn't want to. It seems to be necessary for PolymerElement to have access to the behaviors though, and we need to mix in behaviors before we apply local members.

src/model/element-base.ts
rictic: clarified

src/model/element-base.ts
rictic: we can't mutate `methods`, only `this.methods`. this is awkward and should be removed entirely though, added a comment.

src/model/element-base.ts
rictic: maybe? but that would be a bit tricky to do right because the reference would need to know the kind of thing that it's referencing. added a TODO to take a shot at this later.

I'm kinda thinking it should be something like `Reference<Kind>`?

src/generate-analysis.ts
```diff
@@ -361,7 +361,16 @@ function serializeMethod(
     metadata: resolvedElement.emitMethodMetadata(resolvedMethod),
   };
   if (resolvedMethod.params) {
-    method.params = resolvedMethod.params;
+    method.params = resolvedMethod.params.map(({name, description, type}) => {
```
justinfagnani: can this be just:

```js
method.params = resolvedMethod.params.map(({name, description, type}) => ({name, description, type});
```

src/model/element-base.ts
```diff
@@ -45,13 +54,11 @@ export abstract class ScannedElementBase implements Resolvable {
...
-    this.jsdoc.tags.filter((tag) => tag.tag === 'demo' && tag.name)
-        .forEach((tag) => {
-          this.demos.push({
-            desc: tag.description || undefined,
-            path: url.resolve(baseUrl, tag.name!)
-          });
-        });
+    this.demos = this.jsdoc.tags.filter((tag) => tag.tag === 'demo' && tag.name)
+                     .map((tag) => ({
+                            desc: tag.description || undefined,
```
justinfagnani: in what cases do we want to convert falsey descriptions to undefined?

src/polymer/docs.ts
```diff
@@ -99,18 +99,19 @@ export function annotateEvent(annotation: jsdoc.Annotation): ScannedEvent {
...
     warnings: [],
+    params: []
   };
 
   const tags = (annotation && annotation.tags || []);
   // process @params
-  scannedEvent.params =
-      tags.filter((tag) => tag.tag === 'param').map(function(param) {
+  scannedEvent.params.push(
+      ...tags.filter((tag) => tag.tag === 'param').map(function(param) {
```
justinfagnani: use arrow function

src/model/element-base.ts
rictic: In those cases where they're nulls

src/polymer/docs.ts
rictic: done, but leaving as a block

src/generate-analysis.ts
rictic: +1, converted to just an Array.from at that point

Issue #626, opened 2017-04-17
Support named demos


rictic:
e.g. https://github.com/PolymerElements/app-layout/blob/master/app-drawer/app-drawer.html#L50

We should also do a spot check of the analyzer against app-layout afterwards to be sure we're getting the demos right.

usergenic:
@samuelli @rictic Other than the fact that demos weren't coming in from `@demo` annotations in html comments, does this bug mean you want to change the `AnalysisFormat` to emit the `demos` as something other than `string[]`, i.e. `{path: string, description: string}[]`?

samuelli:
Yep. This is to match hydrolysis behaviour so demos can have nicer titles.

samuelli:
Is it possible get a release after #633?

rictic:
@samuelli Yeah definitely, and I'm reviewing it now. Sorry, yesterday was consumed by google3 stuff.

samuelli:
I'd like to push out our analyzer support the next couple of days and this is our last blocking issue. Any update @usergenic?

usergenic:
@sameulli I have a new PR out for this feature as of this morning https://github.com/Polymer/polymer-analyzer/pull/658

Ping reviewers at your discretion.

samuelli:
Awesome, thanks!

usergenic:
#658 has been merged into master, @samuelli so closing this issue.  We'll hopefully release a new cut of Analyzer supersoon!

PR #633, opened 2017-04-19
Make demo annotations work this time

usergenic:
- Fixed issue where jsdocs syntax in adjancent HTML comments were not parsed for polymer elements.

### DIFF
--- a/src/analysis-format.ts
+++ b/src/analysis-format.ts
@@ -140,7 +140,7 @@ export interface ElementLike extends Feature {
    *
    * e.g. `['demos/index.html', 'demos/extended.html']`
    */
-  demos: string[];
+  demos: Demo[];
 
   /** Names of mixines applied to this element.  */
   mixins?: string[];
@@ -360,4 +360,12 @@ export interface Slot extends Feature {
   description: string;
 
   // Something about fallback perhaps?
+}
+
+export interface Demo extends Feature {
+  /** A markdown description of the demo. */
+  description?: string;
+
+  /** Relative URL of the demo. */
+  url: string;
 }
--- a/src/generate-analysis.ts
+++ b/src/generate-analysis.ts
@@ -16,7 +16,7 @@ import * as fs from 'fs';
 import * as jsonschema from 'jsonschema';
 import * as pathLib from 'path';
 
-import {Analysis, Attribute, Demo, Element, ElementLike, ElementMixin, Event, Function, Method, Namespace, Property, SourceRange} from './analysis-format';
+import {Analysis, Attribute, Element, ElementLike, ElementMixin, Event, Function, Method, Namespace, Property, SourceRange} from './analysis-format';
 import {Function as ResolvedFunction} from './javascript/function';
 import {Namespace as ResolvedNamespace} from './javascript/namespace';
 import {Analysis as AnalysisResult} from './model/analysis';
--- a/src/generate-analysis.ts
+++ b/src/generate-analysis.ts
@@ -295,13 +295,8 @@ function serializeElementLike(
       cssVariables: [],
       selectors: [],
     },
-    demos: (elementOrMixin.demos || []).map((demo) => {
-      const json: Demo = {url: demo.path};
-      if (demo.desc) {
-        json.description = demo.desc;
-      }
-      return json;
-    }),
+    demos: (elementOrMixin.demos ||
+            []).map(({path, desc}) => ({url: path, description: desc || ''})),
     slots: elementOrMixin.slots.map((s) => {
       return {description: '', name: s.name, range: s.range};
     }),
--- a/src/generate-analysis.ts
+++ b/src/generate-analysis.ts
@@ -16,7 +16,7 @@ import * as fs from 'fs';
 import * as jsonschema from 'jsonschema';
 import * as pathLib from 'path';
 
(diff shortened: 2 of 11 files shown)


usergenic:
Depending on answer to https://github.com/Polymer/polymer-analyzer/issues/626#issuecomment-295456145 this PR may be prereq to or fix #626.  @rictic PTAL

src/polymer/polymer-element.ts
```diff
@@ -183,7 +183,15 @@ export class ScannedPolymerElement extends ScannedElement implements
...
+  applyDescriptionAsJsdoc() {
+    this.jsdoc = jsdoc.parseJsdoc(this.description);
+    this.description = this.jsdoc.description;
+  }
+
   resolve(document: Document): PolymerElement {
+    if (!this.jsdoc) {
+      this.applyDescriptionAsJsdoc();
+    }
     this.applyJsdocDemoTags(document.url);
```
rictic: I think all of this should go into the body of `resolveElement`, as calling `resolve` shouldn't mutate the `ScannedPolymerElement`. We may call `resolve` many times on the same scanned element, and we don't want it to change every time.

src/analysis-format.ts
```diff
@@ -140,7 +140,7 @@ export interface ElementLike extends Feature {
    *
    * e.g. `['demos/index.html', 'demos/extended.html']`
    */
-  demos: string[];
+  demos: Demo[];
```
justinfagnani: I wonder if we should move demos to metadata? cc @rictic

src/generate-analysis.ts
```diff
@@ -295,7 +295,13 @@ function serializeElementLike(
       cssVariables: [],
       selectors: [],
     },
-    demos: (elementOrMixin.demos || []).map((d) => d.path),
+    demos: (elementOrMixin.demos || []).map((demo) => {
+      const json: Demo = {url: demo.path};
```
justinfagnani: I think you can just return a literal:

```
=> ({
  url: demo.path,
  description: demo.desc,
});
```

justinfagnani: Or is description sometimes `''` rather than `undefined`?

src/polymer/polymer-element.ts
justinfagnani: shouldn't this all go in the constructor? I'd like to not mutate the scanned element even in resolve()

src/polymer/polymer-element.ts
```diff
@@ -400,6 +408,10 @@ function applySelf(
   scannedElement.events.forEach((o) => element.events.push(o));
   element.extends = scannedElement.extends;
   element.jsdoc = scannedElement.jsdoc;
+  if (!element.jsdoc && element.description) {
+    element.jsdoc = jsdoc.parseJsdoc(element.description);
```
justinfagnani: Can you comment the intention behind this block a little?

src/polymer/polymer-element.ts
usergenic: In my prior merged PR which added the `applyJsdocDemoTags()` method I had not realized that `resolve()` was intended to be idempotent and also not modify the scanned element.  Am working on update to address that error as well.

src/generate-analysis.ts
usergenic: I did this to prefer excluding the description key in the object when value was undefined.

usergenic: Okay updated to:
```typescript
demos: (elementOrMixin.demos ||
            []).map(({path, desc}) => ({url: path, description: desc || ''})),
```

This defaults undefined description to an empty string.

src/polymer/polymer-element.ts
usergenic: I've moved the `applyJsdocDemoTags` method to `ElementBase` instead of `ScannedElementBase` now.

src/polymer/polymer-element.ts
usergenic: This is no longer needed because the jsdoc demo tag processing takes place at the end of resolving elements now.

usergenic: Added comment.

usergenic:
@justinfagnani PTAL.  I no longer perform mutations in `ScannedElement*`.`resolve()` and added comments as requested.

src/polymer/polymer-element.ts
```diff
@@ -184,8 +184,8 @@ export class ScannedPolymerElement extends ScannedElement implements
   }
 
   resolve(document: Document): PolymerElement {
-    this.applyJsdocDemoTags(document.url);
-    return resolveElement(this, document);
+    const element = new PolymerElement();
```
justinfagnani: I'd new up the element in `resolveElement`

usergenic: Looks like resolveElement is gone now in master.  There was apparently a big refactoring since this PR opened.  Time for a re-do.

justinfagnani:
@usergenic can we get a rebase?

src/analysis-format.ts
rictic: I think demos should be as standard as descriptions. They're just named links, and they're really useful for all UI-related stuff.

usergenic:
too much changed to do a rebase.  closing pull request and will create a new one.

PR #635, opened 2017-04-20
Mute errors in parsing class definitions

samuelli:

Testing this out for a hydrolysis 1.25.0 release

### DIFF
--- a/src/analyzer.ts
+++ b/src/analyzer.ts
@@ -317,7 +317,8 @@ export class Analyzer {
     for (const element of pseudoElements) {
       element.contentHref = href;
       this.elements.push(element);
-      this.elementsByTagName[element.is] = element;
+      if (element.is)
+        this.elementsByTagName[element.is] = element;
     }
     metadataLoaded = metadataLoaded.then(function(metadata){
       var metadataEntry: DocumentDescriptor = {
@@ -419,7 +420,7 @@ export class Analyzer {
           this.elements.push(element);
           if (element.is in this.elementsByTagName) {
             console.warn('Ignoring duplicate element definition: ' + element.is);
-          } else {
+          } else if (element.is) {
             this.elementsByTagName[element.is] = element;
           }
         });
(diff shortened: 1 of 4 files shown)


rictic:
Do you have more detail with this change? What was it doing before and how does this PR change that?

samuelli:
When using NPM hydrolysis 1.24.1, an exception occurred when analysing repos that targeted Polymer 2.0, including hybrid elements which default to 2.0.

This change simply prevents those exceptions from occurring and allows hybrid elements to be processed by hydrolysis. Justin thinks its where some of the Polymer definitions use getter syntax for `is`, instead of a string literal, which is what hydrolysis expects.

src/analyzer.ts
```diff
@@ -317,7 +317,8 @@ export class Analyzer {
     for (const element of pseudoElements) {
       element.contentHref = href;
       this.elements.push(element);
-      this.elementsByTagName[element.is] = element;
+      if (element.is)
```
justinfagnani: add braces around then clause, here and below

justinfagnani:
@rictic basically the class support in Hydrolysis was getting tripped up by 2.0 classes in core when analyzing hybrid elements.

src/analyzer.ts
samuelli: Done.

samuelli:
> @samuelli can you add a commit to bump the version to 1.25 so we can tag?

Done. I don't have to anything but `npm run build`, right?

justinfagnani:
@samuelli pulling in those changes should be fine. Did you test in wc.org against master with these changes, or just 1.24.1 with them?

justinfagnani:
I'm restarting the tests - it was a few browsers. Since we don't know they current status of the hydrolysis-1.x on Travis, it's hard to say if this broke anything, but I seriously doubt it.

samuelli:
> @samuelli pulling in those changes should be fine. Did you test in wc.org against master with these changes, or just 1.24.1 with them?
> 
> edit: sorry, not master, but the hydrolysis-1.x branch

Tested wc.org locally with hydrolysis-1.x + changes. Hard to test extensively though. Once there's a tagged version I'll be able to deploy it to staging and see more extensive results there.

justinfagnani:
Ok, some browsers still failing, but they probably already were. Merging.

PR #637, opened 2017-04-20
Improve privacy inference

### LINKED ISSUE #586: Better criteria for public/private/protected methods
From https://github.com/Polymer/polymer-analyzer/blob/master/src/polymer/js-utils.ts#L113:

```js
if (explicitPrivacy) {
    return explicitPrivacy;
  } else if (specificName.startsWith('__')) {
    return 'private';
  } else if (specificName.startsWith('_')) {
    return 'protected';
  } else if (specificName.endsWith('_')) {
    return 'private';
  } else if (configurationProperties.has(specificName)) {
    return 'protected';
  } else {
    if (privateUnlessDocumented) {
      // Some members, like methods or properties on classes are private by
      // default unless they have documentation.
      const hasDocs = !!annotation && !jsdoc.isAnnotationEmpty(annotation);
      return hasDocs ? 'public' : 'private';
    } else {
      // Other members, like entries in the Polymer `properties` block are
      // public unless there are clear signals otherwise.
      return 'public';
    }
  }
```
The whole idea of 'private unless documented' doesn't make a lot of sense to me. We're using two different sources of information to infer privacy.

1. Extract privacy from jsdocs
2. Infer privacy from method names (_protected, __private etc.)

Explicit privacy annotation via jsdoc trumps method naming, which makes sense, but why are we assuming undocumented methods are private (I know this is optional, but it's what we're using for scanning Polymer elements, for example). And why does this rule only apply to 'regular' named methods and not to pre/suffixed ones? this has the weird effect that an undocumented, 'regular' method has a higher privacy than a prefixed one:
```
myMethod() {} // --> private,
_otherMethod() {} // --> protected
```
This just seems wildly inconsistent to me. Another weird thing is that add __any__ comment to a method makes it public.
```
// TODO: fixme
myMethod() {} // --> public
```
I think if we infer privacy from method names, we should be consistent about it. I.e. in the **absence** of an explicit jsdoc privacy annotation:

* `regularMethod() {}`: **public**
* `_prefixed() {}`: **protected**
* `__doublePrefixed() {}`, `postFixed_()`: **private**

Optionally we can opt out of inferring privacy from method names altogether and instead assume a default privacy, but then we should do this consistently for all kinds of methods, not just for 'regular' ones.

Here is my proposed patch:

```js
export function getOrInferPrivacy(
    name: string,
    annotation: jsdoc.Annotation|undefined,
    inferFromName = true,
    defaultPrivacy: Privacy = 'private'): Privacy {
  const explicitPrivacy = jsdoc.getPrivacy(annotation);
  const specificName = name.slice(name.lastIndexOf('.') + 1);

  if (explicitPrivacy) {
    return explicitPrivacy;
  } else if (inferFromName) {
    if (specificName.startsWith('__') || specificName.endsWith('_')) {
      return 'private';
    } else if (specificName.startsWith('_') || configurationProperties.has(specificName)) {
      return 'protected';
    } else {
      return 'public';
    }
  } else {
    return defaultPrivacy;
  }
}
```

Thoughts? I'm happy to put up a PR if you guys agree with the changes.

MaKleSoft:
Updates the rules for inferring the privacy of javascript entities as described in https://github.com/Polymer/polymer-analyzer/issues/586. Specifically, undocumented methods without an underscore pre- or postfix are now considered public (were considered private before).

Fixes #586

### DIFF
--- a/src/polymer/js-utils.ts
+++ b/src/polymer/js-utils.ts
@@ -59,7 +59,7 @@ export function toScannedPolymerProperty(
     astNode: node,
     isConfiguration: configurationProperties.has(name),
     jsdoc: parsedJsdoc,
-    privacy: getOrInferPrivacy(name, parsedJsdoc, false)
+    privacy: getOrInferPrivacy(name, parsedJsdoc, true)
   };
 
   return result;
@@ -146,30 +146,26 @@ export function toScannedMethod(
 export function getOrInferPrivacy(
     name: string,
     annotation: jsdoc.Annotation|undefined,
-    privateUnlessDocumented: boolean): Privacy {
+    inferFromName = true,
+    defaultPrivacy: Privacy = 'private'): Privacy {
   const explicitPrivacy = jsdoc.getPrivacy(annotation);
   const specificName = name.slice(name.lastIndexOf('.') + 1);
 
   if (explicitPrivacy) {
     return explicitPrivacy;
-  } else if (specificName.startsWith('__')) {
-    return 'private';
-  } else if (specificName.startsWith('_')) {
-    return 'protected';
-  } else if (specificName.endsWith('_')) {
-    return 'private';
-  } else if (configurationProperties.has(specificName)) {
-    return 'protected';
-  } else {
-    if (privateUnlessDocumented) {
-      // Some members, like methods or properties on classes are private by
-      // default unless they have documentation.
-      const hasDocs = !!annotation && !jsdoc.isAnnotationEmpty(annotation);
-      return hasDocs ? 'public' : 'private';
+  } else if (inferFromName) {
+    if (specificName.startsWith('__')) {
+      return 'private';
+    } else if (specificName.startsWith('_')) {
+      return 'protected';
+    } else if (specificName.endsWith('_')) {
+      return 'private';
+    } else if (configurationProperties.has(specificName)) {
+      return 'protected';
     } else {
-      // Other members, like entries in the Polymer `properties` block are
-      // public unless there are clear signals otherwise.
       return 'public';
     }
+  } else {
+    return defaultPrivacy;
   }
 }
--- a/src/polymer/polymer2-element-scanner.ts
+++ b/src/polymer/polymer2-element-scanner.ts
@@ -162,7 +162,7 @@ export class Polymer2ElementScanner implements JavaScriptScanner {
         methods: getMethods(node, document),
         superClass: this._getExtends(node, docs, warnings, document),
         mixins: jsdoc.getMixins(document, node, docs, warnings),
-        privacy: getOrInferPrivacy(className || '', docs, false),
+        privacy: getOrInferPrivacy(className || '', docs, false, 'private'),
         observers: observers,
         jsdoc: docs,
       });
@@ -188,7 +188,7 @@ export class Polymer2ElementScanner implements JavaScriptScanner {
         description: (docs.description || '').trim(),
         superClass: this._getExtends(node, docs, warnings, document),
         mixins: jsdoc.getMixins(document, node, docs, warnings),
-        privacy: getOrInferPrivacy(className || '', docs, false),
+        privacy: getOrInferPrivacy(className || '', docs, false, 'private'),
         events: [],
         properties: [],
         methods: [],
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
    Unreleased section, uncommenting the header as necessary.
 -->
 
-<!--## Unreleased-->
+## Unreleased
+
+* [BREAKING] Improve rules for infering method privacy. Specifically, undocumented methods without an underscore pre- or postfix are now considered public (were considered private before) 
 
 ## [2.0.0-alpha.38] - 2017-04-13
 
--- a/src/polymer/behavior-scanner.ts
+++ b/src/polymer/behavior-scanner.ts
@@ -159,7 +159,7 @@ class BehaviorVisitor implements Visitor {
       description: comment,
       events: esutil.getEventComments(node),
       sourceRange: this.document.sourceRangeForNode(node),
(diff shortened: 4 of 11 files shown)


MaKleSoft:
Btw, I'm getting a couple of warnings from the typescript compiler (see below). The tests are passing though.

```sh
[gulp-typescript] 2300 Duplicate identifier 'Skip'.
> file: undefined:
> [12]                                                              VisitorOption);
[gulp-typescript] 2300 Duplicate identifier 'Break'.
> file: undefined:
> [12]                                                              VisitorOption);
[gulp-typescript] 2300 Duplicate identifier 'Remove'.
> file: undefined:
> [12]                                                              VisitorOption);
[gulp-typescript] 2403 Subsequent variable declarations must have the same type.  Variable 'enter' must be of type '((node: Node, parentNode: Node) => void | VisitorOption) | undefined', but here has type '((node: any, parentNode: any) => any) | undefined'.
> file: undefined:
> [8] 
[gulp-typescript] 2403 Subsequent variable declarations must have the same type.  Variable 'leave' must be of type '((node: Node, parentNode: Node) => void | VisitorOption) | undefined', but here has type '((node: any, parentNode: any) => any) | undefined'.
> file: undefined:
> [9] 

[gulp-typescript] 2403 Subsequent variable declarations must have the same type.  Variable 'fallback' must be of type '"iteration" | undefined', but here has type 'string | undefined'.
> file: undefined:
> [11] 

[gulp-typescript] 2300 Duplicate identifier 'Skip'.
> file: undefined:
> [17] 

[gulp-typescript] 2432 In an enum with multiple declarations, only one declaration can omit an initializer for its first enum element.
> file: undefined:
> [17] 

[gulp-typescript] 2300 Duplicate identifier 'Break'.
> file: undefined:
> [17] 

[gulp-typescript] 2300 Duplicate identifier 'Remove'.
```

rictic:
Those are warnings that usually come up from `npm link`

src/polymer/polymer2-element-scanner.ts
```diff
@@ -188,7 +188,7 @@ export class Polymer2ElementScanner implements JavaScriptScanner {
         description: (docs.description || '').trim(),
         superClass: this._getExtends(node, docs, warnings, document),
         mixins: jsdoc.getMixins(document, node, docs, warnings),
-        privacy: getOrInferPrivacy(className || '', docs, false),
+        privacy: getOrInferPrivacy(className || '', docs, false, 'private'),
```
rictic: These should be public by default yeah?

src/polymer/js-utils.ts
```diff
@@ -146,30 +146,26 @@ export function toScannedMethod(
 export function getOrInferPrivacy(
     name: string,
     annotation: jsdoc.Annotation|undefined,
-    privateUnlessDocumented: boolean): Privacy {
+    inferFromName = true,
```
rictic: when don't we want to infer privacy from the name?

src/polymer/polymer2-element-scanner.ts
MaKleSoft: Dunno? There are no existing tests for this. I would say private by default makes more sense since quite often element classes are defined within closures and only publicly available via `document.createElement`. E.g.:

```js
(() => {

class MyElement extends Polymer.Element {
...
}

window.customElements.define(MyElement.is, MyElement);

})();
```
That's just my intuition though. I'm sure you could just as well argue for public by default.

src/polymer/js-utils.ts
MaKleSoft: Well, I don't know if we want to adopt the underscore-naming rules to all kinds of javascript entities. Does it make as much sense for element classes as it does for methods, for example? Pretty sure you'll find many people would be offended by the idea of prefixing class names with underscores.

rictic: Explicit privacy annotations always win so they could add an `@private` comment if they don't want to name them with underscores, but IMO underscores are typically communicating privacy.

src/polymer/polymer2-element-scanner.ts
rictic: Scope is separate from privacy. We don't do scope correctly now, but there's a difference IMO between scope (what you can do) and privacy (what the author of the code wants you to do).

src/polymer/js-utils.ts
justinfagnani: I'm not sure when class names would need to be private. If we're documenting classes that are not added to a namespace or registered as an element, maybe we should not do that.

src/polymer/polymer2-element-scanner.ts
justinfagnani: I think top-level members of a file should be assumed to not be visible unless they're part of a namespace, or they're defined elements. That's a good approximation of a scope for now.

MaKleSoft: > unless they're part of a namespace, or they're defined elements
Is this easy to analyse statically, though? 

So, any proposed  changes? Should we leave it at private?

src/polymer/js-utils.ts
MaKleSoft: Well, how about 'abstract' classes? Its not hard to think of scenarios where you might want to implement a common base class for a group of elements that is not meat to be instantiated or used otherwise outside a given scope. Imo your argument for 'top-level members' applies to to classes as well:

> I think top-level members of a file should be assumed to not be visible unless they're part of a namespace, or they're defined elements. That's a good approximation of a scope for now.

rictic:
As a counter-proposal, rather than having a bunch of different privacy rules, what about something simple and consistent.

- If there is an explicit privacy annotation (`@public`, `@private`, `@protected`) that wins.
- If the non-namespaced name of the feature starts with two underscores, or ends in one underscore, the feature is private.
- If the non-namespaced name of the feature starts with one underscore, the feature is protected.
- Otherwise the feature is public.

We can add some nuance later, like whether it makes sense for a class to be protected, but this seems to be to be easy to reason about, easy to understand, easy to remember, and gives the user complete control.

usergenic:
Closing PR because #655 was merged.

PR #639, opened 2017-04-21
Enable stricter checks for indexing expressions

rictic:

This appears to have fixed several actual bugs.

### DIFF
--- a/src/javascript/javascript-document.ts
+++ b/src/javascript/javascript-document.ts
@@ -31,8 +31,9 @@ export {Visitor} from './estree-visitor';
  * analysis time, and nothing happens at runtime.
  */
 const __exampleNode: Node = <any>null;
+type EstreeType = typeof __exampleNode.type;
 interface SkipRecord {
-  type: typeof __exampleNode.type;
+  type: EstreeType;
   depth: number;
 }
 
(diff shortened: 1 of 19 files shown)


src/javascript/javascript-document.ts
```diff
@@ -68,7 +69,11 @@ export class JavaScriptDocument extends ParsedDocument<Node, Visitor> {
           continue;
         }
         if (callbackName in visitor) {
-          const result: VisitResult = visitor[callbackName](node, parent);
+          // TODO(rictic): is there a maintainable way to enforce the
```
justinfagnani: I don't think this is necessary or possible. We're off the type rails with this kind of meta programming.

src/polymer/behavior-scanner.ts
```diff
@@ -27,24 +27,16 @@ import {declarationPropertyHandlers, PropertyHandlers} from './declaration-prope
 import * as docs from './docs';
 import {getOrInferPrivacy, toScannedPolymerProperty} from './js-utils';
 
-interface KeyFunc<T> {
-  (value: T): any;
-}
-
-function dedupe<T>(array: T[], keyFunc: KeyFunc<T>): T[] {
-  const bucket = {};
+function dedupe<T, K>(array: T[], keyFunc: (value: T) => K): T[] {
```
justinfagnani: seems like this should go in utils, btw

src/polymer/polymer-element-mixin.ts
```diff
@@ -116,8 +116,13 @@ export class PolymerElementMixin extends ElementMixin implements
...
   emitPropertyMetadata(property: PolymerProperty) {
-    const polymerMetadata: any = {};
-    const polymerMetadataFields = ['notify', 'observer', 'readOnly'];
+    const polymerMetadata:
+        {notify?: boolean, observer?: string, readOnly?: boolean} = {};
+    const polymerMetadataFields = [
+      'notify' as 'notify',
+      'observer' as 'observer',
+      'readOnly' as 'readOnly'
+    ];
```
justinfagnani: `]: (keyof typeof polymerMetadata)[]'`?

src/polymer/behavior-scanner.ts
rictic: true, though I hope to delete this by just turning these arrays into maps anyways

PR #642, opened 2017-04-25
Emit warnings from scanners

rictic:

After some thought, the IncompleteFeature route is kinda heavyweight, we can instead just emit warnings from scan() and attach them to the scanned document.

If approved this PR would supplant #541 and #542

### DIFF
--- a/src/javascript/class-scanner.ts
+++ b/src/javascript/class-scanner.ts
@@ -67,9 +67,8 @@ export interface ScannedAttribute extends ScannedFeature {
  */
 export class ClassScanner implements JavaScriptScanner {
   async scan(
-      document: JavaScriptDocument, visit: (visitor: Visitor) => Promise<void>):
-      Promise<Array<ScannedPolymerElement|ScannedClass|
-                    ScannedPolymerElementMixin>> {
+      document: JavaScriptDocument,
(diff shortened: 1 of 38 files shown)


src/javascript/class-scanner.ts
```diff
@@ -154,13 +153,14 @@ export class ClassScanner implements JavaScriptScanner {
...
-    // elementDefinitionFinder.warnings.concat(
-    // mixinFinder.warnings).concat(classFinder.warnings);
-
-    return scannedFeatures;
+    return {
+      features: scannedFeatures,
+      warnings: [
+        ...elementDefinitionFinder.warnings,
+        ...classFinder.warnings,
+        ...mixinFinder.warnings
```
justinfagnani: add trailing comma

src/scanning/scanner.ts
```diff
@@ -26,7 +26,7 @@ import {ParsedDocument} from '../parser/document';
  */
 export interface Scanner<D extends ParsedDocument<A, V>, A, V> {
   scan(document: D, visit: (visitor: V) => Promise<void>):
-      Promise<ScannedFeature[]>;
+      Promise<{features: ScannedFeature[], warnings?: Warning[]}>;
```
justinfagnani: immutable?

Issue #614, opened 2017-04-11
Polymer lint directive scanner


rictic:
For https://github.com/Polymer/polymer-linter/issues/58

rictic:
Requirements:

* Supports comments in HTML or Javascript
* Tracks the source range of the comment
* Tracks what lint rule names are enabled and disabled by the comment

Proposed Syntax (🚲🏘ing welcome):

```html
  <!-- polymer-lint enable: rule-name1 rule-name2 -->

  <!-- polymer-lint disable: rule-name2 rule-name3 -->
```

```javascript
// polymer-lint enable: rule-name1 rule-name2

// polymer-lint disable: rule-name1 rule-name2
```

FredKSchott:
Some feedback/suggestions from thinking about this:

#### Comma separate rule list:

This lets us leave the door open for arguments/augmenting rules in the future.

```html
<!-- polymer-lint enable: rule-name1, rule-name2 -->

hypotetical future support (total straw man):
<!-- polymer-lint enable: rule-name1 warn-only, rule-name2 -->
```

#### Support disabling of all rules for code blocks:

```js
/* polymer-lint disable */

/* polymer-lint enable */
```

#### Support disabling for entire files:

If a disable comment is included at the top of a file, the entire file should inherit those disable rules.

abdonrd:
You've probably seen it, but just in case...
[ESLint: Disabling Rules with Inline Comments](http://eslint.org/docs/user-guide/configuring#disabling-rules-with-inline-comments)

FredKSchott:
I'm trying to think of how we can support this generally without building one-off polymer-linter support into the analyzer. This kind of detection will be useful for more than just polymer-linter.

What if we included comments as features in a document? Then we could build the ability to, before running a rule check on a document, convert any `polymer-lint enable`/`polymer-lint disable` comment pairs into source ranges. This would be enough information for a rule to know whether it should report an error or not. If we're worried about this being too noisy, we could make support for them optional / not have them be returned by the default `getFeatures()`.

If we don't want to add comments as features at all, the only other way I could see this supported generally would be a way to walk `astNodes` for comments. Each language would need to encapsulate its own walker, probably within a document method.

@rictic @justinfagnani wdyt?

FredKSchott:
@abdonrd Thanks for linking, I just read through that myself. 

[...]

PR #644, opened 2017-04-25
Comment Directives & Comment Annotations

FredKSchott:
- Originally brought up in: #614
- `polymer-lint` counterpart: https://github.com/Polymer/polymer-linter/issues/58
- I recommend viewing the rendered file & leaving comments in the diff

### DIFF
--- a/design/comment-scanning.md
+++ b/design/comment-scanning.md
@@ -1,0 +1,110 @@
+# Polymer Analyzer - Comment Directives & Comment Annotations
+> Author @FredKSchott
+> Status: Draft
+> Last Update: 2017-04-25
+
+## Objective
+- To create a paradigm for formatting document metadata and instructions for analysis.
+- Immediately driving this discussion is the need to support polymer-linter directives which can enable/disable lint rules within entire documents and/or single sections of documents.
+
+## Goals
+- Allow the analyzer to surface the information contained within polymer-lint directives so that the linter can properly report & ignore warnings.
+- Create a more general definition for a directive that is easy-to-use for code authors and fits into the data model of the analyzer.
+- Create a more general definition for metadata (ex: @polymerElement) that is attached to an already-existing feature.
+- Documentation (README, docs/ entry, docs site entry, etc) that communicates these new definitions.
+
+## Non-Goals
+- Implement any general-purpose metadata/directive scanners. This document will focus solely on the syntax and the way we talk about scannable documentation going forward.
+
+
+## Background
+Currently, our analyzer is able to use documentation to detect implicit features and characteristics of features that would otherwise be missed. For example, the `@customElement` tag is used to signify that the following class definition is a custom element.
+
+However, there is another class of documentation that our analyzer currently doesn’t support. Our linter needs to be able to understand documentation that doesn’t describe a single, already existing feature. For example, the user may want to send an instruction to some analyzer-powered tool in its build chain:
+
+```html
+<!-- @registerServiceWorkerHere -->
+```
+
+Or, a user may want to describe some behavior across an entire source range, including any number of other existing features (including zero). For example:
+
+```js
+/* polymer-lint disable: undefined-elements */
+customElements.define('vanilla-element1', AnElementThatActuallyExists);
+customElements.define('vanilla-element2', IPromiseThisExists);
+customElements.define('vanilla-element3', SeriouslyLinterDontBeMad);
+/* polymer-lint enable: undefined-elements */
+```
+
+The polymer-lint enable/disable use-case is the immediate need here. But instead of building one-off support for this into the analyzer, we’d like to also think about how the analyzer should analyze all comment-based information so that future features don’t need to reinvent the wheel when it comes to comment-based analysis.
+
+
+## Design Overview
+
+This design groups all analysis-aware documentation into two supported formats: **Comment Directives** & **Comment Annotations**. Each supports a different general use of documentation, and together they cover most use cases for sending instructions and explicit information to the analyzer.
+
+### Comment Directive
+
+```
+/* polymer-lint disable: undefined-elements, rule-2 */
+/* polymer-lint enable: undefined-elements */
+/* @lazyImport '../some-url/some-file.html */
+<!-- @registerServiceWorkerHere -->
+```
+
+- A comment directive represents a new **Directive** feature type within a document
+- A new scanner should be written and used for each supported comment directive
+- A directive can describe one of three things:
+  - **Single Line Directive:** A feature that exists on a single line.
+  - **Source Range Directive Start:** A feature that exists on a source range within a document, starting at the directive’s source location.
+  - **Source Range Directive End:** A feature that exists on a source range within a document, ending at the directive’s source location.
+- A comment directive has a unique ID and may take arguments
+  - Because of the wide range of possible features & needs, specific syntax is lax
+  - Parsing is left up to each scanner
+- A comment directive must exist in its own comment block
+- A source range directive comment can describe several Directive features based on the arguments given
+
+
+### Comment Annotation
+
+```js
+/**
+ * @polymer
+ * @mixinFunction
+ * @memberof Polymer
+ */
+const TestMixin = function(superclass) {
+```
+
+- A comment annotation describes another, already-existing feature in a document
+- A comment annotation has a unique ID and may take arguments
+- Multiple comment annotations are allowed within a single code block
+  - Each requires a new line
+- Because they inform already-existing features, no new scanners are required
+  - The scanners for those features are responsible for their scanning
+- `@demo` & `@customElement` are two comment annotations that are already supported by the analyzer
+
+
+## Implementation Example: Polymer Lint
+
+Here is an overview of how the Linter would work using the two comment-types described above.
+
+For the purpose of this example, we’ll denote a Lint Directive as `IgnoreDirective([ruleName])` where ruleName is the name of a specific rule to disable. If no ruleName is provided, all lint rules will be disabled.
+
+### Scanning / Analysis
+
+1. A new pluggable scanner would be added to the analyzer for each language we care to scan for Polymer Lint directives.
+1. That scanner knows to look for two different comment directive IDs:
+  - Source Range Start Directive: “polymer-lint disable:”
+  - Source Range End Directive: “polymer-lint enable:”
+1. That scanner also knows how to parse the arguments for each as a comma-separated list of rules.
+1. For each “polymer-lint disable” directive found, create a new **IgnoreDirective** feature for each rule name provided. If no rule names are provided, create a single new **IgnoreDirective** to disable all rules. For each **IgnoreDirective** created, set the source-range starting location to the comment node.
+1. For each “polymer-lint enable” directive found, find the last previously created **IgnoreDirective** that shares its name and set the source-range ending location to the comment node. If no **IgnoreDirective** is found, return a warning.
+1. If an **IgnoreDirective** has no ending location, set it’s ending location to the end of the document.
+Return all found **IgnoreDirective** as scanned features.
+
+### Linting
+
+1. The linter gets all **IgnoreDirective** for a document before linting a given set of rules
+1. Get all **IgnoreDirective(rule)** source ranges for a given rule.
+1. For each warning returned by a rule’s `check()` method, filter out all warnings that are contained entirely within any ignore directive source ranges.
--- a/design/comment-scanning.md
+++ b/design/comment-scanning.md
@@ -137,7 +137,7 @@ Here is an overview of how the Linter would work using the two comment-types des
 
 For the purpose of this example, we’ll denote a Lint Directive as `LintDirective(command: 'enable'|'disable', rules?: string[])` where rules is a list of rules *or rule sets*. If `rules` is` undefined, all lint rules will be enabled/disabled.
 
-### Scanning / Analysis
+### In the Analyzer...
 
 1. A new pluggable scanner would be added to the analyzer for each language we care to scan for Polymer Lint directives.
 1. That scanner knows to look for two different comment directive IDs:
@@ -148,7 +148,7 @@ For the purpose of this example, we’ll denote a Lint Directive as `LintDirecti
 1. For each "enable" `LintDirective` found, create a new **`LintDirective('enable', rules)`** with the parsed rules (or undefined if none existed).
 1. Return all found `LintDirectives` as scanned features of kind 'directive'.
 
-### Linting
+### In the Linter...
 
 1. The linter gets all `LintDirectives` for a document before linting a given set of rules
 1. It could either:
--- a/design/comment-scanning.md
+++ b/design/comment-scanning.md
@@ -16,7 +16,7 @@
 
 
 ## Non-Goals
-- Implementing any "attached" comment annotation scanning. It was important to outline so that the scope of comment directives could be explained, but implementation will be left as future work.
+- Implementing any "attached" comment annotation scanning. It was important to outline so that the scope of comment directives could be explained, but implementation will be left as future work. The current `jsdoc` property on JavaScript features is good enough for the moment.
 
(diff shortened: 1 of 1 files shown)


design/comment-scanning.md
```diff
@@ -0,0 +1,110 @@
...
+ */
+const TestMixin = function(superclass) {
+```
+
+- A comment annotation describes another, already-existing feature in a document
+- A comment annotation has a unique ID and may take arguments
+- Multiple comment annotations are allowed within a single code block
+  - Each requires a new line
+- Because they inform already-existing features, no new scanners are required
+  - The scanners for those features are responsible for their scanning
```
justinfagnani: so element, mixin, etc. scanner would need to specifically call into some function to scan for attached linting directives?

design/comment-scanning.md
```diff
@@ -0,0 +1,110 @@
...
+
+```
+/* polymer-lint disable: undefined-elements, rule-2 */
+/* polymer-lint enable: undefined-elements */
+/* @lazyImport '../some-url/some-file.html */
+<!-- @registerServiceWorkerHere -->
+```
+
+- A comment directive represents a new **Directive** feature type within a document
+- A new scanner should be written and used for each supported comment directive
```
justinfagnani: No common directive scanner?

design/comment-scanning.md
```diff
@@ -0,0 +1,110 @@
...
+- A directive can describe one of three things:
+  - **Single Line Directive:** A feature that exists on a single line.
+  - **Source Range Directive Start:** A feature that exists on a source range within a document, starting at the directive’s source location.
+  - **Source Range Directive End:** A feature that exists on a source range within a document, ending at the directive’s source location.
+- A comment directive has a unique ID and may take arguments
```
justinfagnani: what's the ID for?

design/comment-scanning.md
```diff
@@ -0,0 +1,110 @@
...
+- A source range directive comment can describe several Directive features based on the arguments given
+
+
+### Comment Annotation
+
+```js
+/**
+ * @polymer
+ * @mixinFunction
+ * @memberof Polymer
```
justinfagnani: jsdoc parsing is already handled, are you proposing to replace it? why, and how does this address the use-case of linting directives?

design/comment-scanning.md
```diff
@@ -0,0 +1,110 @@
...
+
+Here is an overview of how the Linter would work using the two comment-types described above.
+
+For the purpose of this example, we’ll denote a Lint Directive as `IgnoreDirective([ruleName])` where ruleName is the name of a specific rule to disable. If no ruleName is provided, all lint rules will be disabled.
+
+### Scanning / Analysis
+
+1. A new pluggable scanner would be added to the analyzer for each language we care to scan for Polymer Lint directives.
+1. That scanner knows to look for two different comment directive IDs:
+  - Source Range Start Directive: “polymer-lint disable:”
```
justinfagnani: specifically these strings, or are these just examples?

design/comment-scanning.md
```diff
@@ -0,0 +1,110 @@
...
+- Because they inform already-existing features, no new scanners are required
+  - The scanners for those features are responsible for their scanning
+- `@demo` & `@customElement` are two comment annotations that are already supported by the analyzer
+
+
+## Implementation Example: Polymer Lint
+
+Here is an overview of how the Linter would work using the two comment-types described above.
+
+For the purpose of this example, we’ll denote a Lint Directive as `IgnoreDirective([ruleName])` where ruleName is the name of a specific rule to disable. If no ruleName is provided, all lint rules will be disabled.
```
justinfagnani: why `IgnoreDirective`? Might we not have directive that turn _on_ a rule?

design/comment-scanning.md
```diff
@@ -0,0 +1,110 @@
...
+> Last Update: 2017-04-25
+
+## Objective
+- To create a paradigm for formatting document metadata and instructions for analysis.
+- Immediately driving this discussion is the need to support polymer-linter directives which can enable/disable lint rules within entire documents and/or single sections of documents.
+
+## Goals
+- Allow the analyzer to surface the information contained within polymer-lint directives so that the linter can properly report & ignore warnings.
+- Create a more general definition for a directive that is easy-to-use for code authors and fits into the data model of the analyzer.
+- Create a more general definition for metadata (ex: @polymerElement) that is attached to an already-existing feature.
```
justinfagnani: I'm not sure this goal relates to the others on directives.

design/comment-scanning.md
```diff
@@ -0,0 +1,110 @@
...
+- Documentation (README, docs/ entry, docs site entry, etc) that communicates these new definitions.
+
+## Non-Goals
+- Implement any general-purpose metadata/directive scanners. This document will focus solely on the syntax and the way we talk about scannable documentation going forward.
```
justinfagnani: If there are no general purpose scanners, then the thing this doc really talks about would be the `Directive` feature. We should have it sketched: what properties does it have?

design/comment-scanning.md
FredKSchott: Fair enough, what I'm trying to communicate here is that one of my goals is to clearly differentiate directives from other analysis-related-info contained in comments. 

Do you think this shouldn't be a goal of the document? I'd remove most references to Comment Annotations in that case. Or, is there a better way to phrase this goal?

design/comment-scanning.md
FredKSchott: See above, my intention here is just to define this type of comment that describes an already-existing feature as specifically not a directive. I'm not suggesting we make any changes to how these are currently scanned. I'll make that more clear.

design/comment-scanning.md
FredKSchott: Correct, a directive is always a standalone feature. It will never be attached to another feature.

A comment annotation would be used in that case to annotate/add information to an attached feature. They are specifically always attached to a feature.

A plan for pluggable annotation scanners wasn't a goal here, but perhaps a simple plan would help. Will add a brief section on this.

design/comment-scanning.md
justinfagnani: I don't think jsdoc tags should be a goal here, but I do think that attached directive should be. We need a way to select rules for an element, and a range is more cumbersome than an attached comment.

design/comment-scanning.md
FredKSchott: I specifically ignored this use case for v1 because I'm not sure how useful it is, but happy to discuss the value in adding it.

Mainly, I couldn't think a good scenario where this would be useful. Why would a user want to enable a rule for only a single section or file? If they don't care about that rule across their entire project, why would they care in that one place? If a user was migrating to use a new rule, we should encourage them to turn it on globally and then disable individually so that the directives would be removed as the migration progressed.

FWIW, `IgnoreDirective` was pseudo code to help simplify this specific example. In the actual implementation the directive wouldn't be so coupled to ignore-only.

design/comment-scanning.md
FredKSchott: The polymer-linter directive scanner would scan for specifically these strings

design/comment-scanning.md
FredKSchott: +1, will add

design/comment-scanning.md
FredKSchott: See directive examples below:

```
/* polymer-lint disable: undefined-elements, rule-2 */
/* @lazyImport '../some-url/some-file.html */
<!-- @registerServiceWorkerHere -->
```

The current design doc is intentionally flexible around what a directive looks like, but the flip side to that is that writing a common scanner is impossible. Additionally, different directives describe different things, so a common parser would be limited in what it could do. For example, the directives created by the `polymer-lint` enable/disable comments depends on the args list given.

design/comment-scanning.md
FredKSchott: Will explain this more as a part of the fleshing out of the `Directive` feature type

design/comment-scanning.md
justinfagnani: An example would be enabling the hybrid or 2.0 rule set for sections that are being migrated.

design/comment-scanning.md
justinfagnani: it'd be nice to easily create directive scanners, something like: `new DirectiveScanner(lintEnableRegex, getLintEnableProps)`

FredKSchott: +1, a `GenericDirectiveScanner` implementation or a `createDirectiveScanner` factory would be useful for helping consumers add support for their own directives.

design/comment-scanning.md
FredKSchott: Please see changes, I've gone into more detail about Annotations and how they differ from directives. 

I don't defining a range as cumbersome compared to the uncertainty with how an attached directive would behave. For example:
- If it is implicit whether a directive is attached or standalone, how do we communicate that to the user and not give them a 👟🔫? 
- What if you want to define an unattached directive above a feature? It would always be "attached" by the AST parser.
- What if an unattached directive ("lint enable") exists inside of a feature with an attached directive ("lint disable")? Where is linting enabled and where is it disabled? How do they interact?

If you desire for "attached" directives is convenience, I'd recommend we kick that until after we've had some users play with standalone directives.

(As a point of reference, eslint let comment ranges bake for a long before adding any sugar for single-line enable/disable, next-line, etc. They don't have a concept of "attached" and as far as I can tell no one has ever asked for it)

design/comment-scanning.md
FredKSchott: added

design/comment-scanning.md
FredKSchott: fixed, explained better

design/comment-scanning.md
FredKSchott: Changed example to not confuse by using a tag already supported by jsdoc. The set of all JS Annotations, however, is a superset of jsdoc. JSDoc tags would be included in the `feature.annotation` Set. This is an intentional design decision.

design/comment-scanning.md
FredKSchott: Fair enough, that makes sense. Changed example & wording to support both enable & disable rules.

justinfagnani:
Since I can't respond to this comment thread anymore... quoting:

> I don't defining a range as cumbersome compared to the uncertainty with how an attached directive would behave.

A design would address the uncertainty by specifying the behavior to make it... certain :)

> For example:
> 
> If it is implicit whether a directive is attached or standalone, how do we communicate that to the user and not give them a 👟🔫?
> What if you want to define an unattached directive above a feature? It would always be "attached" by the AST parser.
> What if an unattached directive ("lint enable") exists inside of a feature with an attached directive ("lint disable")? Where is linting enabled and where is it disabled? How do they interact?
> If you desire for "attached" directives is convenience, I'd recommend we kick that until after we've had some users play with standalone directives.

These are all great questions! These are exactly the kinds of things a design would need to address.

It's also fine to say that some feature was considered and rejected or something is a non-goal because of difficulty, scope, etc. I ask about attached directives because 1) currently almost all of our comment handling is via attached comments, and 2) we will want to turn on and off lint rules for specific elements, which could have directives attached to them. When I read the doc I'd like to be easily able to tell the scope, and see answers to previous questions about design direction. Adding it to non-goals is appreciated.

> 
> (As a point of reference, eslint let comment ranges bake for a long before adding any sugar for single-line enable/disable, next-line, etc. They don't have a concept of "attached" and as far as I can tell no one has ever asked for it)

design/comment-scanning.md
```diff
@@ -0,0 +1,159 @@
...
+/* polymer-lint enable: undefined-elements */
+Directive({identifier: 'polymer-lint enable', args: ['undefined-elements']})
+/* lazy-import: '../some-url/some-file.html' */
+Directive({identifier: 'lazy-import', args: ['../some-url/some-file.html']})
+<!-- polymer-build:register-service-worker-here -->
+Directive({identifier: 'polymer-build:register-service-worker-here', args: null})
+```
+
+
+### Comment Annotation
```
justinfagnani: I would just kill this section. we already handle this and it doesn't seem tied into the the other directives in any way.

design/comment-scanning.md
```diff
@@ -0,0 +1,159 @@
...
+- Because of the wide range of possible features these could support, specific comment syntax is lax. However they should follow these guidelines:
+  - it is unique enough to never collide with other directive formats
+  - it is formal enough to be matched by substring or RegEx matching
+  - if it takes arguments they should be parsed in the scanning phase
+- Each comment directive will have a scanner responsible for scanning for it.
+  - A scanner may be responsible for scanning multiple directives (ex: `polymer-lint enable`, `polymer-lint disable`)
+Here is a definition for the **Directive** class:
+
+```
+class Directive extends Feature {
```
justinfagnani: Do we expect this to be subclassed?

Is there any code that will rely on a common interface between Directives? Is this interface needed if not?

design/comment-scanning.md
```diff
@@ -0,0 +1,159 @@
...
+> Last Update: 2017-04-26
+
+## Objective
+- To analyze standalone comments (directives) for instructions.
+- To analyze "attached" comments (annotations) for information/metadata about an existing feature.
+
+
+## Goals
+- To scan & analyze polymer-lint directives so that the linter can properly report & ignore warnings in a document.
+- Create an extensible system for analyzing and surfacing standalone comments in a document.
```
FredKSchott: I wanted to flesh out this definition to answer the "what if I want to attach information to a feature" question, but I agree that I've probably gone into too much detail at this point. 

Let me see if I can boil this down to just the relevent-to-directives bits.

design/comment-scanning.md
```diff
@@ -0,0 +1,159 @@
...
+- Documentation (README, docs/ entry, docs site entry, etc) that communicates this new system.
+
+
+## Non-Goals
+- Implementing any "attached" comment annotation scanning. It was important to outline so that the scope of comment directives could be explained, but implementation will be left as future work. The current `jsdoc` property on JavaScript features is good enough for the moment.
+

+## Background
+
+Currently, our analyzer scans for explicit documentation to detect features & feature characteristics that would otherwise be missed. For example, the `@customElement` tag is used to signify that the following class definition is a custom element.
```
FredKSchott: Same answer as above :) will shrink this.

design/comment-scanning.md
FredKSchott: 👍 Will add

FredKSchott:
@justinfagnani thanks for the second round of feedback. Slightly unrelated, but I'm coming to the realization that "light" design docs might never really be feasible. Each time I try to write one the sections that I skipped or skimped on are always brought up as questions in review, and I end up writing them anyway (ex: rejected designs). 

Going forward I'm just going to write full design docs, and save my reviewers the time / confusion of having to assume the details I left out.

FredKSchott:
Responding to feedback (really wish github would let us keep these threaded convos going):


>> I don't see defining a range as cumbersome compared to the uncertainty with how an attached directive would behave.

>A design would address the uncertainty by specifying the behavior to make it... certain :)

Sorry, I should clarify. I meant that even if we came up with a set of certain rules to define what an attached directive is, I'm concerned that they would be unintuitive to use and hard to communicate clearly. 

I'll come up with the best proposal I can for how they would work with some explicit behavior, and then we can have something concrete to talk about.

> These are all great questions! These are exactly the kinds of things a design would need to address.

I'll add a "rejected designs" / "non-goals" section to this design document.

design/comment-scanning.md
FredKSchott: Clarified this in the newest draft, but to be explicit directive scanners (including 3rd party) should return `Directive` instances. It supports all use-cases that I could think of. 

Like any other feature, we don't need to explicitly stop anyone from subclassing it. But it should be discouraged and definitely should not be required.

FredKSchott:
Updated the design doc with a summary of what we discussed and where we landed on this. Can everyone read it over and make sure I didn't miss anything important before merging?

PR #649, opened 2017-04-28
Scanner for Polymer 1.x core "features".

aomarks:
- [x] CHANGELOG.md has not been updated


src/model/resolvable.ts
```diff
@@ -18,9 +18,13 @@ import {Feature, ScannedFeature} from './feature';
 /**
  * A ScannedFeature that needs to resolve other Features to build its final
  * representation.
+ *
+ * TODO(aomarks) Is it right that I made this an abstract class? ScannedFeature
```
justinfagnani: The only reason to have a pure abstract class rather than an interface is to be able to inherit from it to get it on the prototype chain and use `instanceof`. Since we want to move away from `instanceof` because node/npm, I'm not sure there is a good reason here.

An interface extending a class isn't a weird thing in TypeScript, where every class also defines an implicit interface. This is the same as if we defined both an interface and base class for ScannedFeature and Resolvable extended the interface. I do question whether this interface actually needs to extend ScannedFeature though.

src/polymer/polymer-core-feature-scanner.ts
```diff
@@ -0,0 +1,100 @@
...
+        if (!isAddFeatureCall) {
+          return;
+        }
+
+        const description = esutil.getAttachedComment(parent) || '';
+        const feature = new ScannedPolymerCoreFeature(
+            document.sourceRangeForNode(node),
+            node,  // TODO(aomarks) Correct? The astNode param is not typed.
+            description,
+            // TODO(aomarks) Should I set jsdoc?
```
justinfagnani: In general yes, but I don't think it matters here, since we don't use any of the docs in this case. Also we'd probably want the docs on the argument, not the function call, and I'm not sure what they would represent yet.

src/polymer/polymer-core-feature.ts
```diff
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
+ * Copyright (c) 2017 The Polymer Project Authors. All rights reserved.
```
justinfagnani: don't update copyright dates

src/polymer/polymer-core-feature.ts
```diff
@@ -12,9 +12,21 @@
...
-export interface ScannedPolymerCoreFeature extends ScannedPolymerElement {}
+export class ScannedPolymerCoreFeature extends Resolvable {
+  sourceRange: SourceRange|undefined;
+  warnings: Warning[] = [];
+
+  /** Properties this `_addFeature` call augments `Polymer.Base` with. */
+  properties: ScannedPolymerProperty[] = [];
+
+  resolve(_document: Document): Feature|undefined {
+    // TODO(aomarks) Implement.
```
justinfagnani: I think you can go ahead and implement a simple version in this PR that just constructs a PolymerCoreFeature object, copying over properties.

src/polymer/polymer-core-feature-scanner.ts
```diff
@@ -0,0 +1,100 @@
...
+      severity: Severity.ERROR,
+      code: 'invalid-polymer-core-feature-call',
+      sourceRange: document.sourceRangeForNode(node)!,
+    });
+    return;
+  }
+
+  for (const prop of arg.properties) {
+    // TODO(aomarks) Can we have sourceRangeForNode never return undefined?
+    feature.properties.push(
```
justinfagnani: This needs to be a little more complex to handle methods.
This isn't really the best implementation, but it's code that handles properties, methods, getters and setters: https://github.com/Polymer/polymer-analyzer/blob/master/src/polymer/polymer-element-scanner.ts#L280

It's be nice to clean this up and pull it out so that you _can_ just call one method to handle a property and then see if you get a ScannedProperty or ScannedMethod back.

src/test/polymer/polymer-core-feature_test.ts
```diff
@@ -0,0 +1,66 @@
...
+import {JavaScriptParser} from '../../javascript/javascript-parser';
+import {PolymerCoreFeatureScanner} from '../../polymer/polymer-core-feature-scanner';
+
+suite('PolymerCoreFeatureScanner', () => {
+
+  test('scans _addFeature call expressions', async () => {
+    const js = `
+      /** Feature A */
+      Polymer.Base._addFeature({
+        methodA: function() {}
```
justinfagnani: I'd add a non-function property to one of the tests

src/polymer/polymer-core-feature.ts
```diff
@@ -12,9 +12,21 @@
...
- * The metadata for a Polymer core feature.
+ * Scanned representation of a Polymer 1.x core "feature".
  */
-export interface ScannedPolymerCoreFeature extends ScannedPolymerElement {}
+export class ScannedPolymerCoreFeature extends Resolvable {
+  sourceRange: SourceRange|undefined;
+  warnings: Warning[] = [];
+
+  /** Properties this `_addFeature` call augments `Polymer.Base` with. */
+  properties: ScannedPolymerProperty[] = [];
```
justinfagnani: add methods

src/test/polymer/polymer-core-feature_test.ts
```diff
@@ -0,0 +1,66 @@
...
+    const {features} = await scanner.scan(doc, visit);
+
+    assert.lengthOf(features, 3);
+    const [a, b, c] = features;
+
+    // TODO(aomarks) Why aren't the comments stripped correctly?
+    assert.equal(a.description, '* Feature A ');
+    assert.lengthOf(a.warnings, 0);
+    assert.lengthOf(a.properties, 1);
+    assert.equal(a.properties[0].name, 'methodA');
```
justinfagnani: this should be a method

src/model/resolvable.ts
rictic: I think Resolvable must extend ScannedFeature, as it doesn't make sense to resolve anything but a scanned feature.

We have resolve methods on some non-features (like references), but they're deliberately called on instances of known specific types so no interfaces are needed there. We may want to add arguments to those resolve method calls too.

I'm kinda in favor of extending from ScannedFeature and having folks pass stuff up through the constructor, as a way to reduce duplication of properties and better ensure that constraints are met. e.g. we could ensure that sourceRange is always set.

src/polymer/polymer-core-feature-scanner.ts
```diff
@@ -0,0 +1,100 @@
...
+            <estree.MemberExpression>node.callee,
+            ['Polymer', 'Base', '_addFeature']);
+        if (!isAddFeatureCall) {
+          return;
+        }
+
+        const description = esutil.getAttachedComment(parent) || '';
+        const feature = new ScannedPolymerCoreFeature(
+            document.sourceRangeForNode(node),
+            node,  // TODO(aomarks) Correct? The astNode param is not typed.
```
rictic: astNode is deliberately untyped because we want to preserve the option to declare an element purely with HTML. If we specified that astNode should be an estree node then we couldn't set the html node.

src/polymer/polymer-core-feature.ts
```diff
@@ -12,9 +12,21 @@
...
+import {ScannedPolymerProperty} from './polymer-element';
 
 /**
- * The metadata for a Polymer core feature.
+ * Scanned representation of a Polymer 1.x core "feature".
  */
-export interface ScannedPolymerCoreFeature extends ScannedPolymerElement {}
+export class ScannedPolymerCoreFeature extends Resolvable {
+  sourceRange: SourceRange|undefined;
+  warnings: Warning[] = [];
```
rictic: only need to re-declare properties from extended classes if you're refining the types.

that said, consider making them readonly since they're only set in the constructor, and making the arrays ImmutableArrays (from src/core/immutable)

src/test/polymer/polymer-core-feature_test.ts
aomarks: It's currently coming out as a PolymerProperty with type:Function. Maybe related to #647.

src/polymer/polymer-core-feature.ts
aomarks: Yeah, I'll add the resolve logic to this PR and ping it again. Just wanted to get a review going on the first part.

src/polymer/polymer-core-feature-scanner.ts
aomarks: Can we define a type union for the possible types then so at least we get some sanity checking? (I tried `estree.Node|ASTNode|null` but things got weird).

src/model/resolvable.ts
aomarks: If Resolvable remains an interface, then the compiler requires me to re-declare all the properties from the ScannedFeature definition (description, sourceRange, etc.), and also I can't call the ScannedFeature constructor -- because I'm no longer extending an abstract class, I'm implementing an interface derived from it, so all the abstract provisions are lost. Which seems to defeat the purpose of having an abstract class in the hierarchy?

@rictic Sorry, I'm not sure if you're advocating for the interface or the abstract class?

src/polymer/polymer-core-feature-scanner.ts
aomarks: What's description supposed to represent vs jsdoc? The documentation for both of these fields just says that they each probably shouldn't be there :P.

src/polymer/polymer-core-feature.ts
aomarks: I do need to re-declare them unless I change Resolved to be an abstract class instead of an interface. So this depends on the discussion above.

src/test/polymer/polymer-core-feature_test.ts
```diff
@@ -0,0 +1,66 @@
...
+    const parser = new JavaScriptParser();
+    const scanner = new PolymerCoreFeatureScanner();
+    const doc = parser.parse(js, 'features.js');
+    const visit = (visitor: Visitor) => Promise.resolve(doc.visit([visitor]));
+    const {features} = await scanner.scan(doc, visit);
+
+    assert.lengthOf(features, 3);
+    const [a, b, c] = features;
+
+    // TODO(aomarks) Why aren't the comments stripped correctly?
```
aomarks: *ping*

src/model/resolvable.ts
aomarks: Talked with Justin, cleared up some misunderstandings I had. Going to keep Resolvable an interface in this PR.

src/polymer/polymer-core-feature-scanner.ts
aomarks: Fixed.

src/polymer/polymer-core-feature-scanner.ts
aomarks: Methods are now supported. No getter/setters yet, but I'm not sure there actually are any on Polymer.Base (I'll double check though). Did not merge the two functions for now, but still an option.

src/polymer/polymer-core-feature.ts
aomarks: Done.

src/polymer/polymer-core-feature.ts
aomarks: Done.

src/polymer/polymer-core-feature.ts
aomarks: Now emitting resolved features: intermediate PolymerCoreFeatures, and a final Class for the aggregated object.

src/test/polymer/polymer-core-feature_test.ts
aomarks: Done.

src/test/polymer/polymer-core-feature_test.ts
aomarks: Done.

src/polymer/polymer-core-feature.ts
aomarks: Still doing it for warnings because I want to add an initializer.

src/test/polymer/polymer-core-feature_test.ts
aomarks: Fixed.

aomarks:
PTAL

I added the other half of this feature (emitting things from resolve), and changed the scanner quite significantly too.

src/polymer/polymer-core-feature.ts
```diff
@@ -12,9 +12,88 @@
...
+
+  resolve(document: Document): Feature|undefined {
+    if (this.warnings.length > 0) {
+      return;
+    }
+
+    // Sniff for the root `_addFeatures` call by presence of this method. This
+    // will normally be in the `polymer.html` document.
+    const isRootAddFeatureCall =
+        this.methods.some((p) => p.name === '_finishRegisterFeatures');
```
rictic: You confirmed that this is only present in the polymer core source in one place? Kevin wasn't sure that it was a good indicator

src/polymer/polymer-core-feature.ts
```diff
@@ -12,9 +12,88 @@
...
+
+    if (!isRootAddFeatureCall) {
+      // We're at the `Polymer.Base` assignment or a regular `_addFeature`
+      // call. We'll merge all of these below.
+      return new PolymerCoreFeature(this.properties, this.methods);
+
+    } else {
```
rictic: can drop the `else` and dedent, as the other path returned

src/polymer/polymer-core-feature.ts
```diff
@@ -12,9 +12,88 @@
...
+        }
+      }
+      return new Class(
+          {
+            name: 'Polymer.Base',
+            className: 'Polymer.Base',
+            properties: allProperties,
+            methods: allMethods,
+            abstract: true,
+            privacy: 'public',  // TODO(aomarks) Correct privacy?
```
rictic: this is correct

src/polymer/polymer-core-feature.ts
```diff
@@ -12,9 +12,88 @@
...
+            // TODO(aomarks) The following should probably come from the
+            // Polymer.Base assignment instead of this final addFeature call.
```
rictic: +1

src/polymer/polymer-core-feature.ts
```diff
@@ -12,9 +12,88 @@
...
+export class PolymerCoreFeature implements Feature {
+  kinds = new Set(['polymer-core-feature']);
+  identifiers = new Set<string>();
+  warnings: Warning[] = [];
+
+  constructor(public properties: Property[], public methods: Method[]) {
+  }
+}
+
+declare module '../model/queryable' {
```
rictic: the convention elsewhere is is to put this above the feature rather than below

src/test/polymer/polymer-core-feature_test.ts
```diff
@@ -0,0 +1,120 @@
...
+    const scanner = new PolymerCoreFeatureScanner();
+    const doc = parser.parse(js, 'features.js');
+    const visit = (visitor: Visitor) => Promise.resolve(doc.visit([visitor]));
+    const {features} = await scanner.scan(doc, visit);
+
+    assert.lengthOf(features, 4);
+    const [a, b, base, invalid] = features;
+
+    assert.equal(a.description, 'Feature A');
+    assert.lengthOf(a.warnings, 0);
```
rictic: Opinion: testing lengths is a last resort, because the error message is so very unhelpful. Consider instead:

```typescript
assert.deepEqual(a.warnings, []);
assert.deepEqual(
  a.methods.map({name, type, description} => ({name, type, description})), 
  [{name: 'methodA', type: 'Function', description: 'Method A'}]
);
```

It's shorter, but more importantly, the error message will be actionable because it will give you a diff of what was expected and what was found, which is usually all you need to track down the bug.

aomarks: Done.

src/polymer/polymer-core-feature.ts
aomarks: Done.

src/polymer/polymer-core-feature.ts
aomarks: There's actually no documentation on (any?) of the calls we're analyzing here, so this kind of doesn't matter, and I might punt for now because it will add some complexity to track the Base assignment.

Hydrolysis actually just hard coded a description: https://github.com/Polymer/polymer-analyzer/blob/hydrolysis-1.x/src/ast-utils/docs.ts#L356. Could do the same thing here. Could also add some jsdoc to the 1.0 polymer branch.

src/polymer/polymer-core-feature.ts
aomarks: Done.

src/polymer/polymer-core-feature.ts
aomarks: Done.

src/polymer/polymer-core-feature.ts
aomarks: Yep, it's only declared in one place. Updated the comment to make it clear we depend on this fact.

src/polymer/polymer-core-feature-scanner.ts
```diff
@@ -0,0 +1,153 @@
...
+ *
+ * Example: https://github.com/Polymer/polymer/blob/1.x/src/mini/debouncer.html
+ */
+export class PolymerCoreFeatureScanner implements JavaScriptScanner {
+  async scan(
+      document: JavaScriptDocument,
+      visit: (visitor: Visitor) => Promise<void>) {
+    const visitor = new PolymerCoreFeatureVisitor(document);
+    await visit(visitor);
+    return {features: visitor.features, warnings: []};
```
aomarks: Done.

aomarks:
@rictic As discussed offline, added serialized `Class.name` and renamed `Element.classname` to `name` (inherited from class).

src/test/polymer/polymer-core-feature_test.ts
```diff
@@ -62,43 +62,62 @@ suite('PolymerCoreFeatureScanner', () => {
...
-          // This directory contains 3 files copied from Polymer 1.x core.
+          // This directory contains files copied from Polymer 1.x core.
           path.resolve(__dirname, '../static/polymer-core-feature/')),
     });
 
     const analysis = await analyzer.analyzePackage();
     const features = analysis.getFeatures({id: 'Polymer.Base', kind: 'class'});
     assert.equal(features.size, 1);
     const polymerBase = features.values().next().value;
-    assert.lengthOf(polymerBase.methods, 35);
```
rictic: It's better to have these assertions than not to, I think. Just in cases where you can reasonably do the assertion as a deepEquals that's much better.

aomarks: Done.

src/polymer/polymer-core-feature.ts
```diff
@@ -12,9 +12,88 @@
...
+
+/**
+ * A resolved Polymer 1.x core "feature".
+ */
+export class PolymerCoreFeature implements Feature {
+  kinds = new Set(['polymer-core-feature']);
+  identifiers = new Set<string>();
+  warnings: Warning[] = [];
+
+  constructor(public properties: Property[], public methods: Method[]) {
```
justinfagnani: why the empty constructor?

aomarks: The public keywords assign the parameters to `this`.

PR #658, opened 2017-05-04
Adjoining HTML comment text is now treated as jsdoc and structure of demos in analysis format is updated.

usergenic:
- HTML comments prepending the dom-modules for polymer elements are parsed as jsdocs first, enabling annotations like `@demo` in the html comment.
 - In cases where HTML comment exists as well as Jsdoc comment, we join the results so we have a set of tags from both and description is merged (separted by an empty line).
 - "demos" property of generated analysis changed from string[] to Demo[] (where demo is `{url: string, description: string}`).
 - Addressed a bug where previously resolve() was modifying scanned element in-place potentially resulting in modifications on multiple successive calls.


usergenic:
cc @samuelli

src/analysis-format.ts
```diff
@@ -383,3 +383,11 @@ export interface Slot extends Feature {
 
   // Something about fallback perhaps?
 }
+
+export interface Demo extends Feature {
```
rictic: Should it extend Feature? Can it have a source range or a metadata?

src/model/element-base.ts
```diff
@@ -44,7 +44,15 @@ export abstract class ScannedElementBase implements Resolvable {
...
-    this.description = this.description || commentText || '';
+    if (commentText) {
+      const commentJsdoc = jsdoc.parseJsdoc(commentText);
+      this.jsdoc =
+          this.jsdoc ? jsdoc.join(this.jsdoc, commentJsdoc) : commentJsdoc;
+      this.description = [
+        commentJsdoc.description || '',
+        this.description || ''
+      ].join('\n\n').trim();
+    }
```
rictic: ooh, I like this. I like it a lot.

src/analysis-format.ts
usergenic: I guess that's not really a supported thing the way demos and jsdoc tags are treated right now.  Not extended anymore.

src/model/element-base.ts
usergenic: :D

rictic: My read of Justin's position was in slack was that he only holds a weak objection against merging, but a strong preference for warning when there's both. That seems reasonable to me, and that merging also gets us unblocked for the site.

I think the logic would be go here and would be pretty easy, just do: `this.warnings.push({...}` if this.jsdoc has a description or any tags.

usergenic: Ok.  Added warning and a test for the warning.

usergenic:
@rictic updated with warning and tests for warning.  PTAL

src/model/element-base.ts
```diff
@@ -46,8 +47,20 @@ export abstract class ScannedElementBase implements Resolvable {
...
       const commentJsdoc = jsdoc.parseJsdoc(commentText);
+      // Add a Warning if there are already jsdoc tags or a description for this
+      // element.
+      if (this.sourceRange &&
+          (this.description || this.jsdoc && this.jsdoc.tags.length > 0)) {
+        this.warnings.push({
+          severity: Severity.WARNING,
+          code: 'multiple-doc-comments',
+          message:
+              `More than one doc comment found for ${this.constructor.name}`,
```
rictic: probably helpful to mention that this is probably because of one comment found in html and another found in js

src/test/polymer/polymer-element_test.ts
```diff
@@ -17,22 +17,33 @@ import {assert} from 'chai';
...
 import {FSUrlLoader} from '../../url-loader/fs-url-loader';
 
 suite('PolymerElement', () => {
   const testFilesDir = path.resolve(__dirname, '../static/polymer2/');
   const urlLoader = new FSUrlLoader(testFilesDir);
   const analyzer = new Analyzer({
     urlLoader: urlLoader,
-    scanners: new Map([[
-      'js',
+    scanners: new Map<string, any[]>([
```
rictic: IMO building an analyzer with a subset of scanners is an antipattern, because what we care about is how the analyzer works with all the scanners.

Cards on the table, I tend to see integration tests as the ideal, and unit tests as compromises for speed or clarity. Reasonable people may disagree :)

src/test/polymer/polymer-element_test.ts
```diff
@@ -206,4 +217,33 @@ suite('PolymerElement', () => {
...
+      assert.isUndefined(warning);
+    });
+
+    test('Elements with more than one doc comment have warning', async() => {
+      const element = await getElement('test-element-15.html');
+      const warning = element.warnings.find(
+          (w: Warning) => w.code === 'multiple-doc-comments')!;
+      assert.isDefined(warning);
+      assert.equal(warning.severity, Severity.WARNING);
+      assert.equal(
```
rictic: deepEqual is preferred pretty much everywhere, even for comparing strings, as it gives better error messages

src/test/polymer/polymer-element_test.ts
usergenic: @rictic for this test, there was a single scanner defined and I tried to add a minimum to get test to pass.  Would you recommend I just strip out the scanners option in this suite altogether?

rictic: Yeah, IMO the most important thing to test is how it works in the default configuration.

src/model/element-base.ts
usergenic: Fixed.

src/test/polymer/polymer-element_test.ts
usergenic: Fixed.

src/test/polymer/polymer-element_test.ts
usergenic: Fixed.

PR #662, opened 2017-05-05
Add a MultiUrlLoader for loader composition.

usergenic:
This `MultiUrlLoader` is a very simple implementation.  It depends on `canLoad` to choose the loader to use, in sequence.

My intent with this loader is to support dispatching to multiple loaders to support cases such as the `redirect` option bundler use to use to deal with Chrome's special protocol paths.  To do that will require a different form of the current FSUrlLoader than we have, but because this multi-loader is such a basic compositional tool for loader assembly, I thought I'd add in its own PR.

### DIFF
--- a/src/test/url-loader/multi-url-loader_test.ts
+++ b/src/test/url-loader/multi-url-loader_test.ts
@@ -1,0 +1,134 @@
+/**
+ * @license
+ * Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at
+ * http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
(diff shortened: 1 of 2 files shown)


src/test/url-loader/multi-url-loader_test.ts
```diff
@@ -0,0 +1,134 @@
...
+      // Verify only the last load is called
+      assert.equal(loaders[0].loadCount, 0);
+      assert.equal(loaders[1].loadCount, 0);
+      assert.equal(loaders[2].loadCount, 1);
+    });
+
+    test('throws an error if no loader can be found to load', async() => {
+      const loaders = mockLoaderArray([null, null, null]);
+      const loader = new MultiUrlLoader(loaders);
+      let error;
```
rictic: can use invertPromise to write this as:

```typescript
const error = await invertPromise(loader.load('test.html'));
```

usergenic: Oh!  Thank you!  That's way simpler.  Updated.

rictic: :)

PR #663, opened 2017-05-05
index.ts should be the only toplevel src file

rictic:


rictic:
hm, a few files are still mis-aligned, closing for the moment.

PR #665, opened 2017-05-05
index.ts should be the only toplevel src file

rictic:
- [x] internal change, CHANGELOG.md not updated


src/test/core/dependency-graph_test.ts
```diff
@@ -17,9 +17,9 @@
 import {assert, use} from 'chai';
 import * as path from 'path';
 
-import {Analyzer} from '../../analyzer';
+import {Analyzer} from '../../core/analyzer';
 import {DependencyGraph} from '../../core/dependency-graph';
-import {FSUrlLoader} from '../../url-loader/fs-url-loader';
+import {FSUrlLoader} from '../../index';
```
justinfagnani: I don't think we should import from index.ts internally, even in tests

justinfagnani:
meh. Honestly, I really dislike `index.js` files. A package does not need to appear like a single module, and it's not actually an "index" of anything. but ok.

rictic:
I agree that a package doesn't need to look like a single module, but I do like having a clear delineation between public API and private implementation details.

src/test/core/dependency-graph_test.ts
rictic: Done, here and elsewhere.

PR #666, opened 2017-05-05
Prefixed url loader 🔥🤘

usergenic:
In combination with the `MultiUrlLoader`, the `PrefixedUrlLoader` provides a mechanism to map multiple url prefixes (containing protocols, domains, paths, whatever), to different delegate loaders.  This provides developers the solve several use cases:

 - Route a specific url to a common directory outside the package root.  E.g. I want to point `/bower_components/polymer/` to `~/src/polymer-fork`.
```js
new MultiUrlLoader([
  new PrefixedUrlLoader('bower_components/polymer', 
    new FSUrlLoader('~/src/polymer-fork')), 
  new FSUrlLoader()])
```
 - Chrome can create a sequence of mappings for each chrome protocol like: (`bin/polymer-bundler` could easily construct this and implement the old vulcanize `--redirect` option.
```js
new PrefixedUrlLoader('chrome://downloads/',
  new FSUrlLoader('/tmp/polymer_bundler_test/folder2/'))
```
 - Route a specific url to a remote copy of the package in question.:
```js
new MultiUrlLoader([
  new PrefixedUrlLoader('bower_components/polymer/',
    new FetchUrlLoader('https://userx.mycdn.com/files/components/polymer/')),
  new FSUrlLoader('.')])
```

### DIFF
--- a/src/test/url-loader/prefixed-url-loader_test.ts
+++ b/src/test/url-loader/prefixed-url-loader_test.ts
@@ -1,0 +1,111 @@
+/**
+ * @license
+ * Copyright (c) 2017 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at
+ * http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import {assert} from 'chai';
+
+import {PrefixedUrlLoader} from '../../url-loader/prefixed-url-loader';
+import {UrlLoader} from '../../url-loader/url-loader';
+import {invertPromise} from '../test-utils';
+
+class MockLoader implements UrlLoader {
+  canLoadUrls: string[];
+  loadUrls: string[];
+  constructor(private _load: string|null) {
+    this.reset();
+  }
+
+  reset() {
+    this.canLoadUrls = [];
+    this.loadUrls = [];
+  }
+
+  canLoad(url: string): boolean {
+    this.canLoadUrls.push(url);
+    return this._load != null;
+  }
+
+  async load(url: string): Promise<string> {
+    this.loadUrls.push(url);
+    if (this._load == null) {
+      throw new Error(`Tried to load "${url}", and delegate can't load it.`);
+    }
+    return this._load;
+  }
+}
+
+suite('PrefixedUrlLoader', () => {
+
+  suite('canLoad', () => {
+
+    test('canLoad is true if the url starts with prefix', () => {
+      const delegate = new MockLoader('stuff');
+      const loader = new PrefixedUrlLoader('path/to/stuff/', delegate);
+      assert.isTrue(loader.canLoad('path/to/stuff/file.html'));
+      // Delegate receives an unprefixed url to check.
+      assert.deepEqual(delegate.canLoadUrls, ['file.html']);
+    });
+
+    test('canLoad is false if the url does not start with prefix', () => {
+      const delegate = new MockLoader('stuff');
+      const loader = new PrefixedUrlLoader('path/to/stuff/', delegate);
+      assert.isFalse(loader.canLoad('path/to/other/file.html'));
+      // Delegate is not consulted.
+      assert.deepEqual(delegate.canLoadUrls, []);
+    });
+
+    test('canLoad is false if the delgate loader says it is', () => {
+      const delegate = new MockLoader(null);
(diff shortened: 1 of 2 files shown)


usergenic:
Note, I'm aware of RedirectResolver, but the primary difference to a solution leveraging PrefixedUrlLoader over RedirectResolver is that you can direct PrefixedUrlLoader at whatever kind of underlying UrlLoader you want, which means you can tie together sources that don't conform to expected path relativity and are outside of package root.

src/url-loader/prefixed-url-loader.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  async load(url: string): Promise<string> {
+    if (!url.startsWith(this.prefix)) {
+      throw new Error(
+          `Can not load "${url}", does not match prefix "${this.prefix}".`);
+    }
+    return this.delegate.load(this.unprefix(url));
+  }
+
+  unprefix(url: string): string {
```
FredKSchott: private?

src/url-loader/prefixed-url-loader.ts
```diff
@@ -0,0 +1,41 @@
...
+  constructor(public prefix: string, public delegate: UrlLoader) {
+  }
+
+  canLoad(url: string): boolean {
+    return url.startsWith(this.prefix) &&
+        this.delegate.canLoad(this.unprefix(url));
+  }
+
+  async load(url: string): Promise<string> {
+    if (!url.startsWith(this.prefix)) {
```
FredKSchott: just sanity checking my understanding of the analyzer: this should never happen, right?

src/url-loader/prefixed-url-loader.ts
```diff
@@ -0,0 +1,41 @@
...
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import {UrlLoader} from './url-loader';
+
+/**
+ * Resolves requests via a given delegate loader for URLs matching a given
+ * prefix.  URLs are provided to their delegate without the prefix.
```
FredKSchott: nit: single space after periods.

src/test/url-loader/prefixed-url-loader_test.ts
```diff
@@ -0,0 +1,111 @@
+/**
+ * @license
+ * Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
```
FredKSchott: 2017

src/url-loader/prefixed-url-loader.ts
```diff
@@ -0,0 +1,41 @@
+/**
+ * @license
+ * Copyright (c) 2015 The Polymer Project Authors. All rights reserved.
```
FredKSchott: party like its 2015

src/url-loader/prefixed-url-loader.ts
usergenic: I am old and still I use two spaces after a period everywhere because it actually is better but the whole world changed when proportional fonts came around and people have since been learning this _wrong_.

I realize I am fighting a losing battle and even our codebase predominantly uses single space after end-of-sentence punctuation-- even files specifically intended to be rendered with monospace fonts, which has typically been the last safe refuge of my two-space kind.

I will convert this two-space to a single-space, but I do it with a heavy heart and only in the knowledge that some day the truth will be known that single-space after the end of sentence punctuation was the wrong direction for a world, though this wisdom may be rediscovered too late.

too-late

src/url-loader/prefixed-url-loader.ts
usergenic: This would happen a bunch if this prefixed loader was working correctly.  Consider you want to route all requests for "bower_components/polymer/" to a temp checkout "/tmp/polymer-test-branch" or something, then the requested url would not start with that prefix most of the time.

Same for if you wanted to handle traffic to "chrome://about/whatever/1" to go to a specific folder, that URL prefix is specific and unlikely to match for all urls.

src/url-loader/prefixed-url-loader.ts
usergenic: Yeah might as well.  Was on the fence about it.  Making it private now.

src/test/url-loader/prefixed-url-loader_test.ts
usergenic: Done.  Thanks :)

src/url-loader/prefixed-url-loader.ts
usergenic: Fixed.

usergenic:
Squashed and pushed.

src/url-loader/prefixed-url-loader.ts
usergenic: OH WAIT - you were saying the load shouldn't actually get called with a bad URL.  Yeah most of the time that's right, because canLoad would be called in advance of load to verify.

So yeah you're right about that.  Shouldn't happen-- but if it did, this guard ensures it doesn't send the load request to the delegate.

PR #667, opened 2017-05-08
Updated CHANGELOG for 2.0.0-alpha.41 release.

usergenic:
* [BREAKING] The `ElementLike` interface in `AnalysisFormat` changed its `demos` property from `string[]` to `Demo[]`, to include more information from `@demo` annotations in Jsdocs.
 * [minor breaking change] Simplify rules for infering privacy. Now all features: classes, elements, properties, methods, etc have one set of rules for inferring privacy. Explicit js doc annotations are respected, otherwise `__foo` and `foo_` are private, `_foo` is protected, and `foo` is public.
 * Fixed issue where jsdocs syntax in HTML comments was not parsed for polymer elements.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,10 +11,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
    Unreleased section, uncommenting the header as necessary.
 -->
 
-## Unreleased
+## [2.0.0-alpha.41] - 2017-05-08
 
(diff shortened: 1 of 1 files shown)


rictic:
IMO, no need to copy the changelog over to here. the whole purpose of the changelog is to be canonical repository of that info

usergenic:
Fair point.  I just liked the way the description showed up in the Slack notification :-p

rictic:
Cool, there's definitely no harm in it

Issue #668, opened 2017-05-08
Don't resolve `@demo` urls.


usergenic:
Apparently some elements get relative urls wrong in their demo annotations and path resolving these essentially loses fidelity.  For now, lets stop resolving them until we know the extent of the problem.

usergenic:
cc @samuelli @justinfagnani

usergenic:
![image](https://cloud.githubusercontent.com/assets/578/25828386/c38f5a68-3404-11e7-9159-5e37a1132430.png)

> (the bower_components in those paths are wc.org related and the package gets installed there)
> -- @samuelli

usergenic:
@samuelli can you provide the `analyze()` call you made to obtain the analysis?

usergenic:
#670 fixed this.

PR #670, opened 2017-05-08
Do not attempt to resolve the Jsdoc `@demo` paths.

usergenic:
- `@demo` annotation paths are treated (by existing documentation tools) as absolute with respect to the packages they are contained in.
 - Analyzer doesn't know about imported package roots and so resolving these paths to the analyzed package root is not possible.
 - We don't try to resolve them any more.
 - Also cleaned up places where `demos` property was being set and removed it from constructor options since it only works as a derived value right now and the presence of it in Options gave the false impression that value was honored.


usergenic:
Fixes https://github.com/Polymer/polymer-analyzer/issues/668

src/model/element-base.ts
```diff
@@ -61,15 +61,16 @@ export abstract class ScannedElementBase implements Resolvable {
       }
       this.jsdoc =
           this.jsdoc ? jsdoc.join(commentJsdoc, this.jsdoc) : commentJsdoc;
+      this.applyJsdocDemoTags();
```
rictic: All the other classes extend ElementBase, can we remove all the other calls to applyJsdocDemoTags()?

usergenic: I have been trying to find a way.  Originally these calls were in `resolve()` because they needed the document url to resolve the paths.   However, we all kind of talked about how resolve was supposed to be idempotent and not really supposed to mutate the scanned element and such so I was happy to pull these out of that method.

The problem is the demos need to be extracted when the jsdoc changes.  This happens in multiple places and there's no hook other than to make the call in the same places AFAIK.

I did pull it out of ScannedElementMixin where it doesn't do any good anyways because jsdoc isn't assigned.  So I have to keep the call in 3 places-- one in ScannedElementBase in the special case where we apply the html comment, and in ScannedPolymerElementMixin and ScannedPolymerElement after jsdoc is set.

src/model/class.ts
```diff
@@ -151,7 +151,7 @@ export class Class implements Feature {
 
     this.warnings =
         init.warnings === undefined ? [] : Array.from(init.warnings);
-    this.demos = init.demos || jsdocLib.extractDemos(init.jsdoc, document.url);
+    this.demos = init.demos || jsdocLib.extractDemos(init.jsdoc);
```
rictic: So this is happening in `Class` so this is at resolution time. The prototype chain of `Scanned*` is kinda crazy, but as of fairly recently, it's very sane for resolved features. I think this is the only place this is needed, so long as we don't care that scanned features don't have demos, which seems fine to me.

src/polymer/polymer-element-mixin.ts
```diff
@@ -68,6 +68,7 @@ export class ScannedPolymerElementMixin extends ScannedElementMixin implements
     this.mixins = mixins;
     this.astNode = astNode;
     this.classAstNode = classAstNode;
+    this.applyJsdocDemoTags();
```
rictic: I think we can drop this

src/model/element-base.ts
```diff
@@ -61,15 +61,16 @@ export abstract class ScannedElementBase implements Resolvable {
...
       this.description = [
         commentJsdoc.description || '',
         this.description || ''
       ].join('\n\n').trim();
     }
   }
 
-  applyJsdocDemoTags(baseUrl: string): void {
-    this.demos = jsdoc.extractDemos(this.jsdoc, baseUrl);
+  applyJsdocDemoTags(): void {
```
rictic: I think we can drop this.

src/polymer/polymer-element.ts
```diff
@@ -185,6 +183,7 @@ export class ScannedPolymerElement extends ScannedElement implements
         (summaryTag !== undefined && summaryTag.description != null) ?
         summaryTag.description :
         '';
+    this.applyJsdocDemoTags();
```
rictic: I think we can drop this

rictic:
This is good simplification and reduction, I think we can go farther!

src/model/element-base.ts
usergenic: Ok dropped applyJsdocDemoTags

src/polymer/polymer-element.ts
usergenic: Dropped

src/model/class.ts
usergenic: Okay the problem was that init.demos always shows up as an empty array so the || was not useful.  I've opted to combine them now `this.demos = [...init.demos || [], ...jsdocLib.extractDemos(init.jsdoc)]`

src/model/class.ts
```diff
@@ -151,7 +151,7 @@ export class Class implements Feature {
 
     this.warnings =
         init.warnings === undefined ? [] : Array.from(init.warnings);
-    this.demos = init.demos || jsdocLib.extractDemos(init.jsdoc, document.url);
+    this.demos = [...init.demos || [], ...jsdocLib.extractDemos(init.jsdoc)];
```
rictic: If `init.demos` is always the empty array, we could consider making it optional and leaving it out most places?

PR #673, opened 2017-05-12
Make warning a class rather than an interface

rictic:

One last thing question here. `Warning` is a bit of an awkward name for this, as warnings can have error severity or info severity, etc. If there's a better name this would be the time to change it, as we'll need to do one last pass to update all uses anyways. The Microsoft world calls them Diagnostics, which I find a touch too abstract. The atom linter calls them Messages, which I kinda like.

### DIFF
--- a/src/model/warning.ts
+++ b/src/model/warning.ts
@@ -13,11 +13,31 @@
  */
 
 import {SourceRange} from './source-range';
-export interface Warning {
+
+export interface WarningInit {
   readonly message: string;
   readonly sourceRange: SourceRange;
   readonly severity: Severity;
   readonly code: string;
+}
+export class Warning {
+  readonly code: string;
+  readonly message: string;
+  readonly sourceRange: SourceRange;
+  readonly severity: Severity;
+
+
+  // Useful while we migrate from object literal warnings to a warning class.
+  protected _warningBrand: never;
+
+  constructor(init: WarningInit) {
+    ({
+      message: this.message,
+      sourceRange: this.sourceRange,
+      severity: this.severity,
+      code: this.code
+    } = init);
+  }
 }
 
 export enum Severity {
--- a/src/model/warning.ts
+++ b/src/model/warning.ts
@@ -12,6 +12,8 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
+import {ParsedDocument} from '../index';
+
 import {SourceRange} from './source-range';
 
 export interface WarningInit {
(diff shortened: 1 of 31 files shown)


justinfagnani:
I think Warning is an ok name, though I see your point. For our use, even errors are situations we _think_ are errors, but due to the über-dynamic nature of the platform, may not be. We can think of a Warning with severity Error as a "Warning: This looks like an Error".

src/model/warning.ts
```diff
@@ -12,12 +12,114 @@
...
+    return {
+      code: this.code,
+      message: this.message,
+      severity: this.severity,
+      sourceRange: this.sourceRange,
+    };
+  }
 }
 
 export enum Severity {
```
justinfagnani: We really need some definition of the levels.

src/model/warning.ts
```diff
@@ -12,12 +12,114 @@
...
+    const colorize = opts.color ? this._severityToColorFunction(this.severity) :
+                                  (s: string) => s;
+    const severity = this._severityToString(colorize);
+
+    if (!this.sourceRange || !this._parsedDocument) {
+      return `INTERNAL ERROR: Tried to print a '${this.code}' ` +
+          `warning without a source range and/or parsed document. ` +
+          `Please report this!\n` +
+          `     https://github.com/Polymer/polymer-analyzer/issues/new\n` +
+          `${this._severityToString(colorize)} ` +
```
justinfagnani: use `severity`

src/model/warning.ts
```diff
@@ -12,12 +12,114 @@
...
+  }
+
+  toString(options: Partial<WarningStringifyOptions> = {}): string {
+    const opts:
+        WarningStringifyOptions = {...defaultPrinterOptions, ...options};
+    const colorize = opts.color ? this._severityToColorFunction(this.severity) :
+                                  (s: string) => s;
+    const severity = this._severityToString(colorize);
+
+    if (!this.sourceRange || !this._parsedDocument) {
```
justinfagnani: why can't you do this check on construction? It'll be closer to the real error and a stack might be useful.

src/model/warning.ts
rictic: Agreed, later.

src/model/warning.ts
rictic: removed

src/model/warning.ts
rictic: +1, done

rictic:
Got a verbal approval from Justin on this. Pinkie swear.

PR #676, opened 2017-05-16
Change many errors to warnings

justinfagnani:

### DIFF
--- a/src/polymer/polymer-core-feature-scanner.ts
+++ b/src/polymer/polymer-core-feature-scanner.ts
@@ -74,7 +74,7 @@ class PolymerCoreFeatureVisitor implements Visitor {
       feature.warnings.push(new Warning({
         message: `Expected assignment to \`Polymer.Base\` to be an object.` +
             `Got \`${rhs.type}\` instead.`,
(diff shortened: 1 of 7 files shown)


src/test/core/analyzer_test.ts
```diff
@@ -175,37 +175,6 @@ suite('Analyzer', () => {
     });
 
     test(
```
rictic: why drop this test?

src/polymer/polymer-core-feature-scanner.ts
```diff
@@ -74,7 +74,7 @@ class PolymerCoreFeatureVisitor implements Visitor {
       feature.warnings.push(new Warning({
         message: `Expected assignment to \`Polymer.Base\` to be an object.` +
             `Got \`${rhs.type}\` instead.`,
-        severity: Severity.ERROR,
+        severity: Severity.WARNING,
```
rictic: Seems like errors here are reasonable to me. They should only show up when something's really messed with with Polymer core?

justinfagnani: This is true, but we should also never see them

PR #679, opened 2017-05-16
Update README with the stable API

rictic:

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,26 +16,36 @@ npm install polymer-analyzer
 const {Analyzer, FSUrlLoader} = require('polymer-analyzer');
 
 let analyzer = new Analyzer({
-  urlLoader: new FSUrlLoader(pathToPackageRoot),
+  urlLoader: new FSUrlLoader('/path/to/package/root'),
 });
 
 // This path is relative to the package root
-analyzer.analyze(['./my-element.html'])
-  .then((analysis) => {
-    // Gets all elements reachable from the file above, including imports.
-    for (const element of analysis.getFeatures({kind: 'element'})) {
-      console.log(element);
+analyzer.analyze(['./my-element.html']).then((analysis) => {
+  // Print the name of every property on paper-button, and where it was
+  // inherited from.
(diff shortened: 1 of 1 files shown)


rictic:
Careful review is welcomed and encouraged, this will be most people's first interaction with the analyzer

README.md
```diff
@@ -19,9 +19,19 @@ let analyzer = new Analyzer({
   urlLoader: new FSUrlLoader(pathToPackageRoot),
```
davidrleonard: Perhaps rather than making this a magic variable name not defined in the scope of this example (`pathToPackageRoot`), this could be a string similar to the path passed to analyzer.analyze?

e.g. `urlLoader: new FSUrlLoader('/path/to/package/root')`

README.md
```diff
@@ -19,9 +19,19 @@ let analyzer = new Analyzer({
...
+    }
+    
+    // To look just in my-element.html
+    const document = analysis.getDocument('my-element.html');
+    // Note, document could be null if it wasn't found, 
+    // or a Warning if it had a parse error.
+    for (const element of maybeDocument.getFeatures({kind: 'element'})) {
       console.log(element);
     }
   });
```
aomarks: It would be nice to see more examples of the kinds of features analyzer finds, and what the features look like. How about a new section with a list of some of the interesting ones?

aomarks: How about adding something about where analyzer is used for context? Mention that it powers all the documentation at webcomponents.org, the polymer CLI build pipeline, linter, etc.?

README.md
rictic: +1, done

README.md
rictic: I gave a more interesting example. This isn't the place for an API doc, though that would be a very useful thing for us to generate.

rictic: +1, done

aomarks: Don't want to mention webcomponents.org?

rictic: oh! yes, added

rictic:
Landing this before Justin's review so that the readme gives a valid example. Happy to iterate more

PR #682, opened 2017-05-18
Track methods on behaviors properly.

### LINKED ISSUE #681: Function properties should be methods
_From @samuelli on May 18, 2017 18:1_

Not sure if this should be filed here or on analyzer. It appears that in some cases, functions are show as properties with type function instead of being a method.

This is an example which has both "methods" and property "functions": https://www.webcomponents.org/element/PolymerElements/app-layout/elements/app-header

_Copied from original issue: PolymerElements/iron-doc-viewer#129_

rictic:
We were treating them as properties, but now if they're function typed we call 'em methods.

Fixes #681

### DIFF
--- a/src/test/static/js-behaviors.js
+++ b/src/test/static/js-behaviors.js
@@ -9,6 +9,9 @@ var SimpleBehavior = {
  * */
 var SimpleNamespacedBehavior = {
   simple: true,
+  method: function (paramA, paramB) {
+
+  },
 };
 
 
@@ -33,4 +36,4 @@ Really.Really.Deep.Behavior = [
 @polymerBehavior
 */
 CustomBehaviorList =
-    [SimpleBehavior, CustomNamedBehavior, Really.Really.Deep.Behavior];
+    [SimpleBehavior, AwesomeBehavior, Really.Really.Deep.Behavior];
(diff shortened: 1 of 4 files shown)


src/test/static/js-behaviors.js
```diff
@@ -33,4 +36,4 @@ Really.Really.Deep.Behavior = [
 @polymerBehavior
 */
 CustomBehaviorList =
-    [SimpleBehavior, CustomNamedBehavior, Really.Really.Deep.Behavior];
+    [SimpleBehavior, AwesomeBehavior, Really.Really.Deep.Behavior];
```
samuelli: This should still be `CustomNamedBehavior`? Only the jsdoc references `AwesomeBehavior`

rictic: The jsdoc is considered canonical in cases where it specifies a name.

samuelli: Isn't this meant to reference the object on line 19 though?

rictic: Yes. So this code is incorrect unless there was some other context that assigned CustomNamedBehavior to the AwesomeBehavior global, but that's what the jsdoc is telling us happens. Fortunately this is just a test fixture, this code is never executed :)

samuelli: I see!

PR #686, opened 2017-05-23
Added Document.isDocument and then used it everywhere.

usergenic:
Created a static method to replace `d instanceof Document` calls in order to support scenarios where multiple versions of `polymer-analyzer` wind up in `node_modules`.

This is mostly for consideration as a general approach to this class of problem.  Not sure if:

1.) sufficient to address problem of multiple versions of analyzer in the mix
2.) people are going to have multiple versions of analyzer now that our tools require it as "^2.x", but it *might* enable a degree of 2.x and 3.x interop?

Mixed feelings now that I went through the exercise...

### DIFF
--- a/src/model/document.ts
+++ b/src/model/document.ts
@@ -139,7 +139,7 @@ export class Document implements Feature, Queryable {
   }
 
   static isDocument(value: any): value is Document {
-    return typeof value === 'object' && !!value.constructor.isDocument;
+    return value && value.constructor && !!value.constructor.isDocument;
   }
 
   get url(): string {
--- a/src/model/document.ts
+++ b/src/model/document.ts
@@ -138,6 +138,10 @@ export class Document implements Feature, Queryable {
     this.warnings = Array.from(base.warnings);
   }
 
+  static isDocument(value: any): value is Document {
+    return typeof value === 'object' && !!value.constructor.isDocument;
+  }
+
   get url(): string {
     return this._scannedDocument.url;
   }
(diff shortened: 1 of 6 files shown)


usergenic:
Note that this whole issue comes about because Analyzer `Anaysis#getDocument()` returns `Document|Warning|undefined`.  Potentially we add `Analysis#getDocumentOrDie()|getDocumentOrThrow()`?

usergenic:
@rictic so this issue is actually still relevant because when people npm link, they will not have deduped versions of node modules; even if the versions are the same under the linked package, the instanceof checks will fail.  For example, we had a user who npm link'd polymer-bundler and the analyzer under their bundler modules was same version as top level, but instanceof checks still failed etc.

src/model/document.ts
```diff
@@ -138,6 +138,10 @@ export class Document implements Feature, Queryable {
     this.warnings = Array.from(base.warnings);
   }
 
+  static isDocument(value: any): value is Document {
+    return typeof value === 'object' && !!value.constructor.isDocument;
```
rictic: This throws if given `null`. `typeof null` is 'object' D: :D D: :D

How about just:

```
return value && value.constructor && value.constructor.isDocument;
```

usergenic: Your fix is good and you should feel good.

*Why did I think null would have a constructor just because it is an object?*

rictic:
I feel slightly uncomfortable with this, just because it seems like such a compromise. If it's the right compromise then it seems like we should move entirely to Foo.isFoo methods, not just doing it on Document.

Should we commit to moving all `instanceof` checks to these Foo.isFoo checks?

usergenic:
@rictic FWIW I'm going with https://github.com/Polymer/polymer-bundler/pull/567 as a means to counter the confusion devs have when they wind up with multiple polymer-analyzers.  I naively think that converting all code to use `X.isX(x)` would fix this class of problem, but then of course we are more likely to open users up to more subtle incompatibility bugs among different versions of analyzer which are pretending to be the same but actually aren't.

Maybe I close this PR for now, having run through the exercise, it isn't clear that this is a win, beyond the `Analysis#getDocument` case.  Have a look at my workaround PR linked above just for reference -- thinking that a `getDocumentOrThrow/Die` method is going to be more ergonomic.

PR #690, opened 2017-06-01
Added an Analysis#getDocumentOrDie(url: string) method.

usergenic:
 - Added getDocumentOrDie() as a convenient option to just get a known analyzed document by url instead of dealing with multiple return type.

### DIFF
--- a/src/model/analysis.ts
+++ b/src/model/analysis.ts
@@ -65,6 +65,12 @@ export class Analysis implements Queryable {
     this._searchRoots = potentialRoots;
   }
 
+  /**
+   * Returns a `Document` for the given url, if one was analyzed.
+   * Returns a `Warning`, if one was generated when analyzing the URL.
+   * Returns `undefined` if the URL does not reference an analyzed
+   * document at all.
+   */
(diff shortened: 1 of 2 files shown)


src/model/analysis.ts
```diff
@@ -82,6 +88,22 @@ export class Analysis implements Queryable {
   }
 
   /**
+   * Simplified version of getDocument which simply throws an error when asked
+   * for a missing document.
+   */
+  getDocumentOrDie(url: string): Document {
```
justinfagnani: `OrThrow`?

justinfagnani: I think this should be viewed as temporary until we have getDocument return a better result class that makes it easier to handle warnings.

justinfagnani:
@usergenic wdyt, should we close this and change the signature of getDocument() in 3.0 to return a better type?

rictic:
Ooh! Yeah! We should totally do that for 3.0

rictic:
Use a real `Result` type

usergenic:
Closing this PR and created https://github.com/Polymer/polymer-analyzer/issues/772

PR #693, opened 2017-06-03
Track static methods, expose them in analysis.json

rictic:

\cc @azakus


src/javascript/class-scanner.ts
```diff
@@ -211,6 +213,7 @@ export class ClassScanner implements JavaScriptScanner {
       astNode,
       properties,
       methods,
+      staticMethods,
```
aomarks: How come this is the class scanner, but everything's talkin' bout elements?

src/polymer/polymer2-config.ts
```diff
@@ -90,3 +86,29 @@ export function getMethods(node: estree.Node, document: JavaScriptDocument):
   }
   return methods;
 }
+
+export function getStaticMethods(
```
aomarks: Why is this in polymer2-config.ts if it's just about class static methods? Also document it?

src/javascript/class-scanner.ts
rictic: We don't want to emit a Class for something that's an element, so there's a bit of a hack in here to notice that something is actual an element and treat it as such. Otherwise we emit the Class.

src/polymer/polymer2-config.ts
rictic: Good point. Moved to ./javascript/esutil.ts

src/javascript/class-scanner.ts
aomarks: Ah I see. I guess it would be nice in terms of organization if there was an element scanner that was separate to the class function, and a minimal shared function that could distinguish between the two.

aomarks: (obviously not for now even if that makes sense)

rictic: Yeah, that'd be a nice refactoring to do. This fell victim to some of the time pressure as we rushed to get things together.

rictic:
Thanks for the quick review!

Issue #691, opened 2017-06-01
analyzer class issues


aomarks:
Found some issues with the analyzer output for classes in the Polymer 2.0 core library:

```sh
$ git clone git@github.com:Polymer/polymer
$ cd polymer
$ polymer analyze > analysis.json
```

1. The `Debouncer` class (https://github.com/Polymer/polymer/blob/master/lib/utils/debounce.html) is in top level `classes`, but I expected it to be in `namespaces.Polymer.classes`. Interesting that it has a `@memberof Polymer` annotation. The only class that *does* make it into the Polymer is  `FlattenedNodesObserver ` (https://github.com/Polymer/polymer/blob/master/lib/utils/flattened-nodes-observer.html) and it does *not* have that annotation.

2. The description for `Debouncer` is `eslint-disable-line no-unused-vars`, which is a standalone comment on the line before the class declaration. The class declaration has no jsdoc block.

rictic:
Hm, 1, 3, and 4 seem like bugs. #2 strikes me as working as intended, and the solution is to write a line or two of documentation for the Debouncer class.

aomarks:
On #2, the comment seems fairly clearly not attached to the class declaration because it's an end of line comment for another symbol's declaration.

```
  let AsyncModule; // eslint-disable-line no-unused-vars

  class Debouncer {
```

web-padawan:
I can confirm that placing the following block causes error in #674:
```
            /* eslint-disable camelcase */
            var DESTINATION_MAP = {
                google_analytics: 'selectAnalyticsAccess',
                click_house: 'selectClickHouseAccess'
            };
            /* eslint-enable camelcase */

            Polymer({
                is: 'my-element',
```

homerjonathan:
I have also had commenting errors.  

 **ScannedPolymerElement has both HTML doc and JSDoc comments.**

I managed to remove the errors I have this at the top

```
<!--
   Top Documentation
->
```

Then in the <script> section I have

`// Script Notes`

This returns the above error.  Remove the line and it no longer shows the errors.

PR #696, opened 2017-06-17
Don't associate comments with nodes unless they are touching.

aomarks:
Fixes one of the issues in https://github.com/Polymer/polymer-analyzer/issues/691


web-padawan:
@aomarks Could you tag this fix so we could use it in production?

aomarks:
@web-padawan I just released polymer-analyzer 2.2.0 with this change.

PR #702, opened 2017-07-05
Take TypeScript out of the production dependencies for Analyzer until support fully implemented

usergenic:
- Updated `package.json` and `package-lock.json` such that typescript is now a `devDependencies` member.
 - Only one non-test source file effected (`analysis-context.ts`), which had an import of the `TypeScriptPreparser`.
 - Updated the `TypeScriptParser` test suite by explicitly setting the parsers map for analyzer used in test, since it is not a registered default parser now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
    Unreleased section, uncommenting the header as necessary.
 -->
 
-<!--## Unreleased-->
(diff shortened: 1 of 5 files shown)


usergenic:
Note that the impetus for this is the overall reduction of 26MB of production dependencies, which will help migration to polymer-analyzer based tools for some users which package analyzer into their fabric of tools.

usergenic:
@freshp86

PR #704, opened 2017-07-06
Fix typescript pruning

usergenic:
- Was trying to release analyzer but typings failed a tsc call which for some reason isn't captured by our tests.
 - [ ] CHANGELOG.md doesn't need updating


usergenic:
Performed fixes on master branch.  `np` CLI had issues with the already present v2.2.1 tag.

PR #709, opened 2017-07-19
Scan fields from ES constructor declarations

### LINKED ISSUE #706: [Feature Request] Collect instance properties set in class constructor
I'd like to be able to collect instance properties for classes to help with closure interface generation with https://github.com/Polymer/gen-closure-declarations.
To do that in an expedient way, it would be enough to just collect assignments or references to instance properties from the `constructor`.

rictic:

This is a long awaited feature. @azakus is the latest to be pushing for it.

Fixes #706

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
   ],
   "scripts": {
     "clean": "mkdir -p lib/ && rm -rf lib/",
-    "build": "gulp build",
+    "build": "gulp build --silent",
     "release": "npm run compile",
     "lint": "gulp lint",
     "test": "npm run clean && npm run build && npm run lint && mocha",
--- a/src/javascript/class-scanner.ts
+++ b/src/javascript/class-scanner.ts
@@ -207,7 +207,7 @@ export class ClassScanner implements JavaScriptScanner {
         } else {
           finalProp = prop;
         }
-        class_.properties.set(prop.name, prop);
+        class_.properties.set(prop.name, finalProp);
       }
       methods = getMethods(astNode, document);
       staticMethods = getStaticMethods(astNode, document);
@@ -417,8 +417,7 @@ class ClassFinder implements Visitor {
     const warnings: Warning[] = [];
     const properties =
         extractPropertiesFromConstructor(astNode, this._document);
-    // TODO(rictic): scan the constructor and look for assignments to this.foo
-    //     to determine properties.
+
     this.classes.push(new ScannedClass(
         namespacedName,
         name,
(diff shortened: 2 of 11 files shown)


package.json
```diff
@@ -19,7 +19,7 @@
   ],
   "scripts": {
     "clean": "mkdir -p lib/ && rm -rf lib/",
-    "build": "gulp build",
+    "build": "gulp build --silent",
```
justinfagnani: why? can we get rid of gulp yet?

azakus:
Seems to work well for me in gen-closure-declarations!

src/javascript/class-scanner.ts
```diff
@@ -574,3 +585,93 @@ class CustomElementsDefineCallFinder implements Visitor {
     return null;
   }
 }
+
+export function extractPropertiesFromConstructor(
+    astNode: estree.Node,
+    document: JavaScriptDocument,
+    properties = new Map<string, ScannedProperty>()) {
+  if (astNode.type === 'ClassExpression' ||
```
justinfagnani: seems a good bail-early place

src/javascript/class-scanner.ts
```diff
@@ -574,3 +585,93 @@ class CustomElementsDefineCallFinder implements Visitor {
...
+          continue;
+        }
+        let name;
+        let astNode;
+        let defaultValue;
+        if (statement.expression.type === 'AssignmentExpression') {
+          name = getPropertyNameOnThisExpression(statement.expression.left);
+          astNode = statement.expression.left;
+          defaultValue = escodegen.generate(statement.expression.right);
+        } else if (statement.expression.type === 'MemberExpression') {
```
justinfagnani: can you document his branch? I assume you looking for just `/** @public */ this.foo;`

src/polymer/polymer-element.ts
```diff
@@ -43,8 +44,78 @@ export interface BasePolymerProperty {
...
+    propB: Readonly<ScannedPolymerProperty>): ScannedPolymerProperty {
+  if (propA.name !== propB.name) {
+    throw new Error(
+        `Tried to merge properties with different names: ` +
+        `'${propA.name}' and ' ${propB.name}'`);
+  }
+  const name = propA.name;
+  let description;
+  if (propA.description !== undefined && propB.description !== undefined) {
+    description = propA.description.length < propB.description.length ?
```
justinfagnani: I hate this logic, but don't have a better idea. Can we put it in a common location so we can share it and document it once?

package.json
rictic: it copies the static test fixtures over, as well as generating the json schema from typescript types.

src/javascript/class-scanner.ts
rictic: +1, done

src/javascript/class-scanner.ts
rictic: Yep, that's right. Done.

src/polymer/polymer-element.ts
rictic: Moved to jsdoc. It duplicates some of `jsdoc.join(...)`. I remember that there was a pretty good argument for that logic.. I think. But this looks correct for this case. Left a comment to investigate.

PR #715, opened 2017-08-26
Use upstream estraverse typings and update estree typings

TimvdLippe:
Estraverse typings landed upstream which are more elaborate (have the new `replace` function which I needed). This also required updating `estree` typings where several fields can now also be `null`. Some typings had to be updated to accept `null`. No logic changes were required as all of them could already be `undefined`.


TimvdLippe:
https://github.com/Polymer/polymer-analyzer/pull/715/files?w=1 makes it easier to review btw, it removes the formatting changes due to the longer argument declaration

PR #716, opened 2017-08-26
Resolve window references in extends clause

TimvdLippe:
While browsing through the http://wpt.fyi source, I found https://github.com/GoogleChrome/wptdashboard/blob/d769c47c245b5e68ba36b925c4950908f0a0d965/components/test-file-results.html#L79 which was showing warnings in my editor. It seems that `window.` parent classes were incorrectly referenced.

I am not 100% convinced this is the correct fix, as it could be that it introduces false positives when defining an object like `Polymer = { Element: ... }`, but that seems a rare edgecase. I also tried to modify https://github.com/Polymer/polymer-analyzer/blob/1d803aceced80c92501fb6144d530139778c66b6/src/javascript/class-scanner.ts#L381 to add it automatically on `window`, but that seemed to break a lot of existing tests.

### DIFF
--- a/src/test/static/polymer2/test-element-3.js
+++ b/src/test/static/polymer2/test-element-3.js
@@ -32,3 +32,17 @@ class SubElement extends BaseElement {
     };
   }
 }
+
+/**
+ * @customElement
+ * @polymer
+ */
+class WindowBaseElement extends window.Polymer.Element {}
+
+window.Polymer.WindowBaseElement = WindowBaseElement;
+
+/**
+ * @customElement
+ * @polymer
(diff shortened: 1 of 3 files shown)


src/test/static/polymer2/test-element-3.js
```diff
@@ -32,3 +32,17 @@ class SubElement extends BaseElement {
     };
   }
 }
+
+/**
+ * @customElement
+ * @polymer
```
rictic: I think you want to add `@memberOf Polymer` here

PR #719, opened 2017-08-29
Fix value of reflectToAttribute for polymer properties

TimvdLippe:
By accident I discovered that this value was incorrectly `false` if it was explicitly set to `true`. Seemed like a copy-paste error, but there was no existing test for it. Thus I also added an extra test to test for the different property configurations.


rictic:
Please update the CHANGELOG to mention this fix

rictic:
(also, good catch!)

TimvdLippe:
Oh right, done :+1:

PR #721, opened 2017-08-31
Change the visit() method of HTMLDocument so it always visits template content.

usergenic:
- The `visit` method now includes template children.
 - The feature scanners will need to decide how to handle features inside templates.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,10 +11,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
    Unreleased section, uncommenting the header as necessary.
 -->
 
- ## Unreleased
+## Unreleased
 
 * Scan for CSS custom variable uses and assignments.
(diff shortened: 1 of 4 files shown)


usergenic:
This behavior change is in service of getting `css-import` and `html-style` imports to be defined even when used inside of `<template>` tags.  If felt like adding explicit template descenders into scanners was about to get really redundant, so I thought maybe lets just do this at document level and leave it to a feature definition to decide if it wants to test for and ignore features defined inside templates instead.

rictic:
Hm. We should review our HTML scanners to be sure that they all want to visit template children, because as I recall some of them didn't. The reason we went with this is that it's easier to opt in like this than to opt out without our current API

usergenic:
I'm going to close this PR and do a separate one to address just stylesheet import and html style features for now.

PR #723, opened 2017-09-01
Set scriptElement on PolymerElement

### LINKED ISSUE #718: scriptElement of PolymerElement is always undefined
`PolymerElement` has a `scriptElement` property which it retrieves from `ScannedPolymerElement`. However, this property is always undefined. This is similar to `domModule`, for which in the constructor of `PolymerElement`, the actual `domModule` is retrieved. A similar fix should be used to obtain and set the correct `scriptElement`.

TimvdLippe:
Previously this attribute was always undefined, now it passed on when scanning script tags.

Fixes #718 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 * Scan for CSS custom variable uses and assignments.
 * Fix value of reflectToAttribute for polymer properties
+* Set scriptElement of PolymerElement to Node if it originated from a HTML Document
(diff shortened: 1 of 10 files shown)


rictic:
Yay to removing domModule on scanned element!

I think we should remove scriptElement too, because this isn't a problem specific to polymer elements. https://github.com/Polymer/polymer-analyzer/issues/557 is the proper solution IMO.

TimvdLippe:
@rictic updated!

rictic:
Nice cleanup!

PR #724, opened 2017-09-02
Resolve several TODO's and implement getDocumentContaining

### LINKED ISSUE #557: API wart: getting source ranges for inline document nodes
e.g.

```javascript
const document = analyzer.analyze('index.html'
const polymerElement = document.getByKind('polymer-element').values().next().value;
const astNode = polymerElement.astNode; // more realistically, walk within polymerElement.astNode somehow.

/** 
 * How to I get a source range for astNode? document.parsedDocument.sourceRangeForNode() 
 * will fail in this case, because the polymer element was almost certainly defined in an 
 * inline script in the HTML.
 */
```

The best solution I have for this is to be able to take a feature and ask the initial Document for its most specific containing Document. We could implement this using source ranges if nothing else.

Alternatively we could add a `document` field to `Feature` but that would mean that we couldn't have classes that implement both ScannedFeature and Feature, because of caching (the scanned cache would have entries from the resolve phase).

TimvdLippe:
Introduce `getDocumentContaining` to find containing inline document for a feature, originally from https://github.com/Polymer/polymer-linter/blob/9d692d3dbc621b9be898006d92f516ffa4a56eaf/src/polymer/call-super-in-callbacks.ts#L125

Also fix several other easy TODO's.

Fixes #557

### DIFF
--- a/src/model/analysis.ts
+++ b/src/model/analysis.ts
@@ -152,21 +152,25 @@ export class Analysis implements Queryable {
    * @param sourceRange Source range to search for in a document
    */
    getDocumentContaining(sourceRange: SourceRange|undefined) {
-   if (!sourceRange) {
-     return undefined;
-   }
-   let mostSpecificDocument: undefined|Document = undefined;
-   for (const doc of this.getFeatures({kind: 'document'})) {
-     if (isPositionInsideRange(sourceRange.start, doc.sourceRange)) {
-       if (!mostSpecificDocument ||
-           isPositionInsideRange(
-               doc.sourceRange!.start, mostSpecificDocument.sourceRange)) {
-         mostSpecificDocument = doc;
+     if (!sourceRange) {
+       return undefined;
+     }
+     let mostSpecificDocument: undefined|Document = undefined;
+     const [outerDocument] = this.getFeatures({kind: 'document', id: sourceRange.file});
+     if (!outerDocument) {
+       return undefined;
+     }
+     for (const doc of outerDocument.getFeatures({kind: 'document'})) {
+       if (isPositionInsideRange(sourceRange.start, doc.sourceRange)) {
+         if (!mostSpecificDocument ||
+             isPositionInsideRange(
+                 doc.sourceRange!.start, mostSpecificDocument.sourceRange)) {
+           mostSpecificDocument = doc;
+         }
        }
      }
+     return mostSpecificDocument;
    }
-   return mostSpecificDocument;
- }
 
   private _getDocumentQuery(query: Query = {}): DocumentQuery {
     return {
--- a/src/model/analysis.ts
+++ b/src/model/analysis.ts
@@ -141,6 +141,33 @@ export class Analysis implements Queryable {
     return Array.from(result);
   }
 
+  /**
+   * Potentially narrow down the document that contains the sourceRange.
+   * For example, if a source range is inside a inlineDocument, this function
+   * will narrow down the document to the most specific inline document.
+   *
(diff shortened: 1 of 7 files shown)


src/model/analysis.ts
```diff
@@ -118,6 +122,33 @@ export class Analysis implements Queryable {
...
 
+  /**
+   * Potentially narrow down the document that contains the sourceRange.
+   * For example, if a source range is inside a inlineDocument, this function
+   * will narrow down the document to the most specific inline document.
+   *
+   * @param sourceRange Source range to search for in a document
+   * @param document The document that contains the source range
+   */
+  getDocumentContaining(sourceRange: SourceRange|undefined, document: Document):
```
rictic: `document` seems unnecessary here if we have the analysis and the source range.

Could be written as:

```typescript
getDocumentContaining(sourceRange: SourceRange|undefined) {
  if (!sourceRange) {
    return undefined;
  }
  let mostSpecificDocument: undefined|Document = undefined;
  for (const doc of this.getFeatures({kind: 'document', id: sourceRange.file})) {
    ...
  }
}
```

src/test/core/analyzer_test.ts
```diff
@@ -257,6 +257,28 @@ suite('Analyzer', () => {
           assert.equal(subBehavior!.className, 'TestBehavior');
         });
 
+    test(
+        'an inline feature can narrow down its containing document',
```
rictic: Recommend using a `testName` local to store the test name so that indentation doesn't get wonky.

src/model/analysis.ts
TimvdLippe: I have updated the code, but I couldn't do the `id: sourceRange.file`, as inline documents explicitly do not have the url as an identifier (https://github.com/Polymer/polymer-analyzer/blob/master/src/model/document.ts#L135) I am not sure in what other way to retrieve this information. Maybe a recursive search starting at the root document would work?

rictic: Huh, I see. What about:

```javascript
const [outerDoc] = this.getFeatures({kind: 'document', id: sourceRange.file}));
if (outerDoc === undefined) {
  return undefined;
}
for (const doc of outerDoc.getFeatures({kind: 'document'})) {
   ...
}
```

src/model/analysis.ts
```diff
@@ -141,6 +144,30 @@ export class Analysis implements Queryable {
...
+   * will narrow down the document to the most specific inline document.
+   *
+   * @param sourceRange Source range to search for in a document
+   */
+   getDocumentContaining(sourceRange: SourceRange|undefined) {
+   if (!sourceRange) {
+     return undefined;
+   }
+   let mostSpecificDocument: undefined|Document = undefined;
+   for (const doc of this.getFeatures({kind: 'document'})) {
```
rictic: (moving this comment from the other collapsed thread because github is bad at code review)

This goes over all documents, which could be hundreds or thousands. We can get the toplevel doc like so:

```javascript
const [outerDoc] = this.getFeatures({kind: 'document', id: sourceRange.file}));
if (outerDoc === undefined) {
  return undefined;
}
for (const doc of outerDoc.getFeatures({kind: 'document'})) {
   ...
}
```

TimvdLippe: Ah yes that was exactly what I was looking for!

rictic:
Oh, you have merge rights now, I should probably have left that for you to merge.

Sorry!

TimvdLippe:
Feel free to merge. Less work for me ;)

PR #732, opened 2017-10-11
Parse script v parse module

usergenic:
In preparation for supporting explicit module import graph traversal and handling of JS modules vs JS scripts, I updated the `javascript-parser.ts` code and tests to add `JavaScriptScriptParser` and `JavaScriptModuleParser` classes.

We had discussed `parseScript` and `parseModule` methods as part of `JavaScriptParser` except the interface for `Parser` in analyzer is they provide a single `parse()` method and this was the way to continue support for that interface without losing anything.

Also, I preserve the `JavaScriptParser` behavior precisely because the analyzer has a file-extension-to-parser map which would have to be redesigned if a single extension could refer to parsers of different types.  That can be refactored later, but this provides a way to add the desired strictness without refactoring the entire analyzer in this step.

### DIFF
--- a/src/javascript/javascript-parser.ts
+++ b/src/javascript/javascript-parser.ts
@@ -43,8 +43,8 @@ export class JavaScriptParser implements Parser<JavaScriptDocument> {
       JavaScriptDocument {
     const isInline = !!inlineInfo;
     inlineInfo = inlineInfo || {};
-    const result =
-        parseJs(contents, url, inlineInfo.locationOffset, this.sourceType);
+    const result = parseJs(
+        contents, url, inlineInfo.locationOffset, undefined, this.sourceType);
     if (result.type === 'failure') {
       // TODO(rictic): define and return a ParseResult instead of throwing.
       const minimalDocument = new JavaScriptDocument({
--- a/src/javascript/javascript-parser.ts
+++ b/src/javascript/javascript-parser.ts
@@ -111,23 +111,19 @@ export function parseJs(
   let program: estree.Program;
 
   try {
+    // If sourceType is not provided, we will try script first and if that
+    // fails, we will try module, since failure is probably that it can't parse
+    // the 'import' or 'export' syntax as a script.
     if (!sourceType) {
       try {
         sourceType = 'script';
-        program = espree.parse(
-            contents,
-            Object.assign(
-                {sourceType: 'script' as SourceType}, baseParseOptions));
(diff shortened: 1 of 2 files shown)


src/javascript/javascript-parser.ts
```diff
@@ -92,45 +102,56 @@ export function parseJs(
...
+      } catch (_ignored) {
+        sourceType = 'module';
+        program = espree.parse(
+            contents,
+            Object.assign(
+                {sourceType: 'module' as SourceType}, baseParseOptions));
+      }
+    } else {
+      program =
+          espree.parse(contents, Object.assign({sourceType}, baseParseOptions));
```
justinfagnani: use object spread

src/javascript/javascript-parser.ts
```diff
@@ -92,45 +102,56 @@ export function parseJs(
...
   try {
+    if (!sourceType) {
+      try {
+        sourceType = 'script';
+        program = espree.parse(
+            contents,
+            Object.assign(
+                {sourceType: 'script' as SourceType}, baseParseOptions));
+      } catch (_ignored) {
+        sourceType = 'module';
```
justinfagnani: I wonder if this should check for a `SyntaxError`. @rictic why didn't we check before? Can we add a comment for posterity now?

src/javascript/javascript-parser.ts
```diff
@@ -92,45 +102,56 @@ export function parseJs(
...
+
+  let program: estree.Program;
+
   try {
+    if (!sourceType) {
+      try {
+        sourceType = 'script';
+        program = espree.parse(
+            contents,
+            Object.assign(
```
justinfagnani:
```js
espree.parse(contents, {sourceType: 'script' as SourceType, ...baseParseOptions});
```

src/javascript/javascript-parser.ts
```diff
@@ -92,45 +102,56 @@ export function parseJs(
...
+                {sourceType: 'script' as SourceType}, baseParseOptions));
+      } catch (_ignored) {
+        sourceType = 'module';
+        program = espree.parse(
+            contents,
+            Object.assign(
```
justinfagnani: use object spread

src/javascript/javascript-parser.ts
usergenic: Nice!  I didn't realize `...` was for more than just array-type splatting.

PR #733, opened 2017-10-13
JavaScript module import scanner

usergenic:
- Scan for module imports in inline and external JavaScript, analyzing the entire import graph.

### DIFF
--- a/src/model/inline-document.ts
+++ b/src/model/inline-document.ts
@@ -39,7 +39,7 @@ export interface InlineDocInfo<AstNode> {
  * @template N The AST node type
  */
 export class ScannedInlineDocument implements ScannedFeature, Resolvable {
-  type: 'html'|'javascript'|'css'|/* etc */ string;
+  type: 'html'|'js'|'css'|/* etc */ string;
 
   contents: string;
 
--- a/src/test/html/html-script-scanner_test.ts
+++ b/src/test/html/html-script-scanner_test.ts
@@ -101,11 +101,16 @@ suite('HtmlScriptScanner', () => {
 
       test('follows import statements found in inline modules', async() => {
         const jsImports = [...analysis.getFeatures({kind: 'js-import'})];
-        assert.equal(jsImports.length, 1);
-        const js = jsImports[0].document.parsedDocument as JavaScriptDocument;
-        assert.equal(js.parsedAsSourceType, 'module');
+        assert.equal(jsImports.length, 2);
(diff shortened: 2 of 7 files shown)


src/model/inline-document.ts
```diff
@@ -39,7 +39,7 @@ export interface InlineDocInfo<AstNode> {
  * @template N The AST node type
  */
 export class ScannedInlineDocument implements ScannedFeature, Resolvable {
-  type: 'html'|'javascript'|'css'|/* etc */ string;
+  type: 'html'|'js'|'css'|/* etc */ string;
```
justinfagnani: this might be a breaking change?

src/test/html/html-script-scanner_test.ts
```diff
@@ -63,5 +69,37 @@ suite('HtmlScriptScanner', () => {
...
+      let analysis: Analysis;
+
+      before(async() => {
+        analysis = await analyzer.analyze(['js-modules.html']);
+      });
+
+      test('finds external module scripts', () => {
+        const htmlScripts = [...analysis.getFeatures({kind: 'html-script'})];
+        assert.equal(htmlScripts.length, 1);
+        const js = <JavaScriptDocument>htmlScripts[0].document.parsedDocument;
```
justinfagnani: style nit: prefer `as` to cast

src/model/inline-document.ts
usergenic: This 'javascript' type value was never actually used anywhere and was just an incorrect declaration insofar as I could tell.  I had to trace code around to verify this, but there were no cases I could find where the type value 'javascript' was used but there actually was one where 'js' was being used and so was against the type def anyways.  So from that perspective, I would expect this not to be a breaking change since the Analyzer never generated an inline document with that value for type.

src/test/html/html-script-scanner_test.ts
usergenic: Done.  I also went ahead and verified the contents of the document contents in the tests to ensure each referred to the expected material.  I also added the missing `./submodule.js` file to verify full import graph traversal.

src/model/inline-document.ts
usergenic: To put this another way: this is not a breaking change.  It's a bug fix; the type was wrong here AFAICT.

usergenic: @rictic ^ would you agree?

rictic: Yep. This also won't cause compile errors, because the fallback type `string` is still valid.

PR #739, opened 2017-10-31
Prepare to release 2.4.1

rictic:


rictic:
Also added v8 to tests. Previously we got it for free because it was the latest version of node, but now v9 is the latest version (which we still want to test) and v8 is a long term release branch.

PR #741, opened 2017-11-02
Fix reporting of internal error.

rictic:
Likely related to https://github.com/Polymer/polymer-cli/issues/768

Previously we were casting an object to Warning, but this isn't valid. Warning is a class now and we depend on its methods.

Constructing a Warning in this internal error case was a little tricky, but this should work.

If this is what's behind https://github.com/Polymer/polymer-cli/issues/768 then we'll still need more work to track down the root cause, because this code path is one of those "this should never happen" situations, but at least it'll be clearer about what's happening and where.


src/core/analysis-context.ts
```diff
@@ -496,4 +487,28 @@ export class AnalysisContext {
     return this.resolver.canResolve(url) ? this.resolver.resolve(url) :
                                            (url as any as ResolvedUrl);
   }
+
+  /**
+   * A warning for a weird situation that should not obtain.
```
justinfagnani: "should not obtain" -> "should not happen"

src/core/analysis-context.ts
```diff
@@ -496,4 +487,28 @@ export class AnalysisContext {
...
+   *
+   * Before calling getDocument(), which is synchronous, a caller must first
+   * have finished loading and scanning, as those phases are asynchronous.
+   *
+   * So we need to construct a warning, but we don't have a parsed document.
```
justinfagnani: `.` => `,` and a lowercase "so"

justinfagnani: It'd be nice if either 1) warnings didn't require documents, or 2) we produced a warning on the initiator document

rictic: done

src/core/analysis-context.ts
rictic: done

src/core/analysis-context.ts
rictic: re 2), for most warnings we do, including parse errors, but for this "impossible" warning we don't even have that much info

I'd like to hold off on 1)

justinfagnani:
@rictic rebase?

rictic:
Sure, but that shouldn't block the review, it's just CHANGELOG :)

src/core/analysis-context.ts
```diff
@@ -498,4 +489,28 @@ export class AnalysisContext {
...
+  /**
+   * A warning for a weird situation that should never happen.
+   *
+   * Before calling getDocument(), which is synchronous, a caller must first
+   * have finished loading and scanning, as those phases are asynchronous.
+   *
+   * So we need to construct a warning, but we don't have a parsed document,   *
+   * so we construct this weird fake one.
+   */
+  private _requestedWithoutLoadingWarning(resolvedUrl: ResolvedUrl) {
```
justinfagnani: could just be a function, not a method

src/core/analysis-context.ts
```diff
@@ -498,4 +489,28 @@ export class AnalysisContext {
...
                                            (url as any as ResolvedUrl);
   }
+
+  /**
+   * A warning for a weird situation that should never happen.
+   *
+   * Before calling getDocument(), which is synchronous, a caller must first
+   * have finished loading and scanning, as those phases are asynchronous.
+   *
+   * So we need to construct a warning, but we don't have a parsed document,   *
```
justinfagnani: trailing `*`

src/core/analysis-context.ts
```diff
@@ -498,4 +489,28 @@ export class AnalysisContext {
...
+   * so we construct this weird fake one.
```
justinfagnani: todo and/or issue if we want to have document-less warnings?

rictic: I think this is working as intended for now. Better to have this awkward function than to make all uses of Warning more awkward or slow.

PR #742, opened 2017-11-02
Use branded url types.

rictic:
This PR introduces three new types:
  - FileRelativeUrl
  - PackageRelativeUrl
  - ResolvedUrl

These three types are entirely compile-time concerns, they're normal strings in every way at runtime.

 - [x] CHANGELOG.md not updated, this is an internal change.

### DIFF
--- a/src/core/analysis-context.ts
+++ b/src/core/analysis-context.ts
@@ -29,6 +29,7 @@ import {JavaScriptParser} from '../javascript/javascript-parser';
 import {NamespaceScanner} from '../javascript/namespace-scanner';
 import {JsonParser} from '../json/json-parser';
 import {Document, InlineDocInfo, LocationOffset, ScannedDocument, ScannedElement, ScannedImport, ScannedInlineDocument, Severity, Warning, WarningCarryingException} from '../model/model';
+import {PackageRelativeUrl, ResolvedUrl} from '../model/url';
 import {ParsedDocument} from '../parser/document';
 import {Parser} from '../parser/parser';
 import {BehaviorScanner} from '../polymer/behavior-scanner';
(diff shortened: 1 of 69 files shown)


src/core/analysis-context.ts
```diff
@@ -224,16 +225,16 @@ export class AnalysisContext {
     }
     const scannedDocument = this._cache.scannedDocuments.get(resolvedUrl);
     if (!scannedDocument) {
-      return <Warning>{
+      return {
         sourceRange: {
-          file: this.resolveUrl(url),
+          file: resolvedUrl,
```
usergenic: Is the Warning interface not capable of matching this because the branded string is given as the `file` property here in the `sourceRange` or was there a different reason for using `as any` instead of `as Warning`?

src/core/analysis-context.ts
```diff
@@ -484,15 +484,16 @@ export class AnalysisContext {
...
 
   /**
    * Resolves a URL with this Analyzer's `UrlResolver` or returns the given
    * URL if it can not be resolved.
    */
-  resolveUrl(url: string): string {
-    return this.resolver.canResolve(url) ? this.resolver.resolve(url) : url;
+  resolveUrl(url: PackageRelativeUrl): ResolvedUrl {
+    return this.resolver.canResolve(url) ? this.resolver.resolve(url) :
+                                           (url as any as ResolvedUrl);
```
usergenic: So the `url as any` before `as ResolvedUrl` is how one de-brands the PackageRelativeUrl string so it can be re-branded as a ResolvedUrl?

src/core/analysis-context.ts
rictic: Good eye, this code is actually just wrong, but I split the fix out into https://github.com/Polymer/polymer-analyzer/pull/741

src/core/analysis-context.ts
rictic: Yes, exactly.

PR #743, opened 2017-11-04
Handle method rest parameters.

aomarks:
Before this change, rest parameters (e.g. `foo(...args)`) would have the `...` included in their name, and would not include any information from their JSDoc annotation.

Now the parameter name does not include the `...`, and the JSDoc annotations are correctly linked (note that the *type* includes the `...` to make it clear this is a rest parameter). This is how regular functions already work in Analyzer

### DIFF
--- a/src/test/static/class/class-methods.js
+++ b/src/test/static/class/class-methods.js
@@ -46,4 +46,14 @@ class Class {
   customInstanceFunctionWithParamsAndPrivateJSDoc() {
     return 8;
   }
+
+  /**
+   * This is the description for customInstanceFunctionWithRestParam.
+   * @param {Number} a The first argument.
+   * @param {...Number} b The second argument.
+   * @returns {Number} - The number 9, always.
+   */
+  customInstanceFunctionWithRestParam(a, ...b) {
+    return 9;
+  }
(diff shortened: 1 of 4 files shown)


src/test/static/class/class-methods.js
```diff
@@ -46,4 +46,14 @@ class Class {
   customInstanceFunctionWithParamsAndPrivateJSDoc() {
     return 8;
   }
+
```
rictic: In this file, we should probably be using `number` rather than the weird `Number` type.

src/test/static/class/class-methods.js
```diff
@@ -46,4 +46,14 @@ class Class {
...
+  /**
+   * This is the description for customInstanceFunctionWithRestParam.
+   * @param {Number} a The first argument.
+   * @param {...Number} b The second argument.
```
rictic: I had to verify that `...number` is correct, rather than typescript's `number[]`.

rictic:
I merged the CHANGELOG.md conflict through the github UI. Is that kosher, or annoying?

aomarks:
Don't see why not

PR #745, opened 2017-11-06
Add defaultValue and rest fields to method params.

aomarks:
Also builds method parameters in a generally more principled way, by unpacking the AST instead of relying on the escodegen library.

### DIFF
--- a/src/javascript/esutil.ts
+++ b/src/javascript/esutil.ts
@@ -18,7 +18,7 @@ import * as escodegen from 'escodegen';
 import * as estraverse from 'estraverse';
 import * as estree from 'estree';
 
-import {ScannedMethod} from '../index';
+import {ScannedMethod, MethodParam} from '../index';
 import {ImmutableSet} from '../model/immutable';
 import {Privacy} from '../model/model';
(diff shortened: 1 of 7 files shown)


src/javascript/esutil.ts
```diff
@@ -251,16 +251,35 @@ export function toScannedMethod(
...
+        // Rest parameter: method(...param)
+        name = nodeParam.argument.name;
+
+      } else if (
+          nodeParam.type === 'AssignmentPattern' &&
+          nodeParam.left.type === 'Identifier' &&
+          nodeParam.right.type === 'Literal') {
+        // Parameter with a default: method(param = "default")
+        name = nodeParam.left.name;
+        defaultValue = String(nodeParam.right.value);
```
rictic: default values can be a lot of things. `[]`, `{}`, `new Set()` are all common and very reasonable values. I'd say just use escodegen to get a string representation

aomarks: Done.

aomarks:
Just pushed a commit that adds an explicit `rest` field to params too.

PR #749, opened 2017-11-10
Switched from estree/espree to babylon/babel for parsing JavaScript.

usergenic:
Converted Analyzer to use babel/babylon instead of estree/espree 

All tests pass!

That doesn't mean everything is identical... A good summary of the differences in the resultant ASTs between estree and babel is  https://github.com/lightscript/babel-to-estree#deviations-addressed

I haven't added *new* tests to demonstrate *new* behaviors as this PR is meant to convert the underlying parser with as little downstream pain as possible.


justinfagnani:
I think this should be a breaking change and cause the version to go to 3.0, so let's tag and release a 2.x before this PR is merged with the unreleased changes that are queued up.

rictic:
@justinfagnani https://github.com/Polymer/polymer-analyzer/pull/755

src/javascript/function-scanner.ts
```diff
@@ -31,7 +31,12 @@ export class FunctionScanner implements JavaScriptScanner {
       visit: (visitor: Visitor) => Promise<void>) {
     const visitor = new FunctionVisitor(document);
     await visit(visitor);
-    return {features: Array.from(visitor.functions)};
+    return {
+      features: Array.from(visitor.functions)
+                    .sort(
+                        (a, b) => compareSourceLocationStart(
+                            a.sourceRange, b.sourceRange))
```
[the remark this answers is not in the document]
usergenic: Ack, you're right.  I messed up the types.  I have deleted the compareSourceLocationStart function from `ast-value.ts` and added a `comparePosition` function to `source-range.ts` and now use that directly.  Thanks!

src/javascript/javascript-parser.ts
```diff
@@ -166,3 +167,24 @@ export function parseJs(
...
+  if (!err.message) {
+    return;
+  }
+  const lineAndColumnMatch =
+      err.message.match(/(Unexpected token.*)\(([0-9]+):([0-9]+)\)/);
+  if (!lineAndColumnMatch) {
+    return;
+  }
+  err.lineNumber = parseInt(lineAndColumnMatch[2]);
+  err.column = parseInt(lineAndColumnMatch[3]) + 1;
```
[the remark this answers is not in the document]
usergenic: Oh right!  Thanks.  Radix'd!

src/javascript/estraverse-shim.ts
```diff
@@ -0,0 +1,41 @@
...
+import * as babel from 'babel-types';
+import {Visitor, VisitResult} from './estree-visitor';
+
+export enum VisitorOption {
+  Skip = 'skip',
+  Break = 'break',
+  Remove = 'remove',
+}
+
+export function traverse(ast: babel.Node, visitor: Visitor): void {
```
justinfagnani: comments

src/javascript/estraverse-shim.ts
```diff
@@ -0,0 +1,41 @@
...
+
+function dispatchVisitMethods(
+    methods: string[], path: NodePath<babel.Node>, visitor: Visitor): void {
+  for (const method of methods) {
+    if (typeof(<any>visitor)[method] === 'function') {
+      const result =
+          (<any>visitor)[method](path.node, path.parent) as VisitResult;
+      switch (result) {
+        case VisitorOption.Break:
+          return path.stop();
```
justinfagnani: We're going to need to be careful if we support batching of babel-traverse visitors. We'll need to wrap the path so that one visitor can't stop another, or just get rid of batching.

src/javascript/esutil.ts
```diff
@@ -100,42 +101,41 @@ const VALID_EXPRESSION_TYPES = new Map([
...
+    node: babel.Node, sourceRange: SourceRange, document: ParsedDocument):
     string|Warning {
   const type = VALID_EXPRESSION_TYPES.get(node.type);
   if (type) {
     return type;
-  } else if (node.type === 'Literal') {
-    return typeof node.value;
-  } else if (node.type === 'Identifier') {
+  } else if (babel.isLiteral(node)) {
+    return typeof astValue.expressionToValue(node);
```
justinfagnani: this is kind of funny to me in it's original form... seems like this should be handled in `VALID_EXPRESSION_TYPES`

src/javascript/esutil.ts
```diff
@@ -171,26 +171,29 @@ function getLeadingComments(node: estree.Node): string[]|undefined {
 }
 
 export function getPropertyValue(
-    node: estree.ObjectExpression, name: string): estree.Node|undefined {
+    node: babel.ObjectExpression, name: string): babel.Node|undefined {
   const properties = node.properties;
   for (const property of properties) {
-    if (objectKeyToString(property.key) === name) {
+    // TODO(usergenic): Can't get property key and value from a SpreadProperty.
```
justinfagnani: let's start filing issues for TODOs

justinfagnani: We can do a little of this with scope analysis. If we can resolve the expression to a literal we can get the keys/values from the literal.

src/javascript/esutil.ts
```diff
@@ -171,26 +171,29 @@ function getLeadingComments(node: estree.Node): string[]|undefined {
...
       return property.value;
     }
   }
 }
 
-export function isFunctionType(node: estree.Node): node is estree.Function {
-  return node.type === 'ArrowFunctionExpression' ||
-      node.type === 'FunctionExpression' || node.type === 'FunctionDeclaration';
+export function isFunctionType(node: babel.Node): node is babel.Function {
+  return babel.isArrowFunctionExpression(node) ||
```
justinfagnani: Is there a more general `isFunction(node)`?

src/javascript/esutil.ts
```diff
@@ -171,26 +171,29 @@ function getLeadingComments(node: estree.Node): string[]|undefined {
...
+      babel.isFunctionExpression(node) || babel.isFunctionDeclaration(node);
 }
 
 
 /**
  * Create a ScannedMethod object from an estree Property AST node.
  */
 export function toScannedMethod(
-    node: estree.Property|estree.MethodDefinition,
+    node: babel.ObjectProperty|babel.ObjectMethod|babel.ClassMethod,
```
justinfagnani: Is there a more general interface?

src/javascript/esutil.ts
```diff
@@ -208,7 +211,27 @@ export function toScannedMethod(
       parsedDocument: document
     }));
   }
-  let type = closureType(node.value, sourceRange, document);
+
+  let value;
```
justinfagnani: const and ternary?

src/javascript/esutil.ts
```diff
@@ -208,7 +211,27 @@ export function toScannedMethod(
...
+  if (babel.isClassMethod(node)) {
+    value = node;
+  } else {
+    value = node.value;
+  }
+
+  let type;
```
justinfagnani: `type` is a type or warning? Was is that way before?

src/javascript/esutil.ts
```diff
@@ -229,9 +252,9 @@ export function toScannedMethod(
     privacy: getOrInferPrivacy(name, parsedJsdoc)
   };
 
-  const value = node.value;
-  if (value.type === 'FunctionExpression' ||
-      value.type === 'ArrowFunctionExpression') {
+  if (value &&
+      (babel.isFunctionExpression(value) ||
```
justinfagnani: If babel doesn't have a more general function predicate, we might want to make one so we have one canonical check and don't miss a node type.

src/javascript/function-scanner.ts
```diff
@@ -31,7 +31,12 @@ export class FunctionScanner implements JavaScriptScanner {
       visit: (visitor: Visitor) => Promise<void>) {
     const visitor = new FunctionVisitor(document);
     await visit(visitor);
-    return {features: Array.from(visitor.functions)};
+    return {
+      features: Array.from(visitor.functions)
+                    .sort(
+                        (a, b) => compareSourceLocationStart(
```
justinfagnani: You could to `.sort(compareSourceLocationStart)` if `compareSourceLocationStart` took features not SourceLocations.

src/javascript/function-scanner.ts
```diff
@@ -83,9 +96,14 @@ class FunctionVisitor implements Visitor {
   /**
    * Scan functions defined inside of object literals.
    */
-  enterObjectExpression(node: estree.ObjectExpression, _parent: estree.Node) {
+  enterObjectExpression(node: babel.ObjectExpression, _parent: babel.Node) {
     for (let i = 0; i < node.properties.length; i++) {
       const prop = node.properties[i];
+      // TODO(usergenic): Can't get value from SpreadProperty.  Is it right to
+      // skip it here?
```
justinfagnani: issue?

src/javascript/javascript-document.ts
```diff
@@ -159,13 +160,16 @@ export class JavaScriptDocument extends ParsedDocument<Node, Visitor> {
   stringify(options: StringifyOptions) {
     options = options || {};
     const formatOptions = {
-      comment: true,
-      format: {indent: {style: '  ', adjustMultilineComment: true, base: 0}}
+      comments: true,
+      retainLines: false,
+      quotes: 'single' as 'single',
+      // TODO(usergenic): babel-generator does not have anything like
```
justinfagnani: file issue and some punctuation problems here

src/javascript/javascript-parser.ts
```diff
@@ -30,13 +30,25 @@ declare class SyntaxError {
...
-  comment: true,
-  loc: true,
+export const baseParseOptions: babylon.BabylonOptions = {
+  plugins: [
+    'asyncGenerators',
+    'dynamicImport',
+    // 'importMeta', // not yet in the @types file
+    'objectRestSpread',
+  ],
+  // TODO(usergenic): None of these parsing options are supported by Babylon.
```
justinfagnani: issue

src/javascript/namespace-scanner.ts
```diff
@@ -61,22 +61,33 @@ class NamespaceVisitor implements Visitor {
    * Look for object assignments with @namespace in the docs.
    */
   enterAssignmentExpression(
-      node: estree.AssignmentExpression, parent: estree.Node) {
+      node: babel.AssignmentExpression, parent: babel.Node) {
     this._initNamespace(parent, node.left);
   }
 
-  enterProperty(node: estree.Property, _parent: estree.Node) {
+  // enterProperty(node: babel.Property, _parent: babel.Node) {
```
justinfagnani: remove?

src/polymer/analyze-properties.ts
```diff
@@ -34,14 +34,19 @@ import {ScannedPolymerProperty} from './polymer-element';
...
+    node: babel.Node, document: JavaScriptDocument): ScannedPolymerProperty[] {
   const analyzedProps: ScannedPolymerProperty[] = [];
 
-  if (node.type !== 'ObjectExpression') {
+  if (!babel.isObjectExpression(node)) {
     return analyzedProps;
   }
 
   for (const property of node.properties) {
+    // TODO(usergenic): SpreadProperty doesn't have value so I have to skip
```
justinfagnani: issue. since this pops up a lot, it'll be nice to be able to search for the issue number in source.

src/polymer/behavior-scanner.ts
```diff
@@ -266,29 +271,27 @@ class BehaviorVisitor implements Visitor {
...
-      // super general. this code is suspicious.
-      return (<any>node).expression.right;
-    case 'VariableDeclaration':
-      return node.declarations.length > 0 ? node.declarations[0].init :
-                                            undefined;
+function behaviorExpression(node: babel.Node): babel.Node|null|undefined {
+  if (babel.isExpressionStatement(node)) {
+    // need to cast to `any` here because ExpressionStatement is super
+    // super general. this code is suspicious.
+    return (<any>node).expression.right;
```
justinfagnani: prefer `as any`. it'll be easier to search for that dangerous cast

src/polymer/expression-scanner.ts
```diff
@@ -161,10 +161,10 @@ export abstract class DatabindingExpression {
...
 
   private _extractAndValidateSubExpression(
-      expression: estree.Node, callAllowed: boolean): void {
-    if (expression.type === 'UnaryExpression' && expression.operator === '-') {
-      if (expression.argument.type !== 'Literal' ||
-          typeof expression.argument.value !== 'number') {
+      expression: babel.Node, callAllowed: boolean): void {
+    if (babel.isUnaryExpression(expression) && expression.operator === '-') {
+      if (!babel.isLiteral(expression.argument) ||
+          typeof astValue.expressionToValue(expression.argument) !== 'number') {
```
justinfagnani: can't this check the type of the node?

src/polymer/js-utils.ts
```diff
@@ -45,7 +45,30 @@ export function toScannedPolymerProperty(
       parsedDocument: document
     }));
   }
-  let type = closureType(node.value, sourceRange, document);
+
+  let value;
```
justinfagnani: This looks the same as code above. Refactor to a function.

src/polymer/polymer-element-scanner.ts
```diff
@@ -159,8 +160,12 @@ class ElementVisitor implements Visitor {
               this.document, elementObject, 'callExpression');
           element.warnings = element.warnings.concat(parseResult.warnings);
           let expressionText = undefined;
-          if (elementObject.type === 'Literal') {
-            expressionText = elementObject.raw;
+          if (babel.isLiteral(elementObject)) {
+            // TODO(usergenic): How do I get equivalent of `raw` from Babylon
```
justinfagnani: issue

usergenic: Ah yes I forgot about this one.  We have zero test-coverage for the `expressionText` returned in the observers set, so I'm not sure what we're expecting here.  (I tried changing the return value to 'potato' and everything still passed...)

@rictic @justinfagnani can you weigh in on what the intent of `expressionText` is before I file the issue and/or add test and correct this bit?

usergenic: While tests don't cover, I ran experiments and found out `.raw` is essentially just the excerpt of the document from start to end.  So, I modified elementObject.raw to be this.document.contents.slice(elementObject.start, elementObject.end) and that returned identical content.  No need to file issue as this was the only occurrence of `.raw` being used anyways.

usergenic:
@justinfagnani addressed comments.  filed a couple issues.  solved `.raw` well enough.

src/polymer/polymer-element-scanner.ts
rictic: This looks correct to me. If there's an issue it'll come up in the linter's tests when we update it to analyzer @3.0.0 and we can fix it then, but this is very likely correct.

src/javascript/ast-value.ts
```diff
@@ -28,6 +28,8 @@ function literalToValue(literal: babel.Literal): LiteralValue {
   if (babel.isNullLiteral(literal)) {
     return null;
   }
+  // Any other literal value is treated as undefined.
+  return undefined;
```
rictic: This is because `undefined` is what this function returns when it can't get a value, because it's from the lossy-goosy old days of the analyzer. If we wrote it today we'd return something with a more structured differentiation between intended results and failure

src/javascript/estraverse-shim.ts
```diff
@@ -1,14 +1,49 @@
...
 import babelTraverse from 'babel-traverse';
 import {NodePath} from 'babel-traverse';
 import * as babel from 'babel-types';
 import {Visitor, VisitResult} from './estree-visitor';
 
+/**
+ * These enum options mimic the estraverse enums that are returned by their
+ * `enterX`/`leaveX` visitor methods to advise flow of the visitor.
+ */
 export enum VisitorOption {
```
rictic: IIRC, almost none of our code uses this stuff, so it can be pretty easily removed.

src/polymer/behavior-scanner.ts
```diff
@@ -273,7 +273,7 @@ function behaviorExpression(node: babel.Node): babel.Node|null|undefined {
   if (babel.isExpressionStatement(node)) {
     // need to cast to `any` here because ExpressionStatement is super
     // super general. this code is suspicious.
-    return (<any>node).expression.right;
+    return (node as any).expression.right;
```
rictic: man this is gonna crash someday

src/polymer/js-utils.ts
```diff
@@ -46,27 +46,9 @@ export function toScannedPolymerProperty(
...
-      code: 'unknown-prop-type',
-      message:
-          `Could not determine type of property from expression of type: ` +
-          `${node.key.type}`,
-      sourceRange: sourceRange,
-      severity: Severity.INFO,
-      parsedDocument: document
-    });
-  }
+  let type = closureType(value, sourceRange, document);
```
rictic: Why is this warning unnecessary?

src/polymer/js-utils.ts
```diff
@@ -46,27 +46,9 @@ export function toScannedPolymerProperty(
...
-  } else {
-    value = node.value;
-  }
+  const value = babel.isObjectProperty(node) ? node.value : node;
 
-  let type;
-  if (value) {
-    type = closureType(value, sourceRange, document);
-  } else {
-    type = new Warning({
```
rictic: Was this warning never created? Warnings are good things when we encounter invalid code, I'd like us to tend towards more of them rather than fewer!

rictic:
Damn it github. I was intending to comment on outdated code! Don't just hide those comments.

rictic:
We should use Reviewable, github reviews eat data.

src/javascript/ast-value.ts
usergenic: 👍

src/javascript/estraverse-shim.ts
usergenic: Eventually.  We do use a few VisitorOption.Skip returns.

src/polymer/js-utils.ts
usergenic: This diff might be misleading -- Master [doesn't have this](https://github.com/Polymer/polymer-analyzer/blob/master/src/polymer/js-utils.ts#L48), it was something I added because I had a scenario where node.value was undefined, but that scenario was because I didn't understand that the new ClassMethod and ObjectMethod both don't have value and that it should only be pulled from ObjectProperty.  After that, I was able to always have a defined node to send to closureType and no longer needed to pre-empt a call to it with my own Warning.  If you look at master today, the warning is not there.

src/polymer/js-utils.ts
usergenic: See note below.  This was a warning I added that turned out to not be a thing needed.

src/javascript/esutil.ts
usergenic: I totally extended VALID_EXPRESSION_TYPES to do this now.

src/polymer/analyze-properties.ts
usergenic: I added issue.  All we have to do is search for SpreadProperty.

src/test/javascript/javascript-parser_test.ts
```diff
@@ -33,12 +35,14 @@ suite('JavaScriptParser', () => {
   suite('parse()', () => {
 
     test('parses classes', () => {
+      // TODO(usergenic): I had to modify this test fixture because Babylon
+      // doesn't appreciate the keyword abuse of `const let = ...`.
```
rictic: This TODO can be removed. `const let` is an early error it seems, so babel's doing the right thing here.

rictic:
Also WDYT about merging this into a 3.0 branch where we can land some other API changes?

rictic:
The linter was pretty quick and easy.

rictic:
Editor service was even easier, just three hardcoded parse error messages in tests that needed to be updated. 

Really great work, to make such a huge change with such a high degree of compatibility.

PR #754, opened 2017-11-15
Upstream edits and fixes from the linter.

rictic:


src/model/warning.ts
```diff
@@ -14,24 +14,40 @@
...
 
+  /**
+   * If the problem has a single automatic fix, this is it.
+   *
+   * Whether and how much something is 'automatic' can be a bit tricky to
+   * delineate. Roughly speaking, if 99% of the time the change solves the
+   * issue completely then it should go in `fix`.
+   */
+  readonly fix: Edit|undefined;
+  readonly actions: ReadonlyArray<Action>|undefined;
```
justinfagnani: document `actions`

src/model/warning.ts
```diff
@@ -14,24 +14,40 @@
...
   readonly severity: Severity;
 
+  /**
+   * If the problem has a single automatic fix, this is it.
+   *
+   * Whether and how much something is 'automatic' can be a bit tricky to
+   * delineate. Roughly speaking, if 99% of the time the change solves the
+   * issue completely then it should go in `fix`.
+   */
+  readonly fix: Edit|undefined;
```
justinfagnani: why not `fix?`?

src/model/warning.ts
```diff
@@ -150,3 +164,197 @@ const defaultPrinterOptions = {
   verbosity: 'full' as 'full',
   color: true
 };
+
+
+export type Action = EditAction | {
+  /** To ensure that type safe code actually checks for the action kind. */
+  kind: 'dummy-action-kind';
```
justinfagnani: `never`?

src/model/warning.ts
```diff
@@ -150,3 +164,197 @@ const defaultPrinterOptions = {
...
+}
+
+/**
+ * Takes the given edits and, provided there are no overlaps, applies them to
+ * the contents loadable from the given loader.
+ *
+ * If there are overlapping edits, then edits earlier in the array get priority
+ * over later ones.
+ */
+export async function applyEdits(
```
justinfagnani: why does the analyzer have code to apply edits?

rictic: It's useful for multiple downstream projects. The CLI and the IDE.

I suppose both depend on the linter as well as the analyzer, but I feel like the applyEdits function is closely tied to the meaning of an Edit as an atomic unit of change.

Also the only legible way to test edits is to apply them and assert on the result. We'll probably want to add some fixes and edits in the core analyzer scanners, and we'll want to test them too.

src/model/warning.ts
rictic: I was thinking about this style for consistent object shape. It doesn't actually assert that we assign to the property in the constructor though so it doesn't actually matter though.

Happy to change to `fix?`, I guess I'm just holding a torch for https://github.com/Microsoft/TypeScript/issues/8476

src/model/warning.ts
rictic: Huh, interesting thought. It looks like that doesn't work well with discriminated unions:

https://www.typescriptlang.org/play/#src=type%20Union%20%3D%20Foo%20%7C%20Bar%3B%0D%0Ainterface%20Foo%20%7B%0D%0A%20%20%20%20kind%3A%20'foo'%3B%0D%0A%20%20%20%20foo%3A%20number%3B%0D%0A%7D%0D%0Ainterface%20Bar%20%7B%0D%0A%20%20%20%20kind%3A%20never%3B%0D%0A%20%20%20%20bar%3A%20string%3B%0D%0A%7D%0D%0A%0D%0Adeclare%20let%20union%3A%20Union%3B%0D%0A%0D%0Aif%20(union.kind%20%3D%3D%3D%20'foo')%20%7B%0D%0A%20%20%20%20console.log(union.foo)%3B%0D%0A%7D%20else%20%7B%0D%0A%20%20%20%20console.log(union.bar)%3B%0D%0A%7D

src/model/warning.ts
rictic: +1 done

src/model/warning.ts
justinfagnani: I actually meant `'never'` :)

src/model/warning.ts
justinfagnani: OK

src/model/warning.ts
justinfagnani: without an initializer the emitted code is the same, and I find `?` easier to read by a hair. If you care about share you need add `= undefined`

src/model/warning.ts
rictic: ah quote escaping. the cause of and s̸̨̠̻͚̲̫̜̞̫̝̮ò̷̖̲̦̟͎͙̜̖̤̝̤̣̥͠l̴̷̡̫̲̤̪̬̙̟̀u̵̡҉͖̭̝̥̺͍̺̮͚̣͔̘͍̹͈͞͠t̢̝͔̗͙̰̝͎͚̗͎̱͞i҉̖͖̠͈͡ͅo̸̧̞̩̻͓̹̞̣̻̖̫̼̥̲̭͉͇̗̩͠͡ͅņ̷̦̞͖̮͓͍̪̳̭͙͝ ̛̥͚̥̟͉̩̯͚̳̰͉̻̦̱͖̦̮͘t̡̩̬̣̕o̷͙̥̞̰̪̪̘͚̝̬͇͢͠ ̵͏̶̪̟̦̭̙͇̹͇̼̲͟ą̕͢͏̷̪͚̰͉͖͔̭͕̝̝̥l̝̼͉͓̬̣̼̦̬͟͠l̶͍̼̹̙͈̼̭̺͈̞̞̜͇͖̀͟͜ ͏͎̩̘̱͕͖̳͍̥̼͈̖͓̜͢ͅó̞̬̙̜̬̲̝̺̬̻͖̘̗̖͉̱͍̜̕͠ͅf̴̨̧̗̲̞̱̤͔̮͍̮̻̲̼̜͈̕͞ ̴̼̰͚̯̝̪̺̳̖̗̤͈̮̹̯̲̹͘͢o̸̫͈̳͍̭̖͇̞̱̻͖̩͢u͎͈̠͓̜͇̩̯̖̬͎͚̣͓̮̰̠͇͢͞r̢̧͞͏̡̺͎̳̘͙̘̖ ̸̴̶̰̭̠̙̘͕̜̙̘̹̜̰̥̦̗͎̀ͅ7̴̧͞͠͏̪̺̬̹1̢̠͈͕̥͙̻̲͘͘͝͡ͅc҉̢̨̗̥͈̯̫͈̲̟̻͎͇̠̟̗̝̙͟ͅ3̧̕҉̜̼̱̮̩̮̞̼̭͓̻̳̫̗̪̀a̫̻̤̳̮̣̟̺͟b̸̧̻͙̙̰̙̣̤̰̳̺̼͠e̷͍͇͈̱͍̝͕͎̮͔̜͝c̸̸̨͇͍͕̠͔͇̳̝͉͓͖̖̙͠f̵̭̺̝͇͟͞g̨͈͖̮͇͔̥̱̤͔̪̺̮̦͟͠͝b̶̛̪̣̯̝͉̣̠͈͖̞͎̼͖͎͍̯͎͘͢͠1̷̶̱͈̖̲̜̤͎̠͕̮̹̪͕̥͉̠͟͠c̢͜҉̧̠̗͉̦̪ȩ̱̯̹̭̪̣̪͚͇͓͡c҉͍̦̫͈̞͉͓̬͞g̷̝̪̘̼͇͈͍̘̩̝͇̝̳͓͙͜3̶̡̢͖͙̲͇͔̝̥̞̕f̸̲̞̣̤̟̝̦̝͓̟͈̩͍̼͍̟͡c̷̴̨͙̖͙͓̩̦̦̯͕͙͎̥͈͙̮̦̀1̵̹̭̠̥͉̠̣̘̻̠b̢̙͖̹̘͍̭̪͉̪̙͍̮f͏̺̣̺̣͚͔̱̀2̶̸̷̢͕͕̗̲̦̤̺̠̞͉̟̭͇̭̘̜̭̤͕5͕͇̱͍͖̺́͠b̶̶̢͉̼͔̘̗̰̮̰̭͞͞6̡̙̳̼̥́4̩̭̥̺̠̣͓͔̠̺̟̞̻̥̲͚̩̖̕͟1̖̥̙͓̣̱̤̝͕̯̥̬̣̙̼̦͉̱͈̀͢͟6̷͘҉͓̪̭͇̖̮̰̺͚̰̳̦͕͙̝̙͜0̛̭̼̗̮̣̣̟̝̘̣̲̯͎̦̙̰ͅ9̡͏̧̛̣̗̮͙̲̥̗̩̰̘͕̫̪̠̭̼̯2҉̢̢͙͙̝͓̭̺̀ͅ4̨͓̗͓̗̳̗̲͚̰̞̝͇̯̠͞͠8̵͔̠̘̹͖̱̬̭͚͈͇̫͓̤̼͜

src/model/warning.ts
```diff
@@ -150,3 +175,197 @@ const defaultPrinterOptions = {
...
+      const leftEdgeComp =
+          comparePositionAndRange(b.range.start, a.range, true);
+      if (leftEdgeComp !== 0) {
+        return leftEdgeComp;
+      }
+      return comparePositionAndRange(b.range.end, a.range, false);
+    });
+    for (const replacement of replacements) {
+      const offsets = document.sourceRangeToOffsets(replacement.range);
+      contents = contents.slice(0, offsets[0]) + replacement.replacementText +
```
justinfagnani: soo bad we don't have `String#splice()`

src/model/warning.ts
```diff
@@ -150,3 +175,197 @@ const defaultPrinterOptions = {
...
+    }
+  }
+
+  // Ok, we can be applied to the replacements, so add our replacements in.
+  for (const replacement of edit) {
+    if (!replacements.has(replacement.range.file)) {
+      replacements.set(replacement.range.file, [replacement]);
+    } else {
+      const fileReplacements = replacements.get(replacement.range.file)!;
+      // TODO(rictic): insert in sorted order
```
justinfagnani: remove TODO? the later sort seems fine

src/model/warning.ts
rictic: Yeah, this was surprisingly awkward.

PR #760, opened 2017-11-19
Improve analysis-cache.toString()

rictic:
- [x] CHANGELOG.md has not updated, minor internal change.

Note to future self: the bug you're looking for probably isn't in the analysis cache. It's probably much simpler and dumber than that.

### DIFF
--- a/src/core/analysis-cache.ts
+++ b/src/core/analysis-cache.ts
@@ -113,11 +113,23 @@ export class AnalysisCache {
   }
 
   toString() {
-    return `<AnalysisCache
-        scannedDocuments:
-            ${Array.from(this.scannedDocuments.keys()).join('\n            ')}
-        analyzedDocuments:
-            ${Array.from(this.analyzedDocuments.keys()).join('\n            ')}
-      >`;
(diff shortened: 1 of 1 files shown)


src/core/analysis-cache.ts
```diff
@@ -113,11 +113,23 @@ export class AnalysisCache {
...
+    }
+    if (this.analyzedDocuments.size > 0) {
+      result += `
+  analyzedDocuments:
+    ${Array.from(this.analyzedDocuments.keys()).join('\n    ')}`;
+    }
+    if (this.failedDocuments.size > 0) {
+      result += `
+  failedDocuments:
+    ${Array.from(this.analyzedDocuments.keys()).join('\n    ')}`;
```
usergenic: TFW you run with this and then you are sure you don't have a weird analysis-cache bug anymore but the files returned in toString are the wrong ones...

`this.analyzedDocuments.keys()` -> `this.failedDocuments.keys()`

rictic: Oh man. Good catch.

That would've been a painful one to debug.

PR #762, opened 2017-11-20
One indexed

rictic:

Fixes https://github.com/Polymer/polymer-linter/issues/115


rictic:
Note that this is merging into the 3.0 branch, as it might as well be bundled up with other breaking changes we're doing right now.

PR #767, opened 2017-11-27
Indirect url resolver

rictic:


justinfagnani:
Should the be rebased on master now?

rictic:
Rebased, tested, ready to review.

src/url-loader/fs-url-resolver.ts
```diff
@@ -0,0 +1,142 @@
...
+import {parseUrl} from '../core/utils';
+import {FileRelativeUrl, PackageRelativeUrl, ScannedImport} from '../index';
+import {ResolvedUrl} from '../model/url';
+
+import {UrlResolver} from './url-resolver';
+
+const isWindows = process.platform === 'win32';
+
+/**
+ * Resolves package-relative urls to a configured root directory.
```
justinfagnani: Can you describe this a little more? Does this do any mapping? What's it do with URLs that are already `file:`?

src/url-loader/fs-url-resolver.ts
```diff
@@ -0,0 +1,142 @@
...
+    }
+    const url = parseUrl(resolvedHref);
+    if (this.shouldHandleAsFileUrl(url)) {
+      return this.handleFileUrl(url, unresolvedHref);
+    }
+    return this.brandAsResolved(resolvedHref);
+  }
+
+  protected shouldHandleAsFileUrl(url: Url) {
+    if (this.hostname !== undefined && url.hostname === this.hostname) {
```
justinfagnani: don't we want to use the whole `host`? For the purposes of URLs as identifiers, same `hostname` but different `port` might as well be a different domain.

src/url-loader/fs-url-resolver.ts
```diff
@@ -0,0 +1,142 @@
...
+
+  /**
+   * Take the given URL which is either a file:// url or a url with the
+   * configured hostname, and treat its pathname as though it points to a file
+   * on the local filesystem, producing a file:/// url.
+   *
+   * Also corrects sibling URLs like `../foo` to point to
+   * `./${component_dir}/foo`
+   */
+  private handleFileUrl(url: Url, unresolvedHref: string) {
```
justinfagnani: prefix private member names with `_`

src/url-loader/fs-url-resolver.ts
```diff
@@ -0,0 +1,142 @@
...
+
+import {UrlResolver} from './url-resolver';
+
+const isWindows = process.platform === 'win32';
+
+/**
+ * Resolves package-relative urls to a configured root directory.
+ */
+export class FsUrlResolver extends UrlResolver {
+  protected readonly packageDir: string;
```
justinfagnani: Is this a directory of packages, or the directory for a special package (like a root package for analysis)?

src/url-loader/fs-url-resolver.ts
```diff
@@ -0,0 +1,142 @@
...
+   *
+   * Also corrects sibling URLs like `../foo` to point to
+   * `./${component_dir}/foo`
+   */
+  private handleFileUrl(url: Url, unresolvedHref: string) {
+    let pathname: string;
+    const unresolvedUrl = parseUrl(unresolvedHref);
+    if (unresolvedUrl.pathname && unresolvedUrl.pathname.startsWith('/') &&
+        unresolvedUrl.protocol !== 'file:') {
+      // Absolute urls point to the package root.
```
justinfagnani: Should that behavior be configurable? In a lot of monorepo scenarios, absolute URLs might not make any sense - they may just point to a server resource that's unavailable to us.

src/url-loader/fs-url-resolver.ts
```diff
@@ -0,0 +1,142 @@
...
+                                                 [this.packageUrl, fromOrTo];
+    return this.simpleUrlRelative(from, to);
+  }
+
+  protected filesystemPathForPathname(decodedPathname: string) {
+    return normalizeFsPath(Uri.file(decodedPathname).fsPath);
+  }
+}
+
+function normalizeFsPath(fsPath: string) {
```
justinfagnani: describe the normalization

src/url-loader/indirect-url-resolver.ts
```diff
@@ -0,0 +1,122 @@
...
+
+
+// Declare these as classes rather than interfaces so that the properties
+// can be private.
+declare class RuntimeUrlBrand extends FileRelativeUrlBrand {
+  private RuntimeUrlBrand: never;
+}
+
+/**
+ * A URL resolver for very large codebases where source files map in an
```
justinfagnani: "very large codebases" -> "monorepo" might be more descriptive. I think the key is that this supports many packages in one arbitrary FS layout, rather than requiring dependencies to be installed in a specific place according to a package manager.

justinfagnani: Also, seems like `PackageUrlResolver` is a special case of this, and that it might be easier to think about if we had a `BowerUrlResolver` and `NpmUrlResolver`, rather than configuration in `PackageUrlResolver`

src/url-loader/indirect-url-resolver.ts
```diff
@@ -0,0 +1,122 @@
...
+export class IndirectUrlResolver extends FsUrlResolver {
+  private readonly runtimeUrlToResolvedUrl:
+      ReadonlyMap<RuntimeUrl, ResolvedUrl>;
+  private readonly resolvedUrlToRuntimeUrl:
+      ReadonlyMap<ResolvedUrl, RuntimeUrl>;
+
+  /**
+   * @param rootPath All loadable source code must be a descendent of this
+   *     directory. Should be the same as FsUrlLoader.
+   * @param packagePath The base directory for paths from the user. Usually the
```
justinfagnani: "paths from the user"? I don't parse the specific meaning there. Where to absolute URLs resolve to?

src/url-loader/indirect-url-resolver.ts
```diff
@@ -0,0 +1,122 @@
...
+   *     files on the filesystem. The runtime URLs should all be relative paths
+   *     from the same base url. The filesystem paths should all be relative
+   *     paths from `rootPath`.
+   */
+  constructor(
+      rootPath: string, packagePath: string,
+      indirectionMap: Map<string, string>) {
+    super(packagePath);
+
+    const rootResolver = new FsUrlResolver(rootPath);
```
justinfagnani: Can you describe this operation here, and/or put it in a named method?

src/url-loader/indirect-url-resolver.ts
```diff
@@ -0,0 +1,122 @@
...
+    const urlspaceToFilesystem = new Map<RuntimeUrl, ResolvedUrl>();
```
justinfagnani: not sure you really need the locals

src/url-loader/indirect-url-resolver.ts
```diff
@@ -0,0 +1,122 @@
...
+      urlspaceToFilesystem.set(url, fsUrl);
+      filesystemToUrlspace.set(fsUrl, url);
+    }
+    this.runtimeUrlToResolvedUrl = urlspaceToFilesystem;
+    this.resolvedUrlToRuntimeUrl = filesystemToUrlspace;
+  }
+
+  resolve(url: FileRelativeUrl|PackageRelativeUrl, baseUrl?: ResolvedUrl):
+      ResolvedUrl|undefined {
+    // If we're just resolving user input, do the basic thing.
```
justinfagnani: why is not having a baseUrl always user input? Would we never use the package root as a base for user input?

src/url-loader/indirect-url-resolver.ts
```diff
@@ -0,0 +1,122 @@
...
+    if (baseUrl === undefined) {
+      return super.resolve(url);
+    }
+    const runtimeBaseUrl = this.resolvedUrlToRuntimeUrl.get(baseUrl);
+    if (runtimeBaseUrl === undefined) {
+      return super.resolve(url, baseUrl);
```
justinfagnani: why is it ok to delegate to super here? If we can't handle the base, we don't we return `undefined`?

src/url-loader/indirect-url-resolver.ts
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  private runtimeResolve(
+      url: FileRelativeUrl|PackageRelativeUrl,
+      runtimeBaseUrl: RuntimeUrl): RuntimeUrl {
+    return this.simpleUrlResolve(url, this.brandAsResolved(runtimeBaseUrl)) as
+        any as RuntimeUrl;
+  }
+
+  // resolveFileUrl(url: FileRelativeUrl, baseUrl: ResolvedUrl): ResolvedUrl {
```
justinfagnani: remove?

FredKSchott:
@rictic rebase needed

src/test/url-loader/indirect-url-resolver_test.ts
```diff
@@ -0,0 +1,116 @@
+/**
+ * @license
+ * Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
```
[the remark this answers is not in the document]
rictic: 2017, done

src/url-loader/fs-url-resolver.ts
rictic: clarified

src/url-loader/fs-url-resolver.ts
rictic: clarified

src/url-loader/fs-url-resolver.ts
rictic: Quoting from the Google TypeScript Style Guide, as I agree with them here :)

```
*_ prefix/suffix*: Identifiers must not use _ as a prefix or suffix. The exception is TypeScript code that is used by JavaScript programmers, e.g. code published on npm. Such code may use _ to mark private properties for outside users not using a compiler.
```

This code is deep within the internals of the analyzer, and I think it's fine not to prefix it with an underscore.

Otherwise every other field and method would have underscores.

src/url-loader/fs-url-resolver.ts
rictic: Perhaps, but punting this for now.

src/url-loader/fs-url-resolver.ts
rictic: Done

src/url-loader/indirect-url-resolver.ts
rictic: Agreed, but would like to punt this

src/url-loader/indirect-url-resolver.ts
rictic: Clarified

src/url-loader/indirect-url-resolver.ts
rictic: The fields are ReadonlyMaps, documenting that they're not changed after creation.

src/url-loader/indirect-url-resolver.ts
rictic: Described. I also did my best to use types to help describe this. Lemmy know if it's still unclear. Some of it is just tricksy, as there's a number of closely related concepts that we're translating between.

src/test/url-loader/indirect-url-resolver_test.ts
```diff
@@ -0,0 +1,115 @@
...
+  const mapping = new Map<string, string>([
+    ['/foo/foo.html', 'sub/package/foo/foo.html'],
+    ['/foo/foo.css', 'sub/package/foo/foo.css'],
+    ['/bar/bar.html', 'different/x/y/bar.html'],
+    ['/bar/bar.css', 'different/x/y/bar.css'],
+  ]);
+  const indirectResolver =
+      new IndirectUrlResolver('/root', '/root/sub/package', mapping);
+
+  test('the full use case', async () => {
```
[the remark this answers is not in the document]
justinfagnani: This is more of an integration test too, consider putting in a separate file?

src/url-loader/indirect-url-resolver.ts
```diff
@@ -0,0 +1,122 @@
...
+ */
+export class IndirectUrlResolver extends FsUrlResolver implements UrlResolver {
+  private readonly runtimeUrlToResolvedUrl:
+      ReadonlyMap<RuntimeUrl, ResolvedUrl>;
+  private readonly resolvedUrlToRuntimeUrl:
+      ReadonlyMap<ResolvedUrl, RuntimeUrl>;
+
+  /**
+   * @param rootPath All loadable source code must be a descendent of this
+   *     directory. Should be the same as FsUrlLoader.
```
justinfagnani: What does "Should be the same as FsUrlLoader." mean?

src/url-loader/indirect-url-resolver.ts
```diff
@@ -0,0 +1,122 @@
...
+   * @param packagePath The base directory for package-relative paths. Usually
+   * the current working directory.
```
justinfagnani: style nit: indent 4 spaces

justinfagnani: Why would this be the cwd? I would think the resolver configuration would be independent of the process's cwd, especially for the case of a multi-package repo.

src/url-loader/indirect-url-resolver.ts
```diff
@@ -0,0 +1,122 @@
...
+   * @param urlToFilesystemMap Maps the runtime URL space to the paths for those
+   *     files on the filesystem. The runtime URLs should all be relative paths
```
justinfagnani: This doesn't describe the contents of the mapping that precisely to me. Is "the runtime URL space" all the URLs that might be encountered in the monorepo? Are the keys URLs?

src/test/url-loader/indirect-url-resolver_test.ts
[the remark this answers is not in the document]
rictic: Moved to a suite. I don't think we want to add ceremony for integration tests, as most tests should be integration tests.

src/url-loader/indirect-url-resolver.ts
rictic: Clarified

src/url-loader/indirect-url-resolver.ts
rictic: Don't get super caught up on the term package in package relative path. Package relative urls are just paths that come in from the user without further context about how to resolve them. In a megamonorepo the user is often in the directory of one package within the larger repo. rootPath is the base of the repo, packagePath is then the cwd.

Or put another way, suppose you're on the donuts project, you do the following:

```
cd /monorepo/food/dessert/circular/donuts
polymer lint components/donut-conveyor/donut-conveyor.html
```

In this example `/monorepo` is the rootPath, `/monorepo/food/dessert/circular/donuts` is the packagePath, `components/donut-conveyor/donut-conveyor.html` is the package-relative path.

src/url-loader/indirect-url-resolver.ts
rictic: Clarified, PTAL

rictic:
All comments addressed. Ignore travis, filed as https://github.com/Polymer/polymer-analyzer/issues/874

rictic:
Thanks!

PR #770, opened 2017-11-28
Scan behavior shorthand methods as methods, not properties.

### LINKED ISSUE #769: Object method shorthand syntax regression in babel branch
The following snippet should produce a Behavior with a *method*.
```
/**
 * @polymerBehavior
 */
MyObj = {
  myMethod() { }
};
```

Instead it produces a Behavior with a *property* called `myMethod` with type `Object`.

Note that the syntax `myMethod: function() {}` still works correctly.

aomarks:
This was a regression from the espree version because babel scans this
syntax differently.

Fixes #769

### DIFF
--- a/src/polymer/behavior-scanner.ts
+++ b/src/polymer/behavior-scanner.ts
@@ -104,7 +104,7 @@ class BehaviorVisitor implements Visitor {
       }
       if (name in this.propertyHandlers) {
         this.propertyHandlers[name](prop.value);
-      } else if (babel.isFunction(prop.value)) {
+      } else if (babel.isMethod(prop) || babel.isFunction(prop.value)) {
         const method = esutil.toScannedMethod(
             prop, this.document.sourceRangeForNode(prop)!, this.document);
         this.currentBehavior.addMethod(method);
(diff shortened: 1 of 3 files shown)


src/polymer/behavior-scanner.ts
```diff
@@ -104,7 +104,7 @@ class BehaviorVisitor implements Visitor {
       }
       if (name in this.propertyHandlers) {
         this.propertyHandlers[name](prop.value);
-      } else if (babel.isFunction(prop.value)) {
+      } else if (babel.isMethod(prop) || babel.isFunction(prop.value)) {
```
rictic: What about when we're scanning legacy polymer element bodies and classes? Are we already checking for shorthand methods there?

aomarks: - `polymer-element-scanner`: Already worked, added a test case in new commit.
- `polymer2-element-scanner`: Already worked, including test case (https://github.com/Polymer/polymer-analyzer/blob/master/src/test/static/polymer2/test-element-10.js).
- `polymer-core-feature-scanner`: Added shorthand method support and test in new commit.

aomarks:
PTAL

PR #773, opened 2017-11-29
Implement Dynamic Imports as a js-import (lazy: true)

usergenic:
* Added `js-import` feature with `lazy: true` for dynamic imports call expressions of the form `import()`.
* Fixes https://github.com/Polymer/polymer-analyzer/issues/740
* [x] Changelog updated


src/test/javascript/javascript-import-scanner_test.ts
```diff
@@ -40,6 +40,24 @@ suite('JavaScriptImportScanner', () => {
...
+        file, '/static/javascript/dynamic-import.js' as ResolvedUrl);
+
+    const visit = (visitor: Visitor) =>
+        Promise.resolve(document.visit([visitor]));
+
+    const {features} = await scanner.scan(document, visit);
+    assert.equal(features.length, 1);
+    assert.equal(features[0].type, 'js-import');
+    assert.equal(features[0].url, '/static/javascript/submodule.js');
+    assert.equal(features[0].lazy, true);
```
rictic: This style of test is legible, but it fails poorly (and I know we do it elsewhere in this file).

Prefer asserting against the interesting subset (or with deepEqual, though for anything with a source range that's not best).

e.g. consider the case where two imports are found. Currently the error message would be `expected 2 to be 1`, but if written something like this:

```typescript
assert.subsetOf(features, [{type: 'js-import', url: '/static/javascript/submodule.js', lazy: true}]);
```

The error would be actionable, you'll see a printout of both imports found, with a nice diff rendering of expected vs actual.

src/test/static/javascript/dynamic-import.js
```diff
@@ -0,0 +1 @@
+import('./submodule.js').then((subModule) => console.log(subModule.subThing));
```
rictic: Recommend a few more imports of various kinds to verify that we're early bailing in the right cases.

e.g. an import call with an identifier, an import call for a name instead of a path, etc

src/test/javascript/javascript-import-scanner_test.ts
usergenic: Awesome suggestion.  Switched both tests to use `assert.constainSubset()` (`subsetOf` not found?)

src/test/static/javascript/dynamic-import.js
usergenic: Updated the dynamic-import.js to be like this:
```js
import('./submodule.js').then((submodule) => {
  console.log(submodule.subThing);
});

import('names/not/supported/yet').then((importByName) => {
  console.log('you know this is not supported yet');
});

const someString = './module.js';
import(someString).then((someModule) => {
  console.log('this is conceivably statically analyzable, but we do not do this yet');
});
```

So two cases not yet supported.

rictic: +1

PR #778, opened 2017-12-03
[Redesigned URLS 3] Remove UrlResolver#canResolve

rictic:

This revealed a number of places where we were not handling unresolvable URLs well. Places where we'd have thrown errors rather than gracefully failing.

### DIFF
--- a/src/core/analysis-context.ts
+++ b/src/core/analysis-context.ts
@@ -487,7 +487,7 @@ export class AnalysisContext {
    * Returns true if the url given is resovable by the Analyzer's `UrlResolver`.
    */
   canResolveUrl(url: PackageRelativeUrl): boolean {
-    return this.resolver.canResolve(url);
+    return this.resolver.resolve(url) !== undefined;
   }
 
   /**
@@ -495,7 +495,10 @@ export class AnalysisContext {
    * URL if it can not be resolved.
    */
   resolveUrl(url: PackageRelativeUrl): ResolvedUrl {
-    return this.resolver.canResolve(url) ? this.resolver.resolve(url) :
-                                           (url as any as ResolvedUrl);
+    const resolved = this.resolver.resolve(url);
+    if (resolved === undefined) {
+      return url as any as ResolvedUrl;
+    }
+    return resolved;
   }
 }
--- a/src/core/analysis-context.ts
+++ b/src/core/analysis-context.ts
@@ -484,13 +484,6 @@ export class AnalysisContext {
   }
 
   /**
-   * Returns true if the url given is resovable by the Analyzer's `UrlResolver`.
-   */
-  canResolveUrl(url: PackageRelativeUrl): boolean {
-    return this.resolver.resolve(url) !== undefined;
-  }
-
-  /**
    * Resolves a URL with this Analyzer's `UrlResolver` or returns the given
    * URL if it can not be resolved.
    */
--- a/src/core/analysis-context.ts
+++ b/src/core/analysis-context.ts
@@ -494,9 +494,7 @@ export class AnalysisContext {
   }
 
   /**
-   * Resolves all resolvable URLs in the list.
-   *
-   * Unresolvable URLs will be filtered out.
+   * Resolves all resolvable URLs in the list, removes unresolvable ones.
    */
   resolveUrls(urls: PackageRelativeUrl[]): ResolvedUrl[] {
     return urls.map((u) => this.resolveUrl(u)!).filter((u) => u !== undefined);
--- a/src/core/analysis-context.ts
+++ b/src/core/analysis-context.ts
@@ -139,8 +139,7 @@ export class AnalysisContext {
    * Returns a copy of this cache context with proper cache invalidation.
    */
   filesChanged(urls: PackageRelativeUrl[]) {
-    const newCache =
-        this._cache.invalidate(urls.map((url) => this.resolveUrl(url)));
+    const newCache = this._cache.invalidate(this.resolveUrls(urls));
     return this._fork(newCache);
   }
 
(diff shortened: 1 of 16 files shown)


src/core/analysis-context.ts
```diff
@@ -483,19 +485,26 @@ export class AnalysisContext {
...
+  resolveUrl(url: PackageRelativeUrl): ResolvedUrl|undefined {
+    const resolved = this.resolver.resolve(url);
+    if (resolved === undefined) {
+      return url as any as ResolvedUrl;
+    }
+    return resolved;
+  }
+
+  resolveUrls(urls: PackageRelativeUrl[]): ResolvedUrl[] {
+    const results = [];
```
justinfagnani: I do like functional style for this sort of this: `return urls.map((u) => this.resolveUrl(u)).filter((u) => u !== undefined)`

src/core/analysis-context.ts
```diff
@@ -483,19 +485,26 @@ export class AnalysisContext {
...
+    for (const url of urls) {
+      const resolved = this.resolveUrl(url);
+      if (resolved !== undefined) {
```
justinfagnani: Why the check for `undefined`? `resolveUrl` returns the input if the resolver can't resolve, so this is just filtering `undefined` as an input URL. Did you mean to filter out URLs that couldn't be resolved instead?

src/core/analyzer.ts
```diff
@@ -88,10 +88,12 @@ export class Analyzer {
       return await previousContext.analyze(this.brandUserInputUrls(urls));
     })();
     const context = await this._analysisComplete;
-    const results = new Map(this.brandUserInputUrls(urls).map(
-        (url) => [url, context.getDocument(context.resolveUrl(url))] as
-                     [string, Document | Warning]));
-    return new Analysis(results);
+    const resolvedUrls = context.resolveUrls(this.brandUserInputUrls(urls));
+    const results = new Map(resolvedUrls.map((url): [
```
justinfagnani: That type annotation is a little confusing to read to me in that position. How about casting the return value?

related: https://github.com/Microsoft/TypeScript/issues/10195

src/core/analysis-context.ts
```diff
@@ -483,19 +485,26 @@ export class AnalysisContext {
...
-                                           (url as any as ResolvedUrl);
+  resolveUrl(url: PackageRelativeUrl): ResolvedUrl|undefined {
+    const resolved = this.resolver.resolve(url);
+    if (resolved === undefined) {
+      return url as any as ResolvedUrl;
+    }
+    return resolved;
+  }
+
+  resolveUrls(urls: PackageRelativeUrl[]): ResolvedUrl[] {
```
justinfagnani: definitely comment that this might return a different length array

src/core/analyzer.ts
```diff
@@ -203,17 +206,10 @@ export class Analyzer {
...
-   */
-  canResolveUrl(url: string): boolean {
-    return this._urlResolver.canResolve(this.brandUserInputUrl(url));
-  }
-
   /**
    * Resoves `url` to a new location.
    */
-  resolveUrl(url: string): ResolvedUrl {
+  resolveUrl(url: string): ResolvedUrl|undefined {
```
justinfagnani: Why not return the original URL if it can't be resolved like in AnalysisContext? I know that would be a change in behavior, but it seems worth it to be consistent. If you want to keep this behavior, maybe change the name of one the methods to indicate the difference in behaviour?

If it's a common pattern to fallback to the input in case the URL can't be resolved, maybe we want an additional method on UrlResolver to do that - it could be implemented in the base class.

src/core/analyzer.ts
rictic: I'd like to avoid the cast here, because a cast will turn a `string` into a `ResolvedUrl`, which we don't want here.

src/core/analyzer.ts
rictic: We don't want to fall back to the original URL, the client should do that if that's the behavior it wants (which most of them don't).

src/core/analysis-context.ts
```diff
@@ -484,18 +486,19 @@ export class AnalysisContext {
...
+  resolveUrl(url: PackageRelativeUrl): ResolvedUrl|undefined {
+    return this.resolver.resolve(url);
   }
 
   /**
-   * Resolves a URL with this Analyzer's `UrlResolver` or returns the given
-   * URL if it can not be resolved.
+   * Resolves all resolvable URLs in the list.
+   *
+   * Unresolvable URLs will be filtered out.
```
FredKSchott: nit: this is important info, consider moving up to top line.

src/core/analysis-context.ts
```diff
@@ -484,18 +486,19 @@ export class AnalysisContext {
...
    */
-  resolveUrl(url: PackageRelativeUrl): ResolvedUrl {
-    return this.resolver.canResolve(url) ? this.resolver.resolve(url) :
-                                           (url as any as ResolvedUrl);
+  resolveUrls(urls: PackageRelativeUrl[]): ResolvedUrl[] {
+    return urls.map((u) => this.resolveUrl(u)!).filter((u) => u !== undefined);
```
FredKSchott: super duper nit: the `!` in `this.resolveUrl(u)!` shouldn't be necessary

src/typescript/typescript-analyzer.ts
```diff
@@ -173,8 +173,13 @@ class AnalyzerCompilerHost implements ts.CompilerHost {
       // since we have a path, we can simply resolve it
       const fileName = path.resolve(path.dirname(containingFile), moduleName) as
           PackageRelativeUrl;
-      const resolvedFileName = this.context.resolveUrl(fileName);
+      const resolvedFileName = this._failSafeResolveUrl(fileName);
       return {resolvedFileName};
     });
   }
+
+  private _failSafeResolveUrl(url: string): ResolvedUrl {
```
FredKSchott: jsdoc this guy especially

src/core/analysis-context.ts
justinfagnani: A cast after the whole expression might be more clear

PR #780, opened 2017-12-05
Construct WarningFilter from ProjectConfig

rictic:
Also add support for filtering by file globs.

This will make it much easier to have unified behavior across the IDE and CLI for configuring which warnings to ignore.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -43,7 +43,8 @@ task('depcheck', function() {
   return new Promise((resolve, reject) => {
            depcheck_lib(
                __dirname,
-               {ignoreDirs: [], ignoreMatches: ['@types/*']},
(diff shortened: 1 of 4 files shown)


usergenic:
Travis says the depcheck determined polymer-project-config is an unused dependency.  Can this be true?

usergenic:
Like, https://github.com/Polymer/polymer-analyzer/pull/780/files#diff-766c3b9df92341d73ad01e583c47f54b totally uses it so that seems like a strange thing for depcheck to be on about.

rictic:
Hm, depcheck is probably reading the JS and not the typescript. We're only using it in type position so it's gone from the JS. I'll add it to depcheck's blacklist.

PR #786, opened 2017-12-09
[Redesigned URLS 1] Add Analyzer.forDirectory; cleanup tests.

rictic:
Two things here:
  * Add a method `Analyzer.forDirectory` which encapsulates the process of getting a well configured analyzer rooted at a given directory. This is nice to land here because it's about to change, as the package url resolver is going to want to know about the root dir too.
  * Modified way that we write most of our tests to depend less on internals, and to the degree that we do depend on internals, use a single test method to do so.

This is a piece pulled out the url resolution PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [3.0.0-pre.3] - 2017-12-08
 
-* Added missing model typings for branded `url` types to top-level package exports.
-* Added `templateTypes` property to functions, extracted from `@template` annotations.
+* Added missing model typings for branded `url` types to top-level package
+  exports.
+* Added `templateTypes` property to functions, extracted from `@template`
+  annotations.
+* Add Analyzer.forDirectory() for easily getting a well configured analyzer
+  for a given directory.
 
 ## [3.0.0-pre.2] - 2017-11-30
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,8 +14,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   exports.
 * Added `templateTypes` property to functions, extracted from `@template`
   annotations.
-* Add Analyzer.forDirectory() for easily getting a well configured analyzer
-  for a given directory.
+* Add `Analyzer.createForDirectory()` for easily getting a well configured
+  analyzer for a given directory.
 
 ## [3.0.0-pre.2] - 2017-11-30
 
--- a/src/core/analyzer.ts
+++ b/src/core/analyzer.ts
@@ -85,7 +85,7 @@ export class Analyzer {
    * at the given directory, but in the future it may take configuration from
    * files including polymer.json or similar.
    */
-  static forDirectory(dirname: string): Analyzer {
+  static createForDirectory(dirname: string): Analyzer {
     return new Analyzer({urlLoader: new FSUrlLoader(dirname)});
   }
 
--- a/src/core/analyzer.ts
+++ b/src/core/analyzer.ts
@@ -20,6 +20,7 @@ import {Analysis, Document, Warning} from '../model/model';
 import {PackageRelativeUrl, ResolvedUrl} from '../model/url';
 import {Parser} from '../parser/parser';
 import {Scanner} from '../scanning/scanner';
+import {FSUrlLoader} from '../url-loader/fs-url-loader';
 import {UrlLoader} from '../url-loader/url-loader';
 import {UrlResolver} from '../url-loader/url-resolver';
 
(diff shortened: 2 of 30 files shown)


rictic:
Strongly recommend reviewing with ?w=1 as well, it reduces the noise greatly.

CHANGELOG.md
```diff
@@ -10,8 +10,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [3.0.0-pre.3] - 2017-12-08
 
-* Added missing model typings for branded `url` types to top-level package exports.
-* Added `templateTypes` property to functions, extracted from `@template` annotations.
+* Added missing model typings for branded `url` types to top-level package
+  exports.
+* Added `templateTypes` property to functions, extracted from `@template`
+  annotations.
+* Add Analyzer.forDirectory() for easily getting a well configured analyzer
```
FredKSchott: nit: `Analyzer.forDirectory()`

src/core/analyzer.ts
```diff
@@ -77,6 +78,17 @@ export class Analyzer {
...
   }
 
+  /**
+   * Returns the best analyzer for the given directory.
+   *
+   * Currently this just creates a simple analyzer with a FS loader rooted
+   * at the given directory, but in the future it may take configuration from
+   * files including polymer.json or similar.
+   */
+  static forDirectory(dirname: string): Analyzer {
```
FredKSchott:
```js
const  analyzer = Analyzer.forDirectory(str);`
```

nit: This method name is a little off gramatically, or at least I've never seen us use this grammar for a static factory method. Just by looking, I'm unsure if I need to do `Analyzer.forDirectory()` vs. `new Analyzer.forDirectory()` (types help, at least, but JS consumers don't get that). I'd prefer to see `Analyzer.createForDirectory()`, which is explicitly a factory method.

It sounds like all 5 PRs might be based on this, so I'm conscious of how hard this might be to change, but this stood out to me.

src/demo/polymer-lint.ts
```diff
@@ -12,17 +12,14 @@
...
 /**
  * A basic demo of a linter CLI using the Analyzer API.
  */
 async function main() {
   const basedir = process.cwd();
-  const analyzer = new Analyzer({
-    urlLoader: new FSUrlLoader(basedir),
-    urlResolver: new PackageUrlResolver()
-  });
+  const analyzer = Analyzer.forDirectory(basedir);
```
FredKSchott: yea, again, this method name feels off :/

CHANGELOG.md
rictic: +1

src/core/analyzer.ts
rictic: +1, this is better.

src/demo/polymer-lint.ts
rictic: Changed!

justinfagnani:
Did you want to pull this method into URLResolver?

rictic:
See: https://github.com/Polymer/polymer-analyzer/pull/793

src/test/test-utils.ts
```diff
@@ -101,3 +103,22 @@ function isReadonlyArray(maybeArr: any): maybeArr is ReadonlyArray<any> {
...
+    analyzer: Analyzer,
+    scanner: Scanner<ParsedDocument, any, any>,
+    url: string): Promise<{features: ScannedFeature[], warnings: Warning[]}> {
+  const context = await analyzer['_analysisComplete'];
+  const resolvedUrl = analyzer.resolveUrl(url);
+  const parsedDocument = await context['_parse'](resolvedUrl);
+  return scan(parsedDocument, [scanner]);
+}
+
+export async function runScannerOnContents(
```
justinfagnani: These names are very similar. Document or disambiguate a bit? I could even see a way to combine them with optional arguments.

src/core/analyzer.ts
```diff
@@ -77,6 +78,17 @@ export class Analyzer {
     }
   }
 
+  /**
+   * Returns the best analyzer for the given directory.
```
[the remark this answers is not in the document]
rictic: Improved wording

src/core/analyzer.ts
```diff
@@ -77,6 +78,17 @@ export class Analyzer {
...
+   *
+   * Currently this just creates a simple analyzer with a FS loader rooted
+   * at the given directory, but in the future it may take configuration from
+   * files including polymer.json or similar.
```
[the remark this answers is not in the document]
rictic: Hm, IMO if you want maximum stability you just shouldn't use this method. This method is definitely closer to the Do What I Mean direction, but not in ways that seem like they'd be problematic. Hm, I guess when we add support for including and excluding files from analysis that will be an issue.

I'm not sure if adding a flag is going to be helpful though, because this is specifically for those cases where we _don't_ have any good way for the user of the final tool to configure how they want the analyzer to be constructed except through their config files.

src/test/test-utils.ts
rictic: +1, documented.

PR #787, opened 2017-12-09
[Redesigned URLS 2] Noop - clarification and cleanup

rictic:
- [x] CHANGELOG.md not updated, this is a noop cleanup change to make subsequent PRs cleaner

### DIFF
--- a/src/core/analyzer.ts
+++ b/src/core/analyzer.ts
@@ -102,16 +102,12 @@ export class Analyzer {
     })();
     const context = await this._analysisComplete;
     const resolvedUrls = uiUrls.map((url) => context.resolveUrl(url));
-    const results = new Map(resolvedUrls.map((url): [
-      ResolvedUrl,
-      Document|Warning
-    ] => [url, context.getDocument(url)]));
-    return new Analysis(results, context);
+    return this._constructAnalysis(context, resolvedUrls);
   }
 
   async analyzePackage(): Promise<Analysis> {
     const previousAnalysisComplete = this._analysisComplete;
-    let _package: Analysis|null = null;
+    let analysis: Analysis;
     this._analysisComplete = (async () => {
       const previousContext = await previousAnalysisComplete;
       if (!previousContext.loader.readDirectory) {
@@ -131,15 +127,20 @@ export class Analyzer {
       const newContext = await previousContext.analyze(filesWithParsers);
       const resolvedFilesWithParsers =
           filesWithParsers.map((url) => newContext.resolveUrl(url));
-      const documentsOrWarnings = new Map(resolvedFilesWithParsers.map((url): [
-        ResolvedUrl,
-        Document|Warning
-      ] => [url, newContext.getDocument(url)]));
-      _package = new Analysis(documentsOrWarnings, newContext);
+
+      analysis = this._constructAnalysis(newContext, resolvedFilesWithParsers);
       return newContext;
     })();
     await this._analysisComplete;
-    return _package!;
+    return analysis!;
+  }
+
+  private _constructAnalysis(
+      context: AnalysisContext, urls: ReadonlyArray<ResolvedUrl>) {
+    const getUrlResultPair:
+        (url: ResolvedUrl) => [ResolvedUrl, Document | Warning] =
+            (url) => [url, context.getDocument(url)];
+    return new Analysis(new Map(urls.map(getUrlResultPair)), context);
   }
 
   /**
--- a/src/core/analyzer.ts
+++ b/src/core/analyzer.ts
@@ -62,17 +62,17 @@ export type LazyEdgeMap = Map<ResolvedUrl, PackageRelativeUrl[]>;
  */
 export class Analyzer {
   private _analysisComplete: Promise<AnalysisContext>;
-  private readonly _urlResolver: UrlResolver;
+  readonly urlResolver: UrlResolver;
   private readonly _urlLoader: UrlLoader;
 
   constructor(options: Options) {
     if (options.__contextPromise) {
       this._urlLoader = options.urlLoader;
-      this._urlResolver = options.urlResolver!;
+      this.urlResolver = options.urlResolver!;
       this._analysisComplete = options.__contextPromise;
     } else {
       const context = new AnalysisContext(options);
-      this._urlResolver = context.resolver;
+      this.urlResolver = context.resolver;
       this._urlLoader = context.loader;
       this._analysisComplete = Promise.resolve(context);
     }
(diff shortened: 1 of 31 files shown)


src/test/test-utils.ts
```diff
@@ -118,7 +118,15 @@ export async function runScannerOnContents(
...
   const overlayLoader = new InMemoryOverlayUrlLoader();
   const analyzer = new Analyzer({urlLoader: overlayLoader});
-  overlayLoader.urlContentsMap.set(analyzer.resolveUrl(url), contents);
+  overlayLoader.urlContentsMap.set(analyzer.resolveUrl(url)!, contents);
   const {features, warnings} = await runScanner(analyzer, scanner, url);
-  return {features, warnings, analyzer};
+  return {features, warnings, analyzer, urlLoader: overlayLoader};
+}
+
+export function fUrl([text]: TemplateStringsArray): FileRelativeUrl {
```
FredKSchott:
```
export function fileRelativeUrl
```

explicit function names > saving characters. I believe this is a team value we've already discussed. As someone who didn't have all our Url types in my head I had no idea what this was doing when I saw it used.

Also, jsdoc

src/test/test-utils.ts
```diff
@@ -118,7 +118,15 @@ export async function runScannerOnContents(
...
+  return text as FileRelativeUrl;
+}
+
+export function rUrl([text]: TemplateStringsArray): ResolvedUrl {
```
FredKSchott: likewise:

```
export function resolvedUrl
```

Also, jsdoc

src/core/analyzer.ts
```diff
@@ -62,17 +62,17 @@ export type LazyEdgeMap = Map<ResolvedUrl, PackageRelativeUrl[]>;
  */
 export class Analyzer {
   private _analysisComplete: Promise<AnalysisContext>;
-  private readonly _urlResolver: UrlResolver;
+  readonly urlResolver: UrlResolver;
```
FredKSchott: assuming this is needed in future PR, but if not please explain why no longer private

src/test/core/analyzer_test.ts
```diff
@@ -176,139 +175,130 @@ suite('Analyzer', () => {
...
-          assert.deepEqual(document.getWarnings({imported: false}), []);
-          assert.deepEqual(
-              document.getWarnings({imported: true}).map((w) => w.toJSON()),
-              [expectedWarning]);
-          assert.deepEqual(
-              chainedDocument.getWarnings({imported: false})
-                  .map((w) => w.toJSON()),
-              [expectedWarning]);
-        });
+    testName = `creates "missing behavior" warnings on imported ` +
```
FredKSchott: I really don't like this `testName` pattern vs. a multi-line `test()` call with string literal argument.

Is that just me? This is mostly personal preference so if I'm alone here I'll survive

src/test/test-utils.ts
```diff
@@ -109,7 +109,7 @@ export async function runScanner(
     scanner: Scanner<ParsedDocument, any, any>,
     url: string): Promise<{features: ScannedFeature[], warnings: Warning[]}> {
   const context = await analyzer['_analysisComplete'];
-  const resolvedUrl = analyzer.resolveUrl(url);
+  const resolvedUrl = analyzer.resolveUrl(url)!;
```
FredKSchott: `analyzer.resolveUrl()` never returns undefined, `!` no longer needed?

src/test/test-utils.ts
```diff
@@ -118,7 +118,15 @@ export async function runScannerOnContents(
     scanner: Scanner<ParsedDocument, any, any>, url: string, contents: string) {
   const overlayLoader = new InMemoryOverlayUrlLoader();
   const analyzer = new Analyzer({urlLoader: overlayLoader});
-  overlayLoader.urlContentsMap.set(analyzer.resolveUrl(url), contents);
+  overlayLoader.urlContentsMap.set(analyzer.resolveUrl(url)!, contents);
```
FredKSchott: ditto

src/test/typescript/typescript-import-scanner_test.ts
```diff
@@ -26,7 +26,7 @@ suite('TypeScriptImportScanner', () => {
     const urlLoader = new InMemoryOverlayUrlLoader();
     const analyzer = new Analyzer(
         {parsers: new Map([['ts', new TypeScriptPreparser()]]), urlLoader});
-    urlLoader.urlContentsMap.set(analyzer.resolveUrl('test.ts'), '');
+    urlLoader.urlContentsMap.set(analyzer.resolveUrl('test.ts')!, '');
```
FredKSchott: ditto

src/test/typescript/typescript-preparser_test.ts
```diff
@@ -53,10 +53,10 @@ suite('TypeScriptParser', () => {
 
     test('throws a WarningCarryingException for parse errors', async () => {
       const contents = 'const const const const const #!@(~~)!();';
-      const url = 'ts-parse-error.ts';
+      const url = 'ts-parse-error.ts' as ResolvedUrl;
```
FredKSchott: nit: `rUrl`? Or is that something we want to implicitly test here

src/test/test-utils.ts
FredKSchott: just saw resolveUrl change to return undefined in #778, nvm

src/test/typescript/typescript-preparser_test.ts
rictic: +1, changed

src/core/analyzer.ts
justinfagnani: Mention in changelog

src/core/analyzer.ts
```diff
@@ -95,15 +95,18 @@ export class Analyzer {
...
     const context = await this._analysisComplete;
-    const results = new Map(this.brandUserInputUrls(urls).map(
-        (url) => [url, context.getDocument(context.resolveUrl(url))] as
-                     [string, Document | Warning]));
-    return new Analysis(results);
+    const resolvedUrls = uiUrls.map((url) => context.resolveUrl(url));
+    const results = new Map(resolvedUrls.map((url): [
+      ResolvedUrl,
+      Document|Warning
+    ] => [url, context.getDocument(url)]));
```
justinfagnani: I had suggested the cast in a recent PR because I find this combination of arrow function, type declaration, return expression and formatting really, really hard to read.

src/core/analyzer.ts
```diff
@@ -126,11 +129,13 @@ export class Analyzer {
...
-
-      const documentsOrWarnings = new Map(filesWithParsers.map(
-          (url) => [url, newContext.getDocument(newContext.resolveUrl(url))] as
-                       [string, Document | Warning]));
-      _package = new Analysis(documentsOrWarnings);
+      const resolvedFilesWithParsers =
+          filesWithParsers.map((url) => newContext.resolveUrl(url));
+      const documentsOrWarnings = new Map(resolvedFilesWithParsers.map((url): [
+        ResolvedUrl,
+        Document|Warning
```
justinfagnani: ditto

One potential solution is to extract the mapper function.

src/test/test-utils.ts
```diff
@@ -118,7 +118,15 @@ export async function runScannerOnContents(
...
   const {features, warnings} = await runScanner(analyzer, scanner, url);
-  return {features, warnings, analyzer};
+  return {features, warnings, analyzer, urlLoader: overlayLoader};
+}
+
+export function fileRelativeUrl([text]: TemplateStringsArray): FileRelativeUrl {
+  return text as FileRelativeUrl;
```
justinfagnani: I'd definitely use a no-op tag function under both of these:

```ts
const noOpTag = (strings: TemplateStringsArray, ...values: any[]) => 
    values.reduce((r, v, i) => r + String(v) + strings[i + 1], strings[0])
```

rictic:
All comments addressed, only the last three commits are new. PTAL

Issue #765, opened 2017-11-25
URL resolvers should produce absolute URLs, including protocol


rictic:
Pros:
  - Would be a very clear and unambiguous model, as we'd have absolute `file://` urls for the default use case
  - I think this would help us better deal with use cases where unresolved and resolved urls are very different (e.g. the internal-to-google case)
  - This will definitely help us better deal with use cases where all resources are remote, like the Polymer Designer case.
  - I've used `file://` urls a lot in the editor service, and they are both powerful and can be made to work consistently cross-platform easily.
  - If we used whatwg urls it would let us give feedback to the WHATWG URL spec.
  - Would let us more easily handle a mixture of local files and remote resources in a project, solving cases like Polymer/polymer-editor-service#95
  - Node's core `fs` module is [working on supporting whatwg urls for filesystem APIs](https://nodejs.org/api/fs.html#fs_whatwg_url_object_support), and starting in 7.0 [whatwg urls are the preferred url objects](https://nodejs.org/api/url.html#url_the_whatwg_url_api). 
  - Would make the division of responsibility between url resolvers and url loaders clearer.

Cons:
  - Resolved URLs would be longer when printed out.
  - Significant migration cost.

rictic:
There's a maintained library for whatwg-urls here: https://www.npmjs.com/package/whatwg-url

rictic:
Open question:
  * Do we use full URL objects, or just require absolute URLs as strings?
    * We use ResolvedUrls as keys in a number of maps, so we would need a canonicalization step if we kept them as objects.
    * URL objects are mutable, which would could really cause problems in places where we're using them as keys in maps
    * If we kept them as strings we'd do a decent amount of re-parsing them over and over again

justinfagnani:
Summary of some chat discussion:

 * `URL`s are hard to use as Map and Set keys, we'd have to make a Map subclass
 * The browser has well-defined semantics for interpreting strings as URLs, and we try to match those, we should keep doing that while supporting external URLs
 * Resolving URLs requires a base URL to resolve from
 * Resolving npm package names requires a base path as well
 * We should update the UrlResolver interface to require a base URL
 * We might want to get rid of `FileRelativeUrl` and `PackageRelativeUrl` branded string types

rictic:
Core changes needed:
  * A ResolvedURL is redefined to mean a complete URL `file:///path/to/foo` or `https://example.com/file.html`. 
[...]

PR #788, opened 2017-12-10
[Redesigned URLS 5] Resolve to absolute urls

rictic:

Fixes https://github.com/Polymer/polymer-analyzer/issues/765

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   `result.successful` is true, `result.value` is a Document. When
   `result.successful` is false, then `result.value` is
   either a Warning or undefined.
-* [BREAKING] UrlResolvers should now return complete urls, like
+* [BREAKING] UrlResolvers must now return complete URLs, like
   file:///path/to/paper-button/paper-button.html or
   https://example.com/components/paper-button/paper-button.html
 * Introduce getDocumentContaining to find containing inline document for a
@@ -20,13 +20,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [3.0.0-pre.5] - 2017-12-15
 * [BREAKING] Removed `Analyzer#resolveUrl` in favor of just exposing the
   `UrlResolver` at `Analyzer#urlResolver.resolve`
+* [BREAKING] Polymer property types are now assumed to be possibly
+  `null|undefined` unless an explicit `@type` annotation says otherwise.
 * Add `Analyzer.createForDirectory()` for easily getting a well configured
   analyzer for a given directory.
 * Add `Import#originalUrl` which has the original url of the import as it was
   encountered in the document, before it was resolved relative to the base url
   of its containing document.
-* [BREAKING] Polymer property types are now assumed to be possibly
-  `null|undefined` unless an explicit `@type` annotation says otherwise.
 * Added `attributeType` field to Polymer property, which contains the name
   of the Polymer property declaration `type` field Constructor.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   `result.successful` is true, `result.value` is a Document. When
   `result.successful` is false, then `result.value` is
   either a Warning or undefined.
-* Introduce getDocumentContaining to find containing inline document for a feature
+* [BREAKING] UrlResolvers should now return complete urls, like
+  file:///path/to/paper-button/paper-button.html or
+  https://example.com/components/paper-button/paper-button.html
+* Introduce getDocumentContaining to find containing inline document for a
+  feature
 <!-- Add new, unreleased changes here. -->
 
 ## [3.0.0-pre.5] - 2017-12-15
--- a/src/core/analyzer.ts
+++ b/src/core/analyzer.ts
@@ -90,7 +90,7 @@ export class Analyzer {
   static createForDirectory(dirname: string): Analyzer {
     return new Analyzer({
       urlLoader: new FSUrlLoader(dirname),
-      urlResolver: new PackageUrlResolver({})
+      urlResolver: new PackageUrlResolver({packageDir: dirname})
     });
   }
 
--- a/src/model/analysis.ts
+++ b/src/model/analysis.ts
@@ -36,6 +36,7 @@ export type Result<T, E> = {
 
 // A regexp that matches paths to external code.
 // TODO(rictic): Make this part of the URL Resolver.
+//     https://github.com/Polymer/polymer-analyzer/issues/803
 // Note that we will match any directory name prefixed by `bower_components` or
 // `node_modules` in order to ignore `polymer install`'s variants, which look
 // like bower_components-foo
--- a/src/model/analysis.ts
+++ b/src/model/analysis.ts
@@ -35,11 +35,11 @@ export type Result<T, E> = {
 };
 
 // A regexp that matches paths to external code.
-// TODO(rictic): Make this extensible (polymer.json?).
-// Note that we match directories named exactly `build`, but will match any
-// directory name prefixed by `bower_components` or `node_modules`, in order to
-// ignore `polymer install`'s variants, which look like bower_components-foo
-const MATCHES_EXTERNAL = /(^|\/)(bower_components|node_modules|build($|\/))/;
+// TODO(rictic): Make this part of the URL Resolver.
+// Note that we will match any directory name prefixed by `bower_components` or
+// `node_modules` in order to ignore `polymer install`'s variants, which look
+// like bower_components-foo
+const MATCHES_EXTERNAL = /(^|\/)(bower_components|node_modules($|\/))/;
 
 /**
  * Represents a queryable interface over all documents in a package/project.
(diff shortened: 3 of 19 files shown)


rictic:
(Tests pass on my machine but so far fail on travis. Investigating)

rictic:
It turns out that travis runs in a directory named `build`, and we filter out `build` directories as being package external.

rictic:
Filed https://github.com/Polymer/polymer-analyzer/issues/790 to improve detecting external files, as that part has been a hack for a while anyways.

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
+* [BREAKING] UrlResolvers should now return complete urls, like
```
justinfagnani: "should" -> "must"

justinfagnani: and capitalize URL

src/model/analysis.ts
```diff
@@ -24,11 +24,11 @@ import {Warning} from './warning';
 
 
 // A regexp that matches paths to external code.
-// TODO(rictic): Make this extensible (polymer.json?).
-// Note that we match directories named exactly `build`, but will match any
-// directory name prefixed by `bower_components` or `node_modules`, in order to
-// ignore `polymer install`'s variants, which look like bower_components-foo
-const MATCHES_EXTERNAL = /(^|\/)(bower_components|node_modules|build($|\/))/;
+// TODO(rictic): Make this part of the URL Resolver.
```
justinfagnani: Let's start making an issue for every TODO

src/core/analyzer.ts
```diff
@@ -90,7 +90,7 @@ export class Analyzer {
   static createForDirectory(dirname: string): Analyzer {
     return new Analyzer({
       urlLoader: new FSUrlLoader(dirname),
-      urlResolver: new PackageUrlResolver({})
+      urlResolver: new PackageUrlResolver({packageDir: dirname})
```
justinfagnani: Since this method creates a PackageUrlResolver, should the method name indicate that? Like `forPackageDirectory` or `forPackage`?

src/model/analysis.ts
```diff
@@ -24,11 +24,11 @@ import {Warning} from './warning';
...
+// Note that we will match any directory name prefixed by `bower_components` or
+// `node_modules` in order to ignore `polymer install`'s variants, which look
+// like bower_components-foo
+const MATCHES_EXTERNAL = /(^|\/)(bower_components|node_modules($|\/))/;
```
justinfagnani: why did we have "build" in there anyway?

src/url-loader/fs-url-loader.ts
```diff
@@ -58,14 +48,11 @@ export class FSUrlLoader implements UrlLoader {
...
-    const pathname =
-        pathlib.normalize(decodeURIComponent(urlObject.pathname || ''));
-    if (!this._isValid(urlObject, pathname)) {
-      throw new Error(`Invalid URL ${url}`);
+  private getFilePath(url: ResolvedUrl): string {
+    if (!this.canLoad(url)) {
+      throw new Error(`FsUrlLoader can not load url: ${url}`);
     }
-    return this.root ? pathlib.join(this.root, pathname) : pathname;
+    return Uri.parse(url).fsPath;
```
justinfagnani: Can we use the whatwg URL object instead?

src/url-loader/fs-url-loader.ts
```diff
@@ -33,19 +31,11 @@ export class FSUrlLoader implements UrlLoader {
...
-    const pathname =
-        pathlib.normalize(decodeURIComponent(urlObject.pathname || ''));
-    return this._isValid(urlObject, pathname);
-  }
-
-  private _isValid(urlObject: Url, pathname: string) {
-    return (urlObject.protocol === 'file' || !urlObject.hostname) &&
-        !pathname.startsWith('../');
+  canLoad(url: ResolvedUrl): boolean {
+    return url.startsWith('file:///');
```
justinfagnani: We can parse the URL and check the scheme and host.

src/url-loader/package-url-resolver.ts
```diff
@@ -30,55 +32,87 @@ export interface PackageUrlResolverOptions {
...
+  readonly componentDir: string;
+  readonly hostname: string|null;
+  private readonly resolvedComponentDir: string;
 
   constructor(options?: PackageUrlResolverOptions) {
     super();
     options = options || {};
+    let packageDir = options.packageDir || process.cwd();
+    if (process.platform === 'win32') {
+      packageDir = packageDir.replace(/\\/g, '/');
```
justinfagnani: Does this handle escapes correctly? I'd rather delegate to a library to do this. What did we do for converting from Windows to URLs elsewhere?

src/url-loader/package-url-resolver.ts
```diff
@@ -30,55 +32,87 @@ export interface PackageUrlResolverOptions {
...
-    if (!this._isValid(urlObject, pathname)) {
-      return undefined;
+  private shouldHandleAsFileUrl(url: Url) {
+    const isLocalFileUrl =
+        url.protocol === 'file:' && (!url.host || url.host === 'localhost');
+    const isOurHostname = url.hostname === this.hostname;
+    return isLocalFileUrl || isOurHostname;
+  }
+
+  private handleFileUrl(url: Url, unresolvedHref: string) {
```
justinfagnani: this could use some more comments

src/url-loader/package-url-resolver.ts
```diff
@@ -91,6 +125,44 @@ export class PackageUrlResolver extends UrlResolver {
       from = this.packageUrl;
       to = fromOrTo;
     }
+    return this.relativeImpl(from, to);
+  }
+
+  private relativeImpl(from: ResolvedUrl, to: ResolvedUrl): FileRelativeUrl {
```
justinfagnani: why is this extracted? is it used elsewhere?

src/url-loader/package-url-resolver.ts
```diff
@@ -91,6 +125,44 @@ export class PackageUrlResolver extends UrlResolver {
       from = this.packageUrl;
       to = fromOrTo;
     }
+    return this.relativeImpl(from, to);
```
justinfagnani: The above unchanged part could benefit from some destructuring, imo:

```ts
const [from, to] = (maybeTo !== undefined) ? [fromOrTo, maybeTo] : [this.packageUrl, fromOrTo];
```

CHANGELOG.md
rictic: done

src/core/analyzer.ts
rictic: I'm cool with `forPackage`, but my goal with this method is for us to use the best url resolver that we can infer for the directory. We don't have a lot of url resolvers at the moment, but I'm roughly thinking that this would be a place where a plugin could contribute to. (In a follow up PR this method is also made async, giving us more leeway).

So I don't think we'll always be using a PackageUrlResolver here.

src/model/analysis.ts
rictic: +1, done.

src/model/analysis.ts
rictic: It's a common output dir name that was causing issues. Given that this wasn't configurable it was supposed to be a quick stopgap

src/url-loader/fs-url-loader.ts
rictic: We can, my thinking was just that this was quick, easy, and correct.

src/url-loader/fs-url-loader.ts
rictic: I'd like to eventually move that direction, but `vscode-uri` has very nice APIs for going from FS Path -> URI and back again, which is also well tested on windows.

I filed https://github.com/Polymer/polymer-analyzer/issues/804 to track this

rictic: Or were you talking about url objects instead of strings?

src/url-loader/package-url-resolver.ts
rictic: It does, windows doesn't use backslashes for escaping in paths.

I think it'd be good to use a library for handling windows URLs. Would you be ok with pushing windows comments to the next PR (https://github.com/Polymer/polymer-analyzer/pull/792) that adds windows testing?

src/url-loader/package-url-resolver.ts
rictic: I wanted to isolate the fromOrTo / maybeTo handling from the implementation, for clarity. With the destructuring one-liner it's much better, but now there's the issue that `to` is written to but `from` is not, so `const` is an error and `let` is a tslint warning.

src/url-loader/package-url-resolver.ts
rictic: +1

src/url-loader/package-url-resolver.ts
rictic: +1, done

rictic:
(merged in the windows fixes, rebased)

src/core/analyzer.ts
rictic: Plan is now to try to move this out into polymer-project-config. I'd like to keep it here for this PR, as we use it in, like, all of our tests right now.

PR #791, opened 2017-12-13
Assume Polymer properties are possibly null|undefined; add attributeType.

aomarks:
Polymer property definitions include a `type` field. This is a reference to the constructor that will be used to deserialize attribute values.
    
In the absence of an explicit `@type` annotation, this deserializer is also used by Analyzer to infer the Closure syntax property type.
    
Previously, we emitted one of: `string|boolean|number|Object|Array|Date`. Note that in Closure syntax, some of these types are nullable, others aren't (all non-primitive types in Closure are nullable by default), and none can be `undefined`.
   
Now, we always emit a possibly `null|undefined` type when inferring the property type.
    
We also now store the original attribute deserializer constructor name in the `attributeType` field.

For context, see this comment on the generated Polymer TypeScript declarations: https://github.com/Polymer/polymer/pull/4928/files/c9be530d6450238a0ea64a06cf5fa1298a8f824c#r155947127, and some of the other properties there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 * Add `Import#originalUrl` which has the original url of the import as it was
   encountered in the document, before it was resolved relative to the base url
   of its containing document.
+* [BREAKING] Polymer property types are now assumed to be possibly
+  `null|undefined` unless an explicit `@type` annotation says otherwise.
+* Added `attributeType` field to Polymer property, which contains the name
+  of the Polymer property declaration `type` field Constructor.
 <!-- Add new, unreleased changes here. -->
 
 ## [3.0.0-pre.4] - 2017-12-14
--- a/src/polymer/analyze-properties.ts
+++ b/src/polymer/analyze-properties.ts
@@ -64,12 +64,13 @@ export function analyzeProperties(
 
     const value = property.value;
     if (babel.isIdentifier(value)) {
-      // If we've already got a type it's from jsdoc and thus canonical.
-      if (!prop.type) {
-        prop.type = value.name;
-      }
+      // Polymer supports this simple syntax, where only the attribute
+      // deserializer is specified.
+      prop.attributeType = value.name;
+
     } else if (!babel.isObjectExpression(value)) {
       continue;
+
     } else {
       /**
        * Parse the expression inside a property object block. e.g.
@@ -91,19 +92,15 @@ export function analyzeProperties(
 
         switch (propertyKey) {
           case 'type':
-            // If we've already got a type, then it was found in the jsdocs,
-            // and is canonical.
-            if (!prop.type) {
-              prop.type = esutil.objectKeyToString(propertyArg.value);
-              if (prop.type === undefined) {
-                prop.warnings.push(new Warning({
-                  code: 'invalid-property-type',
-                  message: 'Invalid type in property object.',
-                  severity: Severity.WARNING,
-                  sourceRange: document.sourceRangeForNode(propertyArg)!,
-                  parsedDocument: document
-                }));
-              }
+            prop.attributeType = esutil.objectKeyToString(propertyArg.value);
+            if (prop.attributeType === undefined && prop.type === undefined) {
+              prop.warnings.push(new Warning({
+                code: 'invalid-property-type',
+                message: 'Invalid type in property object.',
+                severity: Severity.WARNING,
+                sourceRange: document.sourceRangeForNode(propertyArg)!,
+                parsedDocument: document
+              }));
             }
             break;
           case 'notify':
@@ -150,9 +147,20 @@ export function analyzeProperties(
       prop.readOnly = true;
     }
 
-    prop.type = esutil.CLOSURE_CONSTRUCTOR_MAP.get(prop.type!) || prop.type;
+    // If we've already got a type, then it was found in the jsdocs, and is
+    // canonical. Otherwise we can infer it now from the attribute
+    // deserializer.
+    if (prop.type === undefined && prop.attributeType !== undefined) {
+      prop.type = prop.attributeType;
+      // Map primitive constructors to their primitive value types (e.g. String
+      // => string).
+      prop.type = esutil.CLOSURE_CONSTRUCTOR_MAP.get(prop.type) || prop.type;
+      // Assume this type is possibly null or undefined. Users should add a
+      // @type annotation if they want to tighten this type.
+      prop.type += ' | null | undefined';
+    }
 
-    if (!prop.type) {
+    if (prop.type === undefined) {
       prop.warnings.push(new Warning({
         code: 'no-type-for-property',
         message: 'Unable to determine type for property.',
--- a/src/polymer/polymer-element-mixin.ts
+++ b/src/polymer/polymer-element-mixin.ts
@@ -112,16 +112,14 @@ export class PolymerElementMixin extends ElementMixin implements
   }
 
   emitPropertyMetadata(property: PolymerProperty) {
-    const polymerMetadata:
-        {notify?: boolean, observer?: string, readOnly?: boolean} = {};
-    const polymerMetadataFields: Array<keyof typeof polymerMetadata> =
-        ['notify', 'observer', 'readOnly'];
-    for (const field of polymerMetadataFields) {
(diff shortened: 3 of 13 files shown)


src/polymer/analyze-properties.ts
```diff
@@ -152,6 +152,13 @@ export function analyzeProperties(
 
     prop.type = esutil.CLOSURE_CONSTRUCTOR_MAP.get(prop.type!) || prop.type;
 
+    if (prop.type && !typeTag) {
+      if (!nullableByDefault(prop.type)) {
```
justinfagnani: A comment here would be great - say that in Polymer we assume even non-nullable types to be nullable, or something.

src/polymer/analyze-properties.ts
```diff
@@ -152,6 +152,13 @@ export function analyzeProperties(
 
     prop.type = esutil.CLOSURE_CONSTRUCTOR_MAP.get(prop.type!) || prop.type;
 
+    if (prop.type && !typeTag) {
```
justinfagnani: I think lines 153-170 could be factored into a named method just concerned with constructing the right type. The  function name will help document the intention.

src/polymer/analyze-properties.ts
```diff
@@ -167,3 +174,13 @@ export function analyzeProperties(
   }
   return analyzedProps;
 };
+
+function nullableByDefault(closureType: string) {
+  switch (closureType) {
```
justinfagnani: A link to Closure docs on this would be great.

justinfagnani:
I think this is a fine change to make. It's more accurate, and if we think it's less ergonomic, we can work on that in the source we analyze.

I am concerned about the situation I think we have where we use a type annotation on the property _metadata_, but that's separate.

aomarks:
PTAL.

Significant change following discussion. See the new PR description

src/polymer/polymer-element-mixin.ts
```diff
@@ -112,16 +112,14 @@ export class PolymerElementMixin extends ElementMixin implements
...
+    return {
+      polymer: {
+        notify: property.notify,
+        observer: property.observer,
+        readOnly: property.readOnly,
+        attributeType: property.attributeType,
       }
-    }
-    return {polymer: polymerMetadata};
+    };
```
rictic: Ha… that's much simpler

src/test/static/analysis/behaviors/analysis.json
```diff
@@ -39,7 +39,7 @@
             }
           },
           "metadata": {},
-          "type": "number",
+          "type": "?number|undefined",
```
rictic: I'm in favor of everything about this PR except emitting types like this in `analysis.json`. It hurts readability, but I also think this kind of precision is inappropriate given the level of checking we have of types in elements code. This level of precision says that if you don't see `?` or `undefined` then you can expect the value to be defined and not null.

But I don't think that we can generally say that with confidence. But maybe I'm wrong, and our types are more reliable I think?

src/polymer/analyze-properties.ts
```diff
@@ -91,19 +92,15 @@ export function analyzeProperties(
...
-                  code: 'invalid-property-type',
-                  message: 'Invalid type in property object.',
-                  severity: Severity.WARNING,
-                  sourceRange: document.sourceRangeForNode(propertyArg)!,
-                  parsedDocument: document
-                }));
-              }
+            prop.attributeType = esutil.objectKeyToString(propertyArg.value);
+            if (prop.attributeType === undefined) {
+              prop.warnings.push(new Warning({
```
justinfagnani: This seems like it'll introduce new warnings in cases where there was a type annotation, but an "invalid" Polymer type value. I'm not even sure the warning is correct, since things like `type: (v) => ...` are completely valid, yet will cause `undefined` to be returned by `objectKeyToString`.

src/polymer/analyze-properties.ts
```diff
@@ -150,7 +147,21 @@ export function analyzeProperties(
       prop.readOnly = true;
     }
 
-    prop.type = esutil.CLOSURE_CONSTRUCTOR_MAP.get(prop.type!) || prop.type;
+    // If we've already got a type, then it was found in the jsdocs, and is
+    // canonical. Otherwise we can infer it now from the attribute
+    // deserializer.
+    if (!prop.type && prop.attributeType) {
```
justinfagnani: do explicit comparisons and don't rely on ToBoolean conversion

src/polymer/analyze-properties.ts
```diff
@@ -150,7 +147,21 @@ export function analyzeProperties(
...
+      prop.type = prop.attributeType;
+      // Map primitive constructors to their primitive value types (e.g. String
+      // => string).
+      prop.type = esutil.CLOSURE_CONSTRUCTOR_MAP.get(prop.type) || prop.type;
+      // Assume this type is possibly null or undefined. Users should add a
+      // @type annotation if they want to tighten this type.
+      if (!nullableByDefault(prop.type)) {
+        prop.type = '?' + prop.type;
```
justinfagnani: update to `prop.type + '|null'`

src/polymer/polymer-element-mixin.ts
justinfagnani: nice :)

src/polymer/analyze-properties.ts
justinfagnani: ok, I was wrong about arbitrary functions being "completely" valid. If Polymer sees something other than one of its supported types it just falls back to `Object`. Still not sure about the warning, since it's not uncommon to put another type there for documentation purposes, but that's a separate issue.

src/polymer/polymer-element-mixin.ts
aomarks: Yeah I was thinking maybe the original intention was to entirely omit fields whose value are undefined. But `JSON.stringify` does that automatically so it doesn't really matter, given that this is for the JSON format.

src/test/static/analysis/behaviors/analysis.json
aomarks: Discussed offline. Updated to always emit as `<type>|null|undefined`, since that's still a valid Closure type, but also more intuitive to people not familiar with Closure type syntax.

CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 * Analyzer#urlResolver is a property that exposes the analyzer's url resolver,
   for cases where more direct access to url resolution is desired.
 * Fix a situation where a warning would be reported as `[Object object]`.
+* Polymer property types are now assumed to be possibly `null|undefined` unless
```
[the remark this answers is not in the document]
aomarks: Done.

src/polymer/analyze-properties.ts
aomarks: Done.

src/polymer/analyze-properties.ts
aomarks: Done.

src/polymer/analyze-properties.ts
aomarks: Done (restored previous behavior, no warning if you have a `@type` override).

aomarks:
All comments addressed. PTAL!

src/polymer/polymer-element-mixin.ts
rictic: That is the original intention. Provided that these don't include `null` then this will be fine.

src/polymer/analyze-properties.ts
```diff
@@ -150,9 +147,20 @@ export function analyzeProperties(
...
+    // canonical. Otherwise we can infer it now from the attribute
+    // deserializer.
+    if (prop.type === undefined && prop.attributeType !== undefined) {
+      prop.type = prop.attributeType;
+      // Map primitive constructors to their primitive value types (e.g. String
+      // => string).
+      prop.type = esutil.CLOSURE_CONSTRUCTOR_MAP.get(prop.type) || prop.type;
+      // Assume this type is possibly null or undefined. Users should add a
+      // @type annotation if they want to tighten this type.
+      prop.type += '|null|undefined';
```
rictic: What about `' | null | undefined'`

I think the spacing helps legibility.

aomarks: Done.

PR #793, opened 2017-12-13
[Redesigned URLS 7] Add UrlResolver.createFromDirectory. Make createFromDirectory methods async.

rictic:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   `result.successful` is true, `result.value` is a Document. When
   `result.successful` is false, then `result.value` is
   either a Warning or undefined.
+* Add `UrlResolver.createForDirectory()` for easily getting a well configured
+  url resolver for a given directory.
 <!-- Add new, unreleased changes here. -->
 
 ## [3.0.0-pre.5] - 2017-12-15
--- a/src/url-loader/url-resolver.ts
+++ b/src/url-loader/url-resolver.ts
@@ -26,6 +26,12 @@ import {FileRelativeUrl, ResolvedUrl} from '../model/url';
  * to '/bower_components/polymer/polymer.html'.
  */
 export abstract class UrlResolver {
+  static async createForDirectory(_dirname: string): Promise<UrlResolver> {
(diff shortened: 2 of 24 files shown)


rictic:
This can land in parallel to the other work, no longer depends on other PRs

CHANGELOG.md
```diff
@@ -8,8 +8,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 * [BREAKING] Removed `Analyzer#resolveUrl` in favor of just exposing the
   `UrlResolver` at `Analyzer#urlResolver.resolve`
+* [BREAKING] Removed the `UrlResolver#canResolve` method. A UrlResolver should
```
justinfagnani: This was added in #778 do you need a rebase?

src/url-loader/url-resolver.ts
```diff
@@ -26,6 +26,12 @@ import {FileRelativeUrl, ResolvedUrl} from '../model/url';
  * to '/bower_components/polymer/polymer.html'.
  */
 export abstract class UrlResolver {
+  static async createForDirectory(_dirname: string): Promise<UrlResolver> {
```
justinfagnani: As mentioned in another PR, since this creates a resolve with package semantics, I think the name should reflect that.

src/url-loader/url-resolver.ts
```diff
@@ -26,6 +26,12 @@ import {FileRelativeUrl, ResolvedUrl} from '../model/url';
...
+    // Break the import loop by doing a dynamic import.
+    const {PackageUrlResolver} = await import('./package-url-resolver');
+    return new PackageUrlResolver(/* reference dirname here soon */);
```
justinfagnani: I feel a little weird that we're adding new API that defaults to using `bower_components` here. I'd rather either:

 1. Default to _no_ component directory, ie, not package-semantics, if the intention here is to be generic.
 2. Require a component directory so we have fewer places to dig up when we want to default to npm settings.

Also, if the only thing this does is delegate to the `PackageUrlResolver`, and that forces it to be async I question the value of this method. Is it that more ergonomic than `new PackageUrlResolver()`?

FredKSchott:
@justinfagnani PTAL
@rictic please rebase

PR #796, opened 2017-12-14
Speed up `npm run test:watch`

rictic:
We're now mostly limited by the speed at which our tests run.

There's two parts of this change:
  * use `tsc-then` to keep the typescript compiler warm
  * read our fixtures out of src/, never copy them over into lib/

### DIFF
--- a/src/test/test-utils.ts
+++ b/src/test/test-utils.ts
@@ -12,6 +12,7 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
+import * as path from 'path';
 import {Analyzer} from '../core/analyzer';
 import {FileRelativeUrl, ParsedDocument, ResolvedUrl, ScannedFeature, UrlResolver} from '../index';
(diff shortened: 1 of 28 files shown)


justinfagnani:
rebase?

rictic:
Rebased

src/test/test-utils.ts
```diff
@@ -155,3 +156,5 @@ export function resolvedUrl(
     strings: TemplateStringsArray, ...values: any[]): ResolvedUrl {
   return noOpTag(strings, ...values) as ResolvedUrl;
 }
+
+export const fixtureDir = path.join(__dirname, '../../src/test/static');
```
FredKSchott: not blocking, but why not make these a top level `fixtures/` or `static/` dir? Does it contain typescript files that need to be  compiled?

rictic: We totally should. Now that we've got this referenced in one place it'd be a quick change. I'll do it separately

PR #797, opened 2017-12-14
Inline Documents inherit Container Document base href

usergenic:
- Fixes https://github.com/Polymer/polymer-analyzer/issues/795 where inline JavaScript module import statements did not honor their containing document's baseUrl.
- Inline documents now inherit baseUrl from their containing documents.
- Thanks @justinfagnani for showing me where scanned inline documents are born.

### DIFF
--- a/src/test/html/html-script-scanner_test.ts
+++ b/src/test/html/html-script-scanner_test.ts
@@ -19,6 +19,7 @@ import {Analyzer} from '../../core/analyzer';
 import {HtmlScriptScanner} from '../../html/html-script-scanner';
 import {JavaScriptDocument} from '../../javascript/javascript-document';
 import {Analysis} from '../../model/analysis';
+import {Document} from '../../model/document';
 import {ScannedImport, ScannedInlineDocument} from '../../model/model';
 import {runScannerOnContents} from '../test-utils';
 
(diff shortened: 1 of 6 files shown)


src/test/html/html-script-scanner_test.ts
```diff
@@ -63,11 +64,14 @@ suite('HtmlScriptScanner', () => {
...
     before(async () => {
-      analysis = await analyzer.analyze(['js-modules.html']);
+      analysis = await analyzer.analyze(
+          ['js-modules.html', 'base-href/imports-js-module-with-base.html']);
     });
 
     test('finds external module scripts', () => {
-      const htmlScripts = [...analysis.getFeatures({kind: 'html-script'})];
+      const htmlScripts =
+          [...(analysis.getDocument('js-modules.html') as Document)
```
rictic: Why get the document here?

rictic: This comment isn't blocking, btw, but the getDocument call seems unnecessary to me.

usergenic: Because I want to restrict the features to those obtained from the js-modules document instead of across the analysis (which I expanded to include the additional document).

PR #800, opened 2017-12-15
Analysis#getDocument returns a Result.

rictic:

### DIFF
--- a/src/model/analysis.ts
+++ b/src/model/analysis.ts
@@ -25,8 +25,11 @@ import {Warning} from './warning';
 
 export type Result<G, B> = {
   successful: true,
-  value: G
-}|{successful: false, value: B};
+  value: G,
+}|{
+  successful: false,
+  error: B,
+};
 
 // A regexp that matches paths to external code.
 // TODO(rictic): Make this extensible (polymer.json?).
@@ -81,14 +84,14 @@ export class Analysis implements Queryable {
     const url = this.context.resolveUserInputUrl(
         packageRelativeUrl as PackageRelativeUrl);
     if (url === undefined) {
-      return {successful: false, value: undefined};
+      return {successful: false, error: undefined};
     }
     const result = this._results.get(url);
     if (result != null) {
       if (result instanceof Document) {
         return {successful: true, value: result};
       } else {
-        return {successful: false, value: result};
+        return {successful: false, error: result};
       }
     }
     const documents =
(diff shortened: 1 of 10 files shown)


rictic:
This eliminates a lot of `instanceof` checks and `as Document` casts.

src/model/analysis.ts
```diff
@@ -23,6 +23,11 @@ import {ResolvedUrl} from './url';
 import {Warning} from './warning';
 
 
+export type Result<G, B> = {
```
aomarks: Cool type.

rictic: It's inspired by the Rust type of the same name. I've got a lib I was poking at that adds all the methods on the Rust type to a Result class. That was probably overkill, but some of them would be quite useful, like a `getOrDie` method

justinfagnani: If this isn't used anywhere else, it might be nice to make it an `AnalysisResult` with `.document` and `.error`.

src/model/analysis.ts
```diff
@@ -23,6 +23,11 @@ import {ResolvedUrl} from './url';
...
+  successful: true,
+  value: G
+}|{successful: false, value: B};
```
justinfagnani: what about `|{successful: false, error: B}`?

rictic: Is that suboptimal for v8? Or are the different types for `value` enough to give success and failure different shapes?

src/model/analysis.ts
rictic: I'd like to swap this out for a Result class later, to add useful methods.

src/model/analysis.ts
justinfagnani: I wouldn't worry too much about V8 for this API. It's relatively low-frequency, and I've learned that the exact meaning of "call-site" wrt mono/poly/megamorphic isn't what you'd expect. 

The call-site is not when a value is passed to functions, but only when properties are accessed. Since this return type is usually going to be gated by a condition, the accesses to .error vs .value are probably going to remain mono-morphic. The check on .successful might go polymorphic, but again, I think that level of VM optimization is a bit much for this API.

src/model/analysis.ts
justinfagnani: G and B are a little meaningless to me. I guess "good" and "bad"? I'd go for T/E or V/E

rictic: Done

Issue #584, opened 2017-03-27
Notice .bowerrc files and consider the `directory` defined therein as package-external


rictic:
Originally filed by @web-padawan at https://github.com/Polymer/polymer-cli/issues/632

web-padawan:
Maybe worth looking at together with [this line](https://github.com/Polymer/polymer-analyzer/blob/a828aad977287b33989e88bec4a4d31a7d14d028/src/model/package.ts#L27)?

I had to hardcode `vendor` and `dist` which we use in our project instead of `bower_components` and `build` respectively.

justinfagnani:
Seems like we need to un-hardcode bower_components and node_modules everywhere and pass in a list of package directories.

PR #801, opened 2017-12-15
Resolve sibling urls based on config files.

rictic:

  UrlResolver.createForDirectory
      ✓ resolves sibling urls to bower_components by default
      ✓ resolves sibling urls to the directory specified in .bowerrc if present
      ✓ resolves sibling urls to bower_components if there is a bower.json and a package.json
      ✓ resolves sibling urls to node_modules if there is package.json but no bower.json


rictic:
Fixes these issues:
Fixes https://github.com/Polymer/polymer-cli/issues/785
Fixes https://github.com/Polymer/polymer-editor-service/issues/104

For these, it's a step in the right direction, but doesn't fully fix:
Addresses https://github.com/Polymer/polymer-cli/issues/765
Addresses https://github.com/Polymer/polymer-analyzer/issues/584
Addresses https://github.com/Polymer/polymer-cli/issues/487
Addresses https://github.com/Polymer/polyserve/issues/101

src/url-loader/url-resolver.ts
```diff
@@ -65,3 +70,25 @@ export abstract class UrlResolver {
     return url as ResolvedUrl;
   }
 }
+
+async function inferComponentDirname(dirname: string): Promise<string> {
+  try {
+    const contents = await fs.readFile(
+        pathlib.join(dirname, '.bowerrc'), {encoding: 'utf8'});
```
justinfagnani: If I recall, Bower has some particular method of finding a .bowerrc file, and there's a library you can use to find it. It might just be a findup-style lookup.

src/url-loader/url-resolver.ts
```diff
@@ -65,3 +70,25 @@ export abstract class UrlResolver {
...
+    const bowerrc = JSON.parse(contents);
```
justinfagnani: I think said library (bower-something), parses it against default options too...

src/url-loader/url-resolver.ts
```diff
@@ -65,3 +70,25 @@ export abstract class UrlResolver {
...
+    if (typeof bowerrc.directory === 'string') {
+      return bowerrc.directory;
+    }
+  } catch { /* don't care */
+  }
+  const [bowerJsonExists, packageJsonExists] = await Promise.all([
+    fs.exists(pathlib.join(dirname, 'bower.json')),
+    fs.exists(pathlib.join(dirname, 'package.json')),
```
justinfagnani: similarly, I think there's a library to find and parse package.json's

src/url-loader/url-resolver.ts
```diff
@@ -65,3 +70,25 @@ export abstract class UrlResolver {
...
+  ]);
+  if (bowerJsonExists) {
```
justinfagnani: I think we need to wire in handling for the case where we force npm or Bower. I'm about to add the --npm flag to polymer-build and friends, and in a future release we'll switch the default to npm. It'd be great if code at this level didn't have to change because the preference for the default and the force option were already passed down.

rictic:
From chat: maybe this logic should live on polymer project config

justinfagnani:
@rictic what's the status here?

FredKSchott:
Closing this PR as abandoned.

PR #802, opened 2017-12-15
Give ranges for css custom property declarations only on the name.

rictic:
This aligns us with the vscode css languageserver, so that users don't see duplicate definitions when they find references/jump to definition. Related: https://github.com/Microsoft/vscode/issues/39890

 - [x] CHANGELOG.md not updated, mostly internal change

### DIFF
--- a/src/css/css-custom-property-scanner.ts
+++ b/src/css/css-custom-property-scanner.ts
@@ -90,7 +90,7 @@ class CssCustomPropertyVisitor implements Visitor {
     if (node.type === shady.nodeType.declaration &&
         node.name.startsWith('--')) {
       this.features.push(new CssCustomPropertyAssignment(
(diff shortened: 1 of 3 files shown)


src/css/css-custom-property-scanner.ts
```diff
@@ -90,7 +90,7 @@ class CssCustomPropertyVisitor implements Visitor {
     if (node.type === shady.nodeType.declaration &&
         node.name.startsWith('--')) {
       this.features.push(new CssCustomPropertyAssignment(
-          node.name, this.document.sourceRangeForNode(node)));
+          node.name, this.document.sourceRangeForShadyRange(node.nameRange)));
```
aomarks: Could you add a description to `sourceRangeForShadyRange` to briefly explain why it's different to `sourceRangeForNode`?

PR #811, opened 2017-12-20
Expose FSUrlLoader#getFilePath

rictic:

This method will be useful in the CLI, where we need to get these paths in order to write out files when fixing lint warnings.

This also fixes a bug whereby we weren't verifying that the loaded file contained within the root dir.

### DIFF
--- a/src/core/utils.ts
+++ b/src/core/utils.ts
@@ -1,7 +1,20 @@
+/**
+ * @license
+ * Copyright (c) 2015 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at
+ * http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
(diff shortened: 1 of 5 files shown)


src/core/utils.ts
```diff
@@ -1,16 +1,3 @@
-/**
```
aomarks: why remove license?

src/core/utils.ts
```diff
@@ -33,6 +20,13 @@ export function trimLeft(str: string, char: string): string {
   return str.substring(leftEdge);
 }
 
+export function isPathInside(directory: string, filePath: string): boolean {
```
aomarks: comment? not 100% intuitive what this does

src/core/utils.ts
rictic: Arg, typescript bug where it removes the license comment if you apply the edit action to remove an unused import.

src/core/utils.ts
rictic: Done

PR #812, opened 2017-12-21
Updated the url resolver code to consider and preserve urls with search and hash values.

usergenic:
- `UrlResolver#resolve()` returns urls containing search and hash components where they were previously stripped out.
- Note this change is a pre-requisite for being able to drop Bundler's custom relative url calculations in favor of those exposed by analyzer's resolvers.
- [x] CHANGELOG updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 * Add FSUrlLoader#getFilePath which will return the file path that would
   be loaded for a given ResolvedUrl, or an error message explaining why
   it can't be.
+* Add a `resolver` option when converting warnings to string, or when
+  using the WarningPrinter. This will resolve file urls in warnings using
+  `resolver.relative`, to avoid super long urls in warning messages.
 <!-- Add new, unreleased changes here. -->
 
 ## [3.0.0-pre.6] - 2017-12-18
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+* `UrlResolver#resolve()` returns urls containing search and hash
+  components where they were previously stripped out.
+<!-- Add new, unreleased changes here. -->
+
 ## [3.0.0-pre.6] - 2017-12-18
 * [BREAKING] `Analysis#getDocument` now returns a `Result` object. When
   `result.successful` is true, `result.value` is a Document. When
@@ -15,7 +20,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   https://example.com/components/paper-button/paper-button.html
 * Introduce getDocumentContaining to find containing inline document for a
   feature
-<!-- Add new, unreleased changes here. -->
 
 ## [3.0.0-pre.5] - 2017-12-15
 * [BREAKING] Removed `Analyzer#resolveUrl` in favor of just exposing the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
-* `UrlResolver#resolve()` returns urls containing search and hash
+* `UrlResolver#resolve()` returns urls containing querystring and fragment
   components where they were previously stripped out.
 * Add FSUrlLoader#getFilePath which will return the file path that would
   be loaded for a given ResolvedUrl, or an error message explaining why
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 * `UrlResolver#resolve()` returns urls containing search and hash
   components where they were previously stripped out.
+* Add FSUrlLoader#getFilePath which will return the file path that would
+  be loaded for a given ResolvedUrl, or an error message explaining why
+  it can't be.
 <!-- Add new, unreleased changes here. -->
 
 ## [3.0.0-pre.6] - 2017-12-18
--- a/src/test/url-loader/package-url-resolver_test.ts
+++ b/src/test/url-loader/package-url-resolver_test.ts
@@ -152,6 +152,27 @@ suite('PackageUrlResolver', function() {
       assert.equal(
           resolver.resolve(fileRelativeUrl`%><><%=`, packageRoot), undefined);
     });
+
+    test('resolves a URL with no pathname', () => {
+      const foo = rootedFileUrl`1/2/foo.html?baz#bat`;
+      const bar = rootedFileUrl`1/2/bar.html`;
+      assert.equal(
+          resolver.resolve(fileRelativeUrl``, foo),
+          rootedFileUrl`1/2/foo.html?baz`);
+      assert.equal(resolver.resolve(fileRelativeUrl``, bar), bar);
+      assert.equal(
+          resolver.resolve(fileRelativeUrl`#buz`, foo),
+          rootedFileUrl`1/2/foo.html?baz#buz`);
+      assert.equal(
+          resolver.resolve(fileRelativeUrl`#buz`, bar),
+          rootedFileUrl`1/2/bar.html#buz`);
+      assert.equal(
+          resolver.resolve(fileRelativeUrl`?fiz#buz`, foo),
+          rootedFileUrl`1/2/foo.html?fiz#buz`);
+      assert.equal(
+          resolver.resolve(fileRelativeUrl`?fiz#buz`, bar),
+          rootedFileUrl`1/2/bar.html?fiz#buz`);
+    });
   });
 
   suite('relative', () => {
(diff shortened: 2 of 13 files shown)


src/url-loader/url-resolver.ts
```diff
@@ -46,8 +49,19 @@ export abstract class UrlResolver {
...
       baseUrl: ResolvedUrl): ResolvedUrl {
     let resolved = urlLibResolver(baseUrl, url);
-    if (url.endsWith('/') && !resolved.endsWith('/')) {
-      resolved += '/';
+
+    // TODO(usergenic): There is no *explicit* test coverage for the missing
+    // trailing slash on urlLibResolver'd urls.  Investigate why this
+    // double-check is necessary here, since I adapted the existing logic
+    // in-place from the prior form to support urls containing search and hash.
+    // I'm guessing this is related to Windows...?
```
rictic: Hm, maybe it's not necessary after all, it looks like urlLibResolver does preserve trailing slashes. Feel free to drop this logic then.

The tests document the expectations of this code, otherwise we'll end up scared to make any changes at all to it. No haunted graveyards! http://noidea.dog/srecon17americas2#no-haunted-graveyards-john-truscott-reese-computer-scientist-google

src/url-loader/url-resolver.ts
```diff
@@ -12,13 +12,16 @@
...
-import {posix} from 'path';
-import {resolve as urlLibResolver} from 'url';
+import * as path from 'path';
+import {format as urlLibFormat, resolve as urlLibResolver} from 'url';
 
 import {parseUrl} from '../core/utils';
 import {PackageRelativeUrl, ScannedImport} from '../index';
 import {FileRelativeUrl, ResolvedUrl} from '../model/url';
 
+const sharedRelativeUrlProperties =
```
rictic: I think it's safe to drop `port` and `hostname` here, as `host` is a combination of the two.

rictic: Also document this please :)

src/url-loader/url-resolver.ts
```diff
@@ -12,13 +12,16 @@
...
+    ['protocol', 'slashes', 'auth', 'host', 'port', 'hostname'];
```
rictic: Can avoid the `as any` business below by writing this as:

```typescript
['protocol' as 'protocol', 'slashes' as 'slashes', 'auth' as 'auth', 
 'host' as 'host', 'port' as 'port', 'hostname' as 'hostname']
```

usergenic: I am both excited to learn this technique and yet bothered by it all the same; but I totally get it.  Thanks!

src/url-loader/url-resolver.ts
usergenic: Okay dropped port and hostname in favor of just host.  This brings the total number of properties I consider to four.  As such I decided against an array of them and just explicitly reference where used.  It makes the code clearer and I also have better type checking for the property values whereas previously only checked for null instead of null|undefined-- now I can just check non-emptiness by testing whether values are of the specific types for the properties.

src/url-loader/url-resolver.ts
usergenic: +1

usergenic: Removed

src/url-loader/url-resolver.ts
usergenic: I went ahead and dropped the array anyways since I the logic for checking each of the four properties was at least as easy to follow as iterating.

CHANGELOG.md
```diff
@@ -6,9 +6,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
+* `UrlResolver#resolve()` returns urls containing search and hash
```
justinfagnani: "search" -> "query string", "hash" -> "fragment"

src/test/url-loader/package-url-resolver_test.ts
```diff
@@ -152,6 +152,27 @@ suite('PackageUrlResolver', function() {
       assert.equal(
           resolver.resolve(fileRelativeUrl`%><><%=`, packageRoot), undefined);
     });
+
+    test('resolves a URL with no pathname', () => {
```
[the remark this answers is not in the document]
justinfagnani: Also a test where the base path contains fragments and query parameters to verify that those are not present on the result.

src/test/url-loader/package-url-resolver_test.ts
```diff
@@ -152,6 +152,27 @@ suite('PackageUrlResolver', function() {
...
+      const foo = rootedFileUrl`1/2/foo.html?baz#bat`;
+      const bar = rootedFileUrl`1/2/bar.html`;
```
justinfagnani: Could you split this test up? There's a combinatorial set of conditions of {path, fragment, query} * {base, target}

src/test/url-loader/package-url-resolver_test.ts
```diff
@@ -152,6 +152,27 @@ suite('PackageUrlResolver', function() {
...
+      assert.equal(
+          resolver.resolve(fileRelativeUrl``, foo),
+          rootedFileUrl`1/2/foo.html?baz`);
```
justinfagnani: why is the fragment missing here?

src/test/url-loader/package-url-resolver_test.ts
```diff
@@ -152,6 +152,27 @@ suite('PackageUrlResolver', function() {
...
+      assert.equal(resolver.resolve(fileRelativeUrl``, bar), bar);
+      assert.equal(
+          resolver.resolve(fileRelativeUrl`#buz`, foo),
+          rootedFileUrl`1/2/foo.html?baz#buz`);
```
justinfagnani: I'm pretty sure `#buz` should not be in the result

src/test/url-loader/package-url-resolver_test.ts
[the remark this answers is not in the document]
usergenic: Added

src/test/url-loader/package-url-resolver_test.ts
usergenic: Fragments can never be implicit; there's no URL form you can construct relative to a URL with a fragment that will preserve the fragment.  So it is clearest to test that assumption here with an empty relative URL.

src/test/url-loader/package-url-resolver_test.ts
usergenic: `#buz` is in the result *because* it was in the url to resolve and because that's part of the main purpose of this PR :)

CHANGELOG.md
usergenic: Updated in CHANGELOG and in the mocha test names.

usergenic: I had just used the property from node's Url object.  I'll use querystring and fragment happily.

src/test/url-loader/package-url-resolver_test.ts
usergenic: I removed the vars and ordered the items progressively.  Maybe is okay now?

src/test/url-loader/package-url-resolver_test.ts
```diff
@@ -152,8 +152,46 @@ suite('PackageUrlResolver', function() {
...
+              fileRelativeUrl`#buz`, rootedFileUrl`1/2/foo.html?baz#bat`),
+          rootedFileUrl`1/2/foo.html?baz#buz`);
+      assert.equal(
+          resolver.resolve(
+              fileRelativeUrl`?fiz#buz`, rootedFileUrl`1/2/foo.html`),
+          rootedFileUrl`1/2/foo.html?fiz#buz`);
+      assert.equal(
+          resolver.resolve(
+              fileRelativeUrl`?fiz#buz`, rootedFileUrl`1/2/foo.html?bar`),
+          rootedFileUrl`1/2/foo.html?fiz#buz`);
```
rictic: Could use a test that just a query string overrides the fragment in the base

```typescript
      assert.equal(
          resolver.resolve(
              fileRelativeUrl`?fiz`, rootedFileUrl`1/2/foo.html?bar#buz`),
          rootedFileUrl`1/2/foo.html?fiz`);
```

src/test/url-loader/url-resolver_test.ts
```diff
@@ -39,6 +39,20 @@ class SimplestUrlResolver extends UrlResolver {
...
+    const resolver = new SimplestUrlResolver();
+    function resolve(url: string, baseUrl: string) {
+      return resolver.resolve(url as FileRelativeUrl, baseUrl as ResolvedUrl);
+    }
+    test('can resolve a url when relative url contains no pathname', () => {
+      assert.equal(resolve('', '/foo.html?fiz#buz'), '/foo.html?fiz');
+      assert.equal(resolve('#fiz', '/foo.html'), '/foo.html#fiz');
+      assert.equal(resolve('#fiz', '/foo.html#buz'), '/foo.html#fiz');
+      assert.equal(resolve('?fiz', '/foo.html'), '/foo.html?fiz');
+      assert.equal(resolve('?fiz', '/foo.html?buz'), '/foo.html?fiz');
```
rictic: Could use a test that just a query string overrides the fragment in the base

```typescript
      assert.equal(
          resolve(
              `?fiz`, `/foo.html?bar#buz`), `/foo.html?fiz`);
```

src/test/url-loader/package-url-resolver_test.ts
usergenic: Added.

src/test/url-loader/url-resolver_test.ts
usergenic: Added.

src/url-loader/url-resolver.ts
```diff
@@ -45,33 +45,44 @@ export abstract class UrlResolver {
...
+      toUrl.pathname =
+          path.posix.relative(fromDir, toDir + '_').replace(/_$/, '');
     }
-    return this.brandAsRelative(relativized);
+    delete toUrl.protocol;
+    delete toUrl.slashes;
+    delete toUrl.auth;
+    delete toUrl.host;
+    delete toUrl.path;
+    return this.brandAsRelative(urlLibFormat(toUrl));
```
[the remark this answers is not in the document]
usergenic: Yeah its clearer if I leave the toUrl alone and just construct a new URL.  Pushed changes.

PR #813, opened 2017-12-22
Add methods to check a parsed document's type without instanceof.

rictic:


src/css/css-document.ts
```diff
@@ -22,12 +22,16 @@ import cssbeautify = require('cssbeautify');
...
 
 export class ParsedCssDocument extends ParsedDocument<shady.Node, Visitor> {
-  type = 'css';
+  readonly type = 'css';
 
   constructor(from: Options<shady.Node>) {
     super(from);
   }
 
+  static is(document: ParsedDocument): document is ParsedCssDocument {
```
justinfagnani: This ends up reading a bit backwards: `ParsedCssDocument.is(document)`

I got tantalizingly close to making `document.is(ParsedCssDocument)` work in a well-typed way, but failed because TypeScript doesn't infer nested type parameters:

```ts
class X {
  type: string;
  is<T>(c: XConstructor<T>): this is T {
    return this.type === c.type;
  }
}

interface XConstructor<T> {
  readonly type: string;
  new(): T;
}

class Y extends X {
  static type = 'Y';
  foo: string;
} 

function go(x: X) {
  if (x.is<Y>(Y)) {
    x.foo;
  }
}
```

The type argument here has to be explicit: `x.is<Y>(Y)` otherwise it's inferred as `X` and there's an error at `x.foo`.

justinfagnani: D'oh! I was trying that on the playground. It works in 2.6 :D

rictic: So that code above only works if we double-declare types, like
```typescript
class Y extends X {
  static readonly type = 'Y';
  readonly type = 'Y';
}
```

We can avoid this in two ways. This one is awkward because typescript doesn't like us assigning to a readonly value, even on the prototype:
```typescript
abstract class X {
  abstract readonly type: string;
  static get type(): string {
    return this.prototype.type;
  }
  is<T>(c: XConstructor<T>): this is T {
    return this.type === c.type;
  }
}

interface XConstructor<T> {
  readonly type: string;
  new(): T;
}

class Y extends X {
  readonly type: 'Y';
  foo: string;
}
Y.prototype.type = 'Y';
```

And this one is awkward because there is no warning if we forget to declare a `static readonly type` property on a subclass:
```typescript
abstract class X {
  get type() {
    return (this.constructor as XConstructor<this>).type;
  }
  static readonly type: string;
  is<T>(c: XConstructor<T>): this is T {
    return this.type === c.type;
  }
}

interface XConstructor<T> {
  readonly type: string;
  new(): T;
}

class Y extends X {
  static readonly type = 'Y';
  foo: string;
}
```

rictic: At that point though, why not just:

```typescript
abstract class X {
  is<T>(c: XConstructor<T>): this is T {
    return this.constructor.name === c.name;
  }
}

interface XConstructor<T> {
  new(): T;
}

class Y extends X {
  foo: string;
}
```

rictic: Or even:
```typescript
abstract class X {
  static[Symbol.hasInstance](v: any) {
    if (v == null) {
      return false;
    }
    return this.name === v.constructor.name;
  }
}

class Y extends X {
  foo: string;
}
```

I guess the last one will slow down all instanceof calls in node v6, so maybe we should hold off on it.

rictic:
I went with the constructor name option

aomarks:
Closing this pull request because this repo is now archived. If this PR is still active, please submit a new one at https://github.com/Polymer/tools.

PR #816, opened 2017-12-27
detect getters/setters as properties

### LINKED ISSUE #645: ES6 Getters aren't recognized as properties
The analyzer seems to completely ignore ES6 getters - they are not recognized as properties.

We're beginning to switch our code base to use ES6 getters instead of computed properties.

```js
class FooElement extends Polymer.Element {
  static get is() { return 'foo-element'; }

  get otherProp() {
    return 'some value';
  }
}
```

43081j:
This is an attempt at adding the ability to detect getters/setters in the class scanner.

There's a bit of duplication as you can see, so feel free to suggest some refactoring.

Also, what should we do with statics? they are ignored right now.

cc @rictic 

Fixes #645

### DIFF
--- a/src/javascript/class-scanner.ts
+++ b/src/javascript/class-scanner.ts
@@ -16,7 +16,7 @@ import generate from 'babel-generator';
 import * as babel from 'babel-types';
 import * as doctrine from 'doctrine';
 
-import {ScannedClass, ScannedFeature, ScannedMethod, ScannedProperty, ScannedReference, Severity, SourceRange, Warning} from '../model/model';
+import {Privacy, ScannedClass, ScannedFeature, ScannedMethod, ScannedProperty, ScannedReference, Severity, SourceRange, Warning} from '../model/model';
 import {extractObservers} from '../polymer/declaration-property-handlers';
 import {mergePropertyDeclarations, Observer, ScannedPolymerElement} from '../polymer/polymer-element';
 import {ScannedPolymerElementMixin} from '../polymer/polymer-element-mixin';
@@ -411,8 +411,7 @@ class ClassFinder implements Visitor {
     const namespacedName = name && getNamespacedIdentifier(name, doc);
 
     const warnings: Warning[] = [];
-    const properties =
-        extractPropertiesFromConstructor(astNode, this._document);
+    const properties = extractPropertiesFromClass(astNode, this._document);
 
     this.classes.push(new ScannedClass(
         namespacedName,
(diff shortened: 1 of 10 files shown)


src/test/static/class/class-properties.js
```diff
@@ -0,0 +1,25 @@
+class Class {
+  /** @public a static property with a getter */
+  static get customStaticPropertyGetter() {
+    return true;
+  }
+
+  /** @public a property with a getter */
+  get customPropertyGetter() { return true; }
+
+  /** @oublic a property with a getter/setter */
```
aomarks: `@oublic`?

src/test/static/class/class-properties.js
```diff
@@ -0,0 +1,25 @@
+class Class {
+  /** @public a static property with a getter */
```
aomarks: does this require the `@public` annotation to be picked up?

src/test/javascript/class-scanner_test.ts
```diff
@@ -154,6 +154,23 @@ suite('Class', () => {
       ]);
     });
 
+    test('finds properties', async () => {
+      const classes = await getScannedClasses('class/class-properties.js');
+      assert.deepEqual(await Promise.all(classes.map((c) => getTestProps(c))), [
```
aomarks: Can you add checks that it picks up a `@return` annotation and `description`?

src/javascript/class-scanner.ts
```diff
@@ -594,70 +668,72 @@ export function extractPropertiesFromConstructor(
...
-        // The comment only counts if there's a jsdoc annotation in there
-        // somewhere.
-        // Otherwise it's just an assignment, maybe to a property in a
-        // super class or something.
-        continue;
+
+      let existing = properties.get(prop.name);
+
+      if (!existing) {
+        prop.readOnly = (method.kind === 'get');
```
aomarks: Seems like we should have an explicit way to indicate that something is a getter or setter.

src/test/static/class/class-properties.js
43081j: it doesn't, i think it will just be used for setting `privacy` on the scanned property :)

src/test/static/class/class-properties.js
43081j: :man_facepalming:

src/javascript/class-scanner.ts
```diff
@@ -586,6 +586,80 @@ class CustomElementsDefineCallFinder implements Visitor {
   }
 }
 
+export function extractPropertyFromMethod(
```
rictic: These functions aren't used outside this file, so they should not be exported.

Every declaration should be as private and as read only as possible, IMO.

src/javascript/class-scanner.ts
rictic: IMO from the outside IMO the only thing that matters is whether a property is readonly or not. It's an implementation detail whether a field is a property or a getter/setter pair.

src/javascript/class-scanner.ts
```diff
@@ -594,70 +668,72 @@ export function extractPropertiesFromConstructor(
...
-      } else {
-        continue;
+
+    if (method.kind === 'constructor') {
+      const constructor = method;
+      for (const statement of constructor.body.body) {
+        if (!babel.isExpressionStatement(statement)) {
+          continue;
+        }
+        const prop = extractPropertyFromExpressionStatement(statement, document);
```
rictic: This organization is a bit confusing for the reader, and it's a bit weird to extract a property from a generalized statement. The function only really makes sense for a constructor body.

What about renaming extractPropertiesFromConstructor to `extractPropertiesFromClass`, then putting all of the constructor handling code in an `extractPropertiesFromConstructor` function.

src/javascript/class-scanner.ts
```diff
@@ -594,70 +668,72 @@ export function extractPropertiesFromConstructor(
...
+        if (!prop) {
+          continue;
+        }
+        properties.set(prop.name, prop);
       }
-      if (name === undefined) {
+    } else if (method.kind === 'get' || method.kind === 'set') {
+      const prop = extractPropertyFromMethod(method, document);
```
rictic: Likewise, this isn't extracting a property from a method, it's extracting a property from a getter or setter. Otherwise we should emit a `Method` not a `Property`

src/test/static/class/class-properties.js
```diff
@@ -0,0 +1,25 @@
...
+
+  /** @public a property with a getter */
+  get customPropertyGetter() { return true; }
+
+  /** @public a property with a getter/setter */
+  get customPropertyWithGetterSetter() {
+    return this._customPropertyWithGetterSetter;
+  }
+
+  set customPropertyWithGetterSetter(val) {
```
rictic: We should also test the case where the setter comes before the getter. I think your code handles this correctly now, but it would be an easy thing for us to accidentally break later.

src/javascript/class-scanner.ts
```diff
@@ -586,6 +586,80 @@ class CustomElementsDefineCallFinder implements Visitor {
...
+    return null;
+  }
+
+  if (method.key.name === undefined) {
+    return null;
+  }
+
+  const annotation = getJSDocAnnotationForNode(method);
+  if (!annotation) {
+    return null;
```
rictic: Do we want to emit nothing if there's no annotation? It's still definitely a property.

43081j: this is what the previous behaviour was so im not sure. i think it was to avoid reading `/** */` as a property or some such thing... happy to change it if we no longer want that.

src/javascript/class-scanner.ts
43081j: yup im currently refactoring this a bit as its nasty right now :zipper_mouth_face:

src/javascript/class-scanner.ts
justinfagnani: At least I hope the differences that JavaScript has aren't ever important

aomarks: Fair enough. I was thinking we'd want to know for the generated TypeScript declarations, but in fact we don't (actually it's not even valid to declare a getter in an ambient context, only in an implementation).

43081j: this is majorly wrong at the min.

reason being, jsdoc states you should only annotate one of the two (the get or the set). so there should never be an `existing`. what we actually need to do is assume the one with an annotation is the correct one, then look for the other at that point.

src/test/static/class/class-properties.js
43081j: hmm it seems they are only picked up if theres at least one jsdoc comment (`@something`). so `/** foo */` doesn't yet work (@rictic commented about this above somewhere too, saying maybe we should still pick them up).

src/test/static/class/class-properties.js
43081j: yup i've added that now

src/javascript/class-scanner.ts
```diff
@@ -586,78 +586,194 @@ class CustomElementsDefineCallFinder implements Visitor {
...
+        accessor.setter = method;
       }
-      const description = getDescription(jsdocAnn);
-      let type = undefined;
-      const typeTag = jsdoc.getTag(jsdocAnn, 'type');
-      if (typeTag && typeTag.type) {
-        type = doctrine.type.stringify(typeTag.type);
+    }
+
+    for (const val of accessors.values()) {
```
43081j: @rictic here's the accessors iterator, what do you think?

43081j:
pushed a few changes:

* accessors are always picked up even if there is no annotation
* accessors look for `readonly` rather than `const`
* updated chai for some new assert methods
* use `@return` for the type of a getter rather than `@type`

i would like to run `format` too. it seems it hasn't been run in a while so there's plenty of changes. but its having some problems with `test-element-12.js` where the indentation gets fudged.

rictic:
Hm, `format` isn't supposed to touch the test fixtures. Let's update clang-format and run it in a separate PR.

rictic:
(As an aside on process, it's better to add additional commits rather than rebasing / squashing down during review, as it's hard for reviewers to keep track of what's new about a commit since the last review. This is mostly github's fault, as their review UI is really bare bones)

src/test/static/class/class-properties.js
```diff
@@ -0,0 +1,49 @@
...
+  set customPropertyWithSetterFirst(val) {
+    this._customPropertyWithSetterFirst = val;
+  }
+
+  get customPropertyWithSetterFirst() {
+    return this._customPropertyWithSetterFirst;
+  }
+
+  /** @readonly */
+  get customPropertyWithReadOnlyGetter() {
```
rictic: ooh, subtle. +1

Issue #748, opened 2017-11-07
Analyze templates in JS strings


justinfagnani:

rictic:
There are a number of subtle issues here. I'm starting with doing the simplest possible, thing in order to work through the rough mechanics. e.g. there are a few places where we assume that HTML is never inline, and that all inline documents are inline in HTML.

Looks like we'll need a more complicated system for mapping source ranges from inside the document to outside. 

Once that's working there's:
  * how do we know that a template string should be parsed as HTML?
  * We need a method on a Element to get its template. Should probably also include information about databinding or other framework specific customizations.

justinfagnani:
Another but subtlety is how to handle expressions. lit-html requires that expressions are only in valid text positions: text content or attribute values. Polymer, because interpolation happens before template parsing, actually allows expressions anywhere, meaning the literal parts so not have to be well formed.

I think it's probably simplest to assume the same constraints lit-html uses and that literals are well formed and collapse expressions to empty strings.

PR #819, opened 2017-12-28
 Add basic support for inline HTML in JS

rictic:

Part of https://github.com/Polymer/polymer-analyzer/issues/748

### DIFF
--- a/src/html/html-document.ts
+++ b/src/html/html-document.ts
@@ -187,18 +187,27 @@ export class ParsedHtmlDocument extends ParsedDocument<ASTNode, HtmlVisitor> {
     const selfClone = mutableDocuments.shift()!;
 
     for (const doc of mutableDocuments) {
+      if (doc.astNode == null || doc.astNode.language !== 'html') {
+        continue;
+      }
+      const docContainingNode = doc.astNode.node;
+      const docContainingLocation = docContainingNode.__location;
       let expectedIndentation;
-      if (doc.astNode.__location) {
-        expectedIndentation = doc.astNode.__location.col;
+      if (docContainingLocation &&
+          !isElementLocationInfo(docContainingLocation)) {
+        expectedIndentation = docContainingLocation.col;
 
-        if (doc.astNode.parentNode && doc.astNode.parentNode.__location) {
-          expectedIndentation -= doc.astNode.parentNode.__location.col;
+        const parentLocation = docContainingNode.parentNode &&
+            docContainingNode.parentNode.__location;
+        if (parentLocation && !isElementLocationInfo(parentLocation)) {
+          expectedIndentation -= parentLocation.col;
         }
-      } else {
+      }
+      if (expectedIndentation === undefined) {
         expectedIndentation = 2;
       }
 
-      dom5.setTextContent(doc.astNode, '\n' + doc.stringify({
+      dom5.setTextContent(docContainingNode, '\n' + doc.stringify({
         indent: expectedIndentation
       }) + '  '.repeat(expectedIndentation - 1));
     }
--- a/src/html/html-document.ts
+++ b/src/html/html-document.ts
@@ -186,9 +186,18 @@ export class ParsedHtmlDocument extends ParsedDocument<ASTNode, HtmlVisitor> {
     const mutableDocuments = clone(immutableDocuments);
     const selfClone = mutableDocuments.shift()!;
 
-    for (const doc of mutableDocuments) {
+    // We must handle documents that are inline to us but mutated here.
+    // If they're not inline us, we'll pass them along to our child documents
+    // when stringifying them.
+    const [ourInlineDocuments, otherDocuments] = partition(
+        mutableDocuments,
+        (d) => d.astNode != null && d.astNode.containingDocument === selfClone);
+
+    for (const doc of ourInlineDocuments) {
       if (doc.astNode == null || doc.astNode.language !== 'html') {
-        continue;
+        throw new Error(
+            `This should not happen, ` +
+            `we already checked for this condition in partition()`);
       }
       const docContainingNode = doc.astNode.node;
       const docContainingLocation = docContainingNode.__location;
@@ -208,7 +217,8 @@ export class ParsedHtmlDocument extends ParsedDocument<ASTNode, HtmlVisitor> {
       }
 
       dom5.setTextContent(docContainingNode, '\n' + doc.stringify({
-        indent: expectedIndentation
+        indent: expectedIndentation,
+        inlineDocuments: otherDocuments
       }) + '  '.repeat(expectedIndentation - 1));
     }
 
@@ -309,3 +319,21 @@ const voidTagNames = new Set([
   'track',
   'wbr'
 ]);
+
+/**
+ * Returns two arrays. One of items that match the predicate, the other of items
+ * that do not.
+ */
+function partition<T>(
+    items: Iterable<T>, predicate: (val: T) => boolean): [T[], T[]] {
+  const matched = [];
+  const notMatched = [];
+  for (const item of items) {
+    if (predicate(item)) {
+      matched.push(item);
+    } else {
+      notMatched.push(item);
+    }
+  }
+  return [matched, notMatched];
+}
--- a/src/html/html-style-scanner.ts
+++ b/src/html/html-style-scanner.ts
@@ -63,7 +63,7 @@ export class HtmlStyleScanner implements HtmlScanner {
               locationOffset,
               commentText,
               document.sourceRangeForNode(node)!,
-              {language: 'html', node}));
+              {language: 'html', node, containingDocument: document}));
         }
       }
       // Descend into templates.
--- a/src/html/html-style-scanner.ts
+++ b/src/html/html-style-scanner.ts
@@ -54,7 +54,8 @@ export class HtmlStyleScanner implements HtmlScanner {
               true));
         } else {
           const contents = dom5.getTextContent(node);
-          const locationOffset = getLocationOffsetOfStartOfTextContent(node);
+          const locationOffset =
+              getLocationOffsetOfStartOfTextContent(node, document);
           const commentText = getAttachedCommentText(node) || '';
           features.push(new ScannedInlineDocument(
               'css',
--- a/src/html/html-style-scanner.ts
+++ b/src/html/html-style-scanner.ts
@@ -62,7 +62,7 @@ export class HtmlStyleScanner implements HtmlScanner {
               locationOffset,
               commentText,
               document.sourceRangeForNode(node)!,
-              node));
+              {language: 'html', node}));
         }
       }
       // Descend into templates.
--- a/src/javascript/html-template-literal-scanner.ts
+++ b/src/javascript/html-template-literal-scanner.ts
@@ -91,10 +91,35 @@ export function getInlineDocument(
     if (previousEnd !== undefined) {
       const fullExpressionTextWithDelimitors =
           parsedDocument.contents.slice(previousEnd, quasi.start);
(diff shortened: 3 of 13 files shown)


rictic:
The commits are pretty well separated and might be easier to review in pieces. Expression handling is refined from its initial implementation, but otherwise they don't overlap much.

src/javascript/html-template-literal-scanner.ts
```diff
@@ -0,0 +1,104 @@
...
+import * as babel from 'babel-types';
+
+import {ScannedInlineDocument} from '../model/model';
+
+import {getIdentifierName} from './ast-value';
+import {Visitor} from './estree-visitor';
+import {JavaScriptDocument} from './javascript-document';
+import {JavaScriptScanner} from './javascript-scanner';
+
+export class InlineHtmlDocumentScanner implements JavaScriptScanner {
```
justinfagnani: Comment

src/javascript/html-template-literal-scanner.ts
```diff
@@ -0,0 +1,104 @@
...
+      }
+    };
+
+    await visit(myVisitor);
+
+    return {features};
+  }
+}
+
+export function getInlineDocument(
```
justinfagnani: comment

src/javascript/html-template-literal-scanner.ts
```diff
@@ -0,0 +1,104 @@
...
+    node: babel.TaggedTemplateExpression,
+    parsedDocument: JavaScriptDocument): ScannedInlineDocument|undefined {
+  const tagName = getIdentifierName(node.tag);
+  if (tagName === undefined || !/(^|\.)html$/.test(tagName)) {
```
justinfagnani: Even at this early stage of this feature, I'd take the tag recognizer as a constructor argument to the scanner. That makes it pretty trivial to construct new scanners for different tags.

src/javascript/html-template-literal-scanner.ts
```diff
@@ -0,0 +1,104 @@
...
+    return;
+  }
+
+  const sourceRangeForLiteral = parsedDocument.sourceRangeForNode(node.quasi);
+  if (sourceRangeForLiteral === undefined) {
+    return;
+  }
+  const sourceRangeForContents = {
```
justinfagnani: If SourceRange were a class it could have some nice transformation methods...

src/javascript/html-template-literal-scanner.ts
```diff
@@ -0,0 +1,104 @@
...
+    if (previousEnd !== undefined) {
+      const fullExpressionTextWithDelimitors =
+          parsedDocument.contents.slice(previousEnd, quasi.start);
+      // Replace everything but whitespace in ${expressions} (including the
+      // ${} delimitor part) with whitespace.
+      // This skips over the problem of handling expressions, and there's lots
+      // of cases it doesn't handle correctly, but it's a start.
+      contents += fullExpressionTextWithDelimitors.replace(/\S/g, ' ');
+    }
+    contents += quasi.value.cooked;
```
justinfagnani: I think we'll need to configure whether to use cooked or raw. `Polymer.html` uses the raw strings, lit-html currently uses cooked, but might change to raw.

src/javascript/html-template-literal-scanner.ts
```diff
@@ -0,0 +1,104 @@
...
+  };
+
+  let contents = '';
+  let previousEnd: number|undefined;
+  for (let i = 0; i < node.quasi.quasis.length; i++) {
+    const quasi = node.quasi.quasis[i];
+    if (previousEnd !== undefined) {
+      const fullExpressionTextWithDelimitors =
+          parsedDocument.contents.slice(previousEnd, quasi.start);
+      // Replace everything but whitespace in ${expressions} (including the
```
justinfagnani: I don't quite understand this. Can't we just skip the whole expression? We're assuming well-formed HTML in the literal parts (which should be documented). I'm not sure we can put anything useful in the contents.

src/javascript/html-template-literal-scanner.ts
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  let commentText;
+  if (node.leadingComments != null) {
+    commentText = node.leadingComments.map((c) => c.value).join('\n');
+  } else {
+    commentText = '';
+  }
+
+  return new ScannedInlineDocument(
```
justinfagnani: I think we should have a subclass here that carries the concatenated string representation, like you've built up, but also the literal and expression parts. For some lit-html analysis we'll want to find which attribute are bound to, and we'll need that information.

src/javascript/html-template-literal-scanner.ts
rictic: On reflection, I don't think this belongs here. Moved it up into the scanner.

src/javascript/html-template-literal-scanner.ts
rictic: +1. Punting for now though.

src/javascript/html-template-literal-scanner.ts
rictic: Done

src/javascript/html-template-literal-scanner.ts
rictic: +1 but I want to cross that bridge when we get there. I'm not sure what that will look like yet or what will be useful. My goal with this PR is to get a first minimally useful thing out there, then to hook it up with polymer element parsing, then to start getting fancy with better handling of the template literal specific parts.

src/javascript/html-template-literal-scanner.ts
justinfagnani: Why do you need to replace with space or newline, and even so, couldn't this be a map over the quasis?

```ts
const contents = node.quasi.quasis.join('');
```

rictic: Consider the js file:

```js
html`<div>${
    'Hello world'
  }</div>
```

If we remove the expression entirely, the html parser receives `<div></div>` and when we ask for the source range of the closing tag it'll give one on the first line, and starting just after the `<div>`. By preserving newlines and replacing every other character with a space, the source map will correctly be able to tell it starts on the third line, fourth character.

justinfagnani: Got it. Add that explanation to the comments ;)

ChadKillingsworth:
I've been able to use this branch to get the analyzer functioning with webpack. One issue though: the analyzer selects parsing types by file extension with no way to override that. While there is a mechanism to specify custom parsers, this then breaks inline document parsing. I'm having to monkey patch the AnalyzerContext class to support this use case.

```js
/**
 * Caching + loading wrapper around _parseContents.
 */
AnalysisContext.prototype._parse = async function(resolvedUrl) {
  return this._cache.parsedDocumentPromises.getOrCompute(resolvedUrl, async () => {
    const content = await this.load(resolvedUrl);
    // const extension = path.extname(resolvedUrl).substring(1);
    return this._parseContents('js', content, resolvedUrl);
  });
};
```

rictic:
@ChadKillingsworth Can you expand on your use case? You have JS files with inline HTML, but you're naming those files with a `.html` file extension?

ChadKillingsworth:
@rictic Correct. In webpack + Polymer 2, HTML files are converted to JS strings and are registered / added to the dom at runtime. By the time the analyzer is running, the HTML has been converted to JS by a webpack loader, but the filename and extension is not modified.

I just need to be able to specify the type independent of the file extension.

ChadKillingsworth:
This will be the same mechanism I intend to use with Polymer 3 to support the HTML template files being in a separate file with a `.html` extension.

justinfagnani:
And your analyzing post WebPack, not pre-WebPack? The analyzer definitely works better on source, as the patterns it recognizes can be obscured by certain transformations.

ChadKillingsworth:
I'm analyzing mid-webpack. There's no way to do this pre or post as the analyzer doesn't understand commonjs modules, node module resolution or handle HTML files which have weird looking HTML imports.

As a webpack plugin, it was amazingly easy to write a custom scanner to recognize `__webpack_require__(42)` and simply add a `ScannedImport` feature with the full path.

src/javascript/html-template-literal-scanner.ts
rictic: Done

src/javascript/html-template-literal-scanner.ts
```diff
@@ -0,0 +1,151 @@
...
+    },
+    end: {
+      line: sourceRangeForLiteral.end.line,
+      column: sourceRangeForLiteral.end.column - 1
+    }
+  };
+
+  let contents = '';
+  let previousEnd: number|undefined;
+  for (let i = 0; i < node.quasi.quasis.length; i++) {
```
justinfagnani: use for/of?

src/html/html-style-scanner.ts
```diff
@@ -54,15 +54,16 @@ export class HtmlStyleScanner implements HtmlScanner {
...
+              getLocationOffsetOfStartOfTextContent(node, document);
           const commentText = getAttachedCommentText(node) || '';
           features.push(new ScannedInlineDocument(
               'css',
               contents,
               locationOffset,
               commentText,
               document.sourceRangeForNode(node)!,
-              node));
+              {language: 'html', node}));
```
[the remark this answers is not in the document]
rictic: Generally we try not to have up references like that, don't we? 

I do think features more generally having a pointer to their Document and ParsedDocument would often be pretty useful though.

In any case, this turned out to be quite useful in stringify, so I added `containingDocument` here. I kept `language` because TypeScript doesn't do type narrowing as well as we'd like:

https://www.typescriptlang.org/play/#src=interface%20Doc%3CNode%3E%20%7B%0D%0A%20%20%20%20readonly%20body%3A%20ReadonlyArray%3CNode%3E%3B%0D%0A%7D%0D%0Ainterface%20NodeAndDoc%3CNode%3E%20%7B%0D%0A%20%20%20%20readonly%20node%3A%20Node%3B%0D%0A%20%20%20%20readonly%20doc%3A%20Doc%3CNode%3E%3B%0D%0A%7D%3B%0D%0A%0D%0Adeclare%20const%20container%3A%20NodeAndDoc%3Cstring%7Cnumber%3E%3B%0D%0A%0D%0Aif%20(typeof%20container.node%20%3D%3D%3D%20'string')%20%7B%0D%0A%20%20%20%20const%20node%20%3D%20container.node%3B%0D%0A%20%20%20%20const%20firstChild%20%3D%20container.doc.body%5B0%5D%3B%0D%0A%20%20%20%20%2F%2F%20node%20and%20firstChild%20must%20both%20be%20strings%2C%20but%20typescript%0D%0A%20%20%20%20%2F%2F%20still%20thinks%20firstChild%20could%20be%20a%20number.%0D%0A%7D%0D%0A

src/javascript/html-template-literal-scanner.ts
rictic: Done

src/html/html-document.ts
```diff
@@ -187,18 +187,27 @@ export class ParsedHtmlDocument extends ParsedDocument<ASTNode, HtmlVisitor> {
     const selfClone = mutableDocuments.shift()!;
 
     for (const doc of mutableDocuments) {
+      if (doc.astNode == null || doc.astNode.language !== 'html') {
```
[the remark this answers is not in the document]
justinfagnani: That shouldn't be needed if you check for `undefined`. How would `null` ever get in there?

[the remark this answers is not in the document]
rictic: It's typed as `|null`, which is probably wrong. Will check on it when I get strict property checks working.

PR #820, opened 2017-12-28
UrlResolver#resolve(baseUrl, fileRelativeUrl)

usergenic:
We talked about reversing `UrlResolver#relative()` to make it `(to, from)` but I thought I'd apply the straw-man gambit and take a stab at flipping around `UrlResolver#resolve(baseUrl, url)` instead.

One nice side effect of this is that the type of the unresolvedUrl is known based on whether optional baseUrl is provided, since the typings of the two method forms are:

1. `UrlResolver#resolve(baseUrl: ResolvedUrl, unresolvedUrl: FileRelativeUrl)`
2. `UrlResolver#resolve(unresolvedUrl: PackageRelativeUrl)`

* [BREAKING]: `UrlResolver#resolve()` argument order swapped so that the optional `baseUrl` argument comes first instead of second.  This makes resolve more similar to `url.resolve`.

### DIFF
--- a/src/test/url-loader/redirect-resolver_test.ts
+++ b/src/test/url-loader/redirect-resolver_test.ts
@@ -25,12 +25,12 @@ suite('RedirectResolver', function() {
           new RedirectResolver(resolvedUrl``, 'proto://site/', 'some/path/');
       assert.equal(
           resolver.resolve(
-              fileRelativeUrl`proto://site/something.html`, resolvedUrl``),
+              resolvedUrl``, fileRelativeUrl`proto://site/something.html`),
           resolvedUrl`some/path/something.html`);
       resolver = new RedirectResolver(resolvedUrl``, '/site/', 'some/path/');
       assert.equal(
           resolver.resolve(
-              fileRelativeUrl`/site/something.html`, resolvedUrl``),
+              resolvedUrl``, fileRelativeUrl`/site/something.html`),
           resolvedUrl`some/path/something.html`);
     });
 
@@ -39,7 +39,7 @@ suite('RedirectResolver', function() {
           new RedirectResolver(resolvedUrl``, 'proto://site/', 'some/path/');
       assert.equal(
           resolver.resolve(
-              fileRelativeUrl`protoz://site/something.html`, resolvedUrl``),
+              resolvedUrl``, fileRelativeUrl`protoz://site/something.html`),
           undefined);
     });
   });
--- a/src/test/url-loader/redirect-resolver_test.ts
+++ b/src/test/url-loader/redirect-resolver_test.ts
@@ -15,7 +15,7 @@
 import {assert} from 'chai';
 
 import {RedirectResolver} from '../../url-loader/redirect-resolver';
-import {fileRelativeUrl, resolvedUrl} from '../test-utils';
+import {packageRelativeUrl, resolvedUrl} from '../test-utils';
 
 
 suite('RedirectResolver', function() {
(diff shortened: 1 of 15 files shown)


src/test/url-loader/redirect-resolver_test.ts
```diff
@@ -25,12 +25,12 @@ suite('RedirectResolver', function() {
           new RedirectResolver(resolvedUrl``, 'proto://site/', 'some/path/');
       assert.equal(
           resolver.resolve(
-              fileRelativeUrl`proto://site/something.html`, resolvedUrl``),
+              resolvedUrl``, fileRelativeUrl`proto://site/something.html`),
```
rictic: Can just make these calls the one-argument packageRelativeUrl form. I think it's clearer and more closely aligned with the intent of the tests

src/url-loader/package-url-resolver.ts
```diff
@@ -57,10 +57,12 @@ export class PackageUrlResolver extends UrlResolver {
   }
 
   resolve(
-      unresolvedHref: FileRelativeUrl|PackageRelativeUrl,
-      baseUrl: ResolvedUrl = this.packageUrl,
+      firstHref: ResolvedUrl|PackageRelativeUrl, secondHref?: FileRelativeUrl,
       _import?: ScannedImport): ResolvedUrl|undefined {
-    const resolvedHref = this.simpleUrlResolve(unresolvedHref, baseUrl);
+    const [baseUrl, unresolvedHref] = typeof secondHref === 'undefined' ?
```
rictic: Prefer `secondHref === undefined`, it's likely to be much faster, and it's what we do everywhere else

rictic: This logic is pretty universally used in url resolvers, and it is a bit subtle. What about putting it in a protected method on UrlResolver? `getBaseAndUnresolved`?

src/test/url-loader/redirect-resolver_test.ts
usergenic: Good point.  Done.

src/url-loader/package-url-resolver.ts
usergenic: Ok.  This `typeof x === 'undefined'` of mine is an old habit.  Will switch!

usergenic: I like that idea!  Added `getBaseAndUnresolved`.

src/test/url-loader/url-resolver_test.ts
```diff
@@ -20,9 +20,8 @@ import {UrlResolver} from '../../url-loader/url-resolver';
 class SimplestUrlResolver extends UrlResolver {
   resolve(
       firstUrl: ResolvedUrl|PackageRelativeUrl, secondUrl?: FileRelativeUrl) {
-    const [baseUrl, url] = secondUrl === undefined ?
-        [this.brandAsResolved('/test/'), firstUrl as PackageRelativeUrl] :
-        [this.brandAsResolved(firstUrl), secondUrl];
+    const [baseUrl = '/test/' as ResolvedUrl, url] =
```
rictic: Woah, I had no idea that syntax was legal.

I'm a little weirded out by it, but it does make sense. Would it be clearer if `getBaseAndUnresolved`  took a third argument (maybe optionally?), a resolved package url?

usergenic: Personally, I like this better because the code itself demonstrates expectation of a possible undefined and it can be handled how the client wants?  I'll merge as is but if it still feels weird we can make the change.

PR #821, opened 2017-12-29
detect methods/properties added to the prototype

43081j:
this is an attempt at detecting methods and properties added to the prototype of a class.

Fixes polymer/polymer#5010 assuming polymer/polymer#5000 gets merged.

**This should be considered WIP for now as some of the code could be cleaned up and people may disagree with the implementation. Do give feedback.**

---

These are what will be detected:

```ts
Class.prototype.prop = 'foo';

Class.prototype.fn = function(a, b) {
  return a+b;
};

/**
 * @function
 * @return {boolean}
 */
Class.prototype.annotatedFn;

/** @type {boolean} */
Class.prototype.annotatedProp;
```

---

The basic logic:

* For a given class, search the document for the supported expressions
* If `@function` is in the annotation, try to scan it as a method
* Otherwise, try scan it as a property
* If it is an assignment, try get the `type` from the value unless the annotation already specifies it

cc @rictic @aomarks


src/javascript/class-scanner.ts
```diff
@@ -412,6 +605,24 @@ class ClassFinder implements Visitor {
 
     const warnings: Warning[] = [];
     const properties = extractPropertiesFromClass(astNode, this._document);
+    const methods = getMethods(astNode, this._document);
+
+    if (name) {
+      const protoFinder = new PrototypeMemberFinder(this._document, name);
+      this._document.visit([protoFinder]);
```
43081j: this feels wrong to do. the finder does belong inside the class finder i think but i wasn't sure if `document.visit` is fine to use...

src/javascript/class-scanner.ts
```diff
@@ -335,6 +335,199 @@ interface CustomElementDefinition {
...
+    const left = node.expression.object;
+
+    // we only want `something.prototype.member`
+    if (!babel.isMemberExpression(left) ||
+        !babel.isIdentifier(left.object) ||
+        !babel.isIdentifier(left.property) ||
+        !babel.isIdentifier(node.expression.property) ||
+        left.property.name !== 'prototype' ||
+        left.object.name !== this._name) {
+      return;
```
43081j: is this the easiest way to do this? what about statics too? `Class.Foo;`. worth detecting them?

src/javascript/class-scanner.ts
rictic: This implementation seems like it's optimized to avoid creating unnecessary candidate properties, but I think the expensive part is going to be walking the AST, and creating candidate properties is super cheap.

I'd recommend instead creating a single PrototypeProperty finder that finds all such properties on all classes, and run that visitor at the same time as the other visitors in the Promise.all call at ~line 77.

Then we can match up the properties and methods with the found classes. Probably good to put the properties and methods in a map keyed by the class name since that's how we'll recognize them.

43081j: that is exactly what i did in my first cut, but moved it to here :joy: 

i just thought it seemed a little confusing because a class could come after a prototype expression i suppose. or should we assume it never does? and just check `classes` for matching class names.

43081j: to avoid having to check in both the finder and the class detection logic i mean

src/javascript/class-scanner.ts
rictic: This is the fastest way of doing it, though there's a more convenient way of doing it with `getIdentifierName` from ast-value.

Let's start with instance properties in this PR. Statics would be nice to get too though.

src/javascript/class-scanner.ts
```diff
@@ -335,6 +347,221 @@ interface CustomElementDefinition {
...
+  private _addPropertyToClass(cls: string, member: ScannedProperty) {
+    let classMembers = this.members.get(cls);
+    if (!classMembers) {
+      classMembers = {
+        methods: new Map<string, ScannedMethod>(),
+        properties: new Map<string, ScannedProperty>()
+      };
+      this.members.set(cls, classMembers);
+    }
+    classMembers.properties.set(member.name, member);
```
43081j: @rictic the question about unifying two methods was these two.. they are identical except the last line. so i was trying to detect if its a method or a property.

src/model/class.ts
```diff
@@ -81,6 +81,22 @@ export class ScannedClass implements ScannedFeature, Resolvable {
...
+      methods: Map<string, ScannedMethod>,
+      properties: Map<string, ScannedProperty>) {
+    const mutableMethods = unsafeAsMutable(this.methods);
+    for (const [name, method] of methods) {
+      if (!mutableMethods.has(name)) {
+        mutableMethods.set(name, method);
+      }
+    }
+    for (const [name, prop] of properties) {
+      if (!this.properties.has(name)) {
```
43081j: we currently skip methods/properties which already exist. i did this because something we picked up on the prototype is less reliable than an actual class method or property we already found i guess. though if we ever tried using this method elsewhere, that could be confusing.

src/javascript/class-scanner.ts
rictic: What about a MapWithDefault class to cover the common code? Something like:

```typescript
class MapWithDefault<K, V> extends Map<K, V> {
  constructor(private readonly makeDefault: (key: K) => V) {
    super();
  }

  get(key: K): V {
    if (!this.has(key)) {
      this.set(key, this.makeDefault(key));
    }
    return super.get(key)!;
  }
}
```

src/model/class.ts
rictic: Hm, is it less reliable? Something on the prototype must have come after the class declaration, so it should override, yeah?

src/javascript/class-scanner.ts
```diff
@@ -335,6 +347,221 @@ interface CustomElementDefinition {
...
+  readonly members = new Map<string,
+    {methods: Map<string, ScannedMethod>, properties: Map<string, ScannedProperty>}>();
+  private readonly _document: JavaScriptDocument;
+
+  constructor(document: JavaScriptDocument) {
+    this._document = document;
+  }
+
+  enterExpressionStatement(node: babel.ExpressionStatement) {
+    if (babel.isAssignmentExpression(node.expression)) {
```
rictic: If you pull `node.expression` out into a local const then `babel.isAssignmentExpression` and friends will narrow its type, so you can avoid checking the type again inside your `_createMemberFrom...` methods

src/javascript/class-scanner.ts
```diff
@@ -335,6 +347,221 @@ interface CustomElementDefinition {
...
+
+    const cls = getIdentifierName(left.object);
+
+    if (!cls) {
+      return;
+    }
+
+    const annotation = getJSDocAnnotationForNode(node);
+
+    if (jsdoc.hasTag(annotation, 'function')) {
```
rictic: What about code like:

```javascript
/**
 * Flips the floob
 * @param {Direction} direction The direction to flip.
 * @return {Floob} The flipped floob
 */
Floob.prototype.flip;
```

We could also check for `param` `return` or `returns`

src/javascript/class-scanner.ts
```diff
@@ -335,6 +347,221 @@ interface CustomElementDefinition {
...
+    const sourceRange = this._document.sourceRangeForNode(node)!;
+    const warnings: Warning[] = [];
+
+    if (jsdocAnn) {
+      const typeTag = jsdoc.getTag(jsdocAnn, 'type');
+      if (typeTag && typeTag.type) {
+        type = doctrine.type.stringify(typeTag.type);
+      }
+      description = getDescription(jsdocAnn);
+      privacy = getOrInferPrivacy(name, jsdocAnn);
```
rictic: We should use get or infer privacy even if we don't have a jsdoc, so that we pay attention to the name of the property

src/javascript/class-scanner.ts
```diff
@@ -335,6 +347,221 @@ interface CustomElementDefinition {
...
+      ret = getReturnFromAnnotation(jsdocAnn);
+      type = ret ? ret.type : undefined;
+
+      if (babel.isFunctionExpression(node)) {
+        (node.params || []).forEach((nodeParam) => {
+          const param = toMethodParam(nodeParam, jsdocAnn);
+          params.set(param.name, param);
+        });
+      } else {
+        for (const tag of (jsdocAnn.tags || [])) {
```
rictic: We should run this even if it's a function expression. The documentation may have more info on the parameters, like type and description.

src/model/class.ts
```diff
@@ -81,6 +81,22 @@ export class ScannedClass implements ScannedFeature, Resolvable {
   resolve(document: Document): Feature|undefined {
     return new Class(this, document);
   }
+
+  finishInitialization(
```
rictic: Document what's up with this method please.

src/javascript/class-scanner.ts
43081j: i've instead made the methods take what they say they take. it was a bit confusing before, being called `createMemberFromAssignment` but not taking an `AssignmentExpression`... so now we just pass the expression in.

src/javascript/class-scanner.ts
43081j: `toMethodParam` does a very similar bit of logic already. so we do pick up types and descriptions from the annotations through that

src/model/class.ts
43081j: that is very true, ill remove this check

43081j:
@rictic i've added a test for private members with/without jsdoc. also resolved the remaining issues you raised :)

src/javascript/class-scanner.ts
```diff
@@ -361,43 +361,38 @@ class PrototypeMemberFinder implements Visitor {
...
-      this._createMemberFromMemberExpression(node);
+      this._createMemberFromMemberExpression(node.expression,
+        getJSDocAnnotationForNode(node));
     }
   }
 
-  private _createMemberFromAssignment(node: babel.ExpressionStatement) {
-    if (!babel.isAssignmentExpression(node.expression)) {
+  private _createMemberFromAssignment(
+      node: babel.AssignmentExpression, jsdocAnn?: jsdoc.Annotation) {
```
rictic: Why is the second arg optional, we always pass it?

If it's because it could be undefined, prefer `jsdoc.Annotation|undefined`, as that way we'll be less likely to accidentally leave off that argument.

src/javascript/class-scanner.ts
```diff
@@ -385,41 +388,27 @@ class PrototypeMemberFinder implements Visitor {
     }
 
     if (babel.isFunctionExpression(node.right)) {
-      const method = this._createMethodFromExpression(leftProperty.name,
+      const member = this._createMethodFromExpression(leftProperty.name,
```
rictic: Super minor, but `method` and `prop` are clearer to me than the generic `member`, when we do know what kind of thing it is.

43081j:
here you go @rictic.

FYI yes it was because `jsdocAnn` could be undefined. i agree its better to be explicit, don't make it seem optional

rictic:
Oh, upstream closureType changed to getClosureType, and it returns a Result<string, Warning> now :X

43081j:
ah yes i saw that! ill rebase for you. im very glad you made that change haha, makes much more sense

43081j:
rebased + did a little change in `getClosureType` so we don't require an annotation @rictic

src/javascript/class-scanner.ts
```diff
@@ -466,12 +466,12 @@ class PrototypeMemberFinder implements Visitor {
     if (babel.isAssignmentExpression(node)) {
       if (!type) {
         const detectedType =
-            closureType(node.right, sourceRange, this._document);
-        if (detectedType instanceof Warning) {
-          warnings.push(detectedType);
+            getClosureType(node.right, jsdocAnn, sourceRange, this._document);
+        if (!detectedType.successful) {
+          warnings.push(detectedType.error);
```
rictic: I think we should drop the warning here. A method's type is Function and that's fine, yeah?

PR #822, opened 2017-12-29
Add a field for getting the template of an element.

rictic:

Related to #748 as currently we get the template of an element manually in a bunch of places in the linter, but this is both ugly, and something that will be broken by other ways of declaring the template.


src/model/element-base.ts
```diff
@@ -105,6 +106,25 @@ export interface ElementBaseInit extends ClassInit {
   readonly slots?: Slot[];
 }
 
+/**
+ * The element's runtime contents.
+ */
+export type ElementTemplate = {
+  /**
+   * HTML that is stamped out without data binding or other
```
justinfagnani: do we actually have this anywhere?

src/model/element-base.ts
```diff
@@ -105,6 +106,25 @@ export interface ElementBaseInit extends ClassInit {
...
+   * interpretation beyond normal HTML semantics.
+   */
+  kind: 'html',
+  contents: ASTNode,
+}|{
```
justinfagnani: This union either seems unnecessary or the content types are weird. For HTML-in-JS wouldn't we have a JS AST node, or a string? Or is this post-parsing by parse5? Is there a way to get back to the real source in case we want to modify it?

src/model/element-base.ts
rictic: Not yet, but I wanted to have a type for when we do add this.

src/model/element-base.ts
rictic: This is post-parsing by parse5. In a follow-up PR I'm gonna add more affordances for working with this, including getting its parsed document for getting the original source, modifying, etc.

PR #823, opened 2017-12-30
Renamed closureType to getClosureType, return a Result.

rictic:

Cleans up some common code and an awkward API I noticed in https://github.com/Polymer/polymer-analyzer/pull/821

### DIFF
--- a/src/polymer/js-utils.ts
+++ b/src/polymer/js-utils.ts
@@ -13,9 +13,8 @@
  */
 
 import * as babel from 'babel-types';
-import * as doctrine from 'doctrine';
 
-import {closureType, configurationProperties, getAttachedComment, getOrInferPrivacy, objectKeyToString} from '../javascript/esutil';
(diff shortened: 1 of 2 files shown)


src/polymer/js-utils.ts
```diff
@@ -48,15 +47,15 @@ export function toScannedPolymerProperty(
...
-  }
-  if (type instanceof Warning) {
-    warnings.push(type);
+  const typeResult = getClosureType(value, parsedJsdoc, sourceRange, document);
+  let type;
+  if (typeResult.successful) {
+    type = typeResult.value;
+  } else {
+    warnings.push(typeResult.error);
     type = 'Object';
```
aomarks: Should this be `*` by default? Closure `Object` excludes primitives.

rictic: If it's a primitive the user should have passed in the primitive constructor, like `String`, `Boolean`, etc

PR #826, opened 2018-01-01
detect return type of methods when not annotated

### LINKED ISSUE #753: Feature request: infer void return type
A function with no `return` statements (or only return statements with no argument) in it, which is neither a generator or an async function, can be inferred to have a return type of void.

Many functions on elements have a void return type, so this could save a lot of typing.

Request came from slack

43081j:
Fixes #753.

This is just something I figured i'd do while i was in there. If we don't want it then this can be closed.

## Work in progress

At the min it doesn't pass tests because it no longer detects `Polymer.Base` methods in the polymer-core tests... I can't seem to track down why adding `return` to some methods causes that failure. Commenting out my changes in `esutil` makes the test pass, confusingly.

Also what do we do about inheritance? what if we detect a method as `void` but the one it is overriding in the base class is typed as something else?

---

What it does:

* If there is no `@return`, it tries to infer it
* If there is no return statement, it is inferred as `void`
* If there is a return statement, or multiple, it tries to get the type (via `closureType`) and joins them
* If one or more of the `closureType` calls create warnings, we don't add a return type at all because its better to not have one than to get it wrong i think

### DIFF
--- a/src/javascript/esutil.ts
+++ b/src/javascript/esutil.ts
@@ -14,6 +14,7 @@
 
 import generate from 'babel-generator';
 import * as babel from 'babel-types';
+import traverse from 'babel-traverse';
 import * as doctrine from 'doctrine';
 
 import {MethodParam, ScannedMethod} from '../index';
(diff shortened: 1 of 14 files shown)


justinfagnani:
I don't think we want to continue down the road of implementing our own inference. I'm personally OK with having very little inference and requiring annotations if you want better docs, TS types, etc. But if we _do_ want inference, I'd rather lean on some existing library's work. TypeScript's support of jsdoc annotations could be useful here.

43081j:
yup i do agree, which is why i went ahead and added the missing type annotations to polymer recently.

while this is kinda nice, it doesn't really feel good trying to guess types. 

its a fairly small PR anyhow so i dont mind closing it if we're all in agreement, at least for its current purpose.

43081j:
closing this as i think its better we just expect good annotations rather than try to infer things.

justinfagnani:
I agree, but thanks for investigating!

PR #827, opened 2018-01-02
A couple minor changes for browserify compatibility.

rictic:
- [x] CHANGELOG.md has been updated, this is a noop for existing users

### DIFF
--- a/src/analysis-format/generate-analysis.ts
+++ b/src/analysis-format/generate-analysis.ts
@@ -12,9 +12,7 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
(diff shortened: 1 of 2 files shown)


justinfagnani:
I'd rather move gradually away from node-only APIs like `require()`. Loading a file from a URL is an abstraction we already have too.

rictic:
True, but we have no way to express a synchronous dependency on some JSON. If we made schema validation asynchronous (so it can load the schema the first time it's called) that would be fine with me.

rictic:
Happy to hash out the longer term plan for loading the schema JSON, but I'd like to land this now.

PR #830, opened 2018-01-04
Make the prototype scanner pick up namespaces

43081j:
This is an attempt at fixing the lack of `@memberof` understanding in the prototype scanner.

As usual, it'll help us finalise polymer/polymer#5000.

---

The problem was:

```ts
/**
 * @memberof Polymer
 */
class MyClass {}

/** @function */
MyClass.prototype.MyMethod;
```

This is currently scanned like so:

* The class is named `Polymer.MyClass`
* The method thinks its class is called `MyClass`
* During class finalisation, it therefore can't find `MyClass` in the scanned classes

So the fix i figure is to respect `@memberof` on methods/properties too.

If this doesn't seem semantically correct, do speak up.

cc @aomarks @rictic

### DIFF
--- a/src/javascript/class-scanner.ts
+++ b/src/javascript/class-scanner.ts
@@ -385,21 +385,24 @@ class PrototypeMemberFinder implements Visitor {
     const leftExpr = node.left.object;
     const leftProperty = node.left.property;
     const cls = getIdentifierName(leftExpr.object);
+
     if (!cls || getIdentifierName(leftExpr.property) !== 'prototype') {
       return;
     }
+
+    const namespacedCls = getNamespacedIdentifier(cls, jsdocAnn);
 
     if (babel.isFunctionExpression(node.right)) {
       const prop = this._createMethodFromExpression(
           leftProperty.name, node.right, jsdocAnn);
       if (prop) {
-        this._addMethodToClass(cls, prop);
+        this._addMethodToClass(namespacedCls, prop);
       }
     } else {
       const method =
           this._createPropertyFromExpression(leftProperty.name, node, jsdocAnn);
       if (method) {
-        this._addPropertyToClass(cls, method);
+        this._addPropertyToClass(namespacedCls, method);
       }
     }
   }
@@ -430,17 +433,19 @@ class PrototypeMemberFinder implements Visitor {
       return;
     }
 
+    const namespacedCls = getNamespacedIdentifier(cls, jsdocAnn);
+
     if (jsdoc.hasTag(jsdocAnn, 'function')) {
       const prop =
           this._createMethodFromExpression(node.property.name, node, jsdocAnn);
       if (prop) {
-        this._addMethodToClass(cls, prop);
+        this._addMethodToClass(namespacedCls, prop);
       }
     } else {
       const method = this._createPropertyFromExpression(
           node.property.name, node, jsdocAnn);
       if (method) {
-        this._addPropertyToClass(cls, method);
+        this._addPropertyToClass(namespacedCls, method);
       }
     }
   }
--- a/src/javascript/class-scanner.ts
+++ b/src/javascript/class-scanner.ts
@@ -368,20 +368,9 @@ interface CustomElementDefinition {
  * When attaching members to a class which has a `@memberof` annotation,
  * the members should also be annotated identically:
  *
- *    @memberof MyNamespace
+ *    @memberof MyNamespace.MyClass.prototype
+ *    @alias MyMethod
  *    MyClass.prototype.MyMethod;
- *
- * One exception to this is when the member is not set directly on
- * the prototype, such as a function which is attached at a later place
- * in the file:
- *
- *    @memberof MyNamespace.MyClass
- *    @instance
- *    function MyMethod() {
- *
- * As you can see, we must specify `@instance` to define the function
- * as an instance method, and `@memberof` to define the class along with
- * its namespace.
  *
  * For more information, see [memberof](http://usejsdoc.org/tags-memberof.html)
  * and [instance](http://usejsdoc.org/tags-instance.html).
@@ -426,7 +415,7 @@ class PrototypeMemberFinder implements Visitor {
       return;
     }
 
-    const namespacedCls = getNamespacedIdentifier(cls, jsdocAnn);
+    const namespacedCls = this._getNamespacedClassName(cls, leftProperty.name, jsdocAnn);
 
     if (babel.isFunctionExpression(node.right)) {
       const prop = this._createMethodFromExpression(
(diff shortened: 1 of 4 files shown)


src/test/static/class/class-memberof.js
```diff
@@ -0,0 +1,18 @@
+/**
+ * @memberof MyNamespace
+ */
+class Class {
+}
+
+/**
+ * @function
+ * @memberof MyNamespace
```
justinfagnani: According to http://usejsdoc.org/tags-memberof.html it looks like we should be doing `@memberof MyNamespace.Class`

justinfagnani (changes requested):
I think this is a fine approach, but the exact use of `@memberof` for class members seems to be a bit off the jsdoc docs.

src/test/static/class/class-memberof.js
43081j: ah! thats what i wasn't sure about. 

i wasn't sure if the `@memberof NS.Class` form was being used only when the associated function isn't defined on the class.

e.g.:

```ts
/** @memberof Namespace */
MyClass.Foo; // becomes Namespace.MyClass.Foo

/** @memberof Namespace.MyClass */
function Foo() { } // somewhere else we add this to MyClass, so we have to tell it the class in memberof, not only the ns. becomes Namespace.MyClass.Foo
```

43081j: @justinfagnani i just ran this through `jsdoc --explain`, i think `Namespace` alone is right but we should also be annotating with `@instance` in polymer otherwise jsdoc detects `scope` as `static`.

so more like:

```ts
/**
 * @memberof Polymer
 * @instance
 * @return {void}
 * @function
 */
MyClass.prototype.MyMethod;
```

however in the analyser we probably don't need to care about the existence of `@instance`. rather that is just something we should include in our annotations.

43081j: yes this is confirmed.

* `@memberof MyName` becomes `MyName.MyClass#MyMethod`
* `@memberof MyName.MyClass` becomes `MyName.MyClass.MyClass#MyMethod`

justinfagnani: Weird. That's fine I guess, but can you document it all with links where helpful?

43081j: sure. i guess it behaves this way because the expression already contains namespace information. whereas in the case of a lone function, it doesn't have any clue. combining an already-namespaced name with a namespace = the duplicate you saw.

src/javascript/class-scanner.ts
```diff
@@ -350,6 +350,42 @@ interface CustomElementDefinition {
...
+ *    @instance
+ *    function MyMethod() {
+ *
+ * As you can see, we must specify `@instance` to define the function
+ * as an instance method, and `@memberof` to define the class along with
+ * its namespace.
+ *
+ * For more information, see [memberof](http://usejsdoc.org/tags-memberof.html)
+ * and [instance](http://usejsdoc.org/tags-instance.html).
+ */
```
43081j: @justinfagnani how is this?

src/test/static/class/class-memberof.js
aomarks: This is pretty confusing!

I spent some time playing around with different forms to see how `jsdoc` would react. The forms that say `GOOD` below render as I would expect in the generated HTML documentation (as a non-static method of `MyNamespace.Class`).

`M1` is the simplest working form that should already work in Analyzer (without this PR), where we just fully qualify the declaration instead of using the local class name. Why don't we do this inside `polymer.dom.html` when declaring the `DomApi` methods (which is the motivation for this PR)?

`M2` is interesting because apparently `jsdoc` is smart enough to follow that `M2` is part of `Class` is part of `MyNamespace` without any annotation hints.

`M3` is the simplest form I could find that has explicit annotations. Note that the `@alias` is required for `jsdoc` (see `M7` for the version without `@alias`).

```js
/**
 * @namespace
 */
var MyNamespace;

/**
 * @memberof MyNamespace
 */
class Class {}

/**
 * GOOD
 *
 * @function
 */
MyNamespace.Class.prototype.M1;

/**
 * GOOD
 *
 * @function
 */
Class.prototype.M2;

/**
 * GOOD
 *
 * @function
 * @memberof MyNamespace.Class.prototype
 * @alias M3
 */
Class.prototype.M3;

/**
 * GOOD
 *
 * @function
 * @memberof! MyNamespace.Class.prototype
 * @alias M4
 */
Class.prototype.M4;

/**
 * BAD: On MyNamespace page as "(static) Class#5"
 *
 * @function
 * @memberof MyNamespace
 */
Class.prototype.M5;

/**
 * BAD: On Class page as "(static) Class#M6"
 *
 * @function
 * @memberof MyNamespace.Class
 */
Class.prototype.M6;

/**
 * BAD: Missing entirely.
 *
 * @function
 * @memberof MyNamespace.Class.prototype
 */
Class.prototype.M7;

/**
 * BAD: On Class page as "Class#M8"
 *
 * @function
 * @memberof! MyNamespace.Class.prototype
 */
Class.prototype.M8;
```

43081j: thanks so much @aomarks  for investigating this. its all rather confusing and not-so-well-documented.

M2 is what we had originally but doesn't work for us because the analyser doesn't know what namespace it belongs to (it doesn't have any knowledge of the `DomApi` class its self at the point it detects these methods).

I did notice the `static` problem of M5 too. adding `@instance` to resolve it ended up with `Namespace#Class#Method` which is nonsense.

I got confused because m5 did end up correct almost (`namespace.class#method` but was marked as static). i guess i was missing `@alias` when i had tried it with the full name.

43081j:
given we now understand what on earth `@memberof` does/is expected to behave like, what do you think we should do here?

it would still be a nice thing to have for when we can't fully qualify the name.

justinfagnani:
@43081j I think we should implement`@memberof` semantics as properly as possible.

From @aomarks analysis, am I correct in that is seems like M7 is a pattern that jsdoc supports and we should?

43081j:
@justinfagnani  i think M3 is the right one but im still not entirely sure what `@alias` is doing in that case, and what we should do if its missing (because we then end up with M7 which is wrong as the member ends up `static` in jsdoc).

43081j:
i've changed this to pretty much do M3/M7.

it doesn't check for `@alias` but the absence of it will break jsdoc. for us, though, we don't need it to know what we want.

let me know if this needs changing though. the implementation of `_getNamespacedClassName` could do with some feedback.

cc @justinfagnani @aomarks

aomarks:
Closing this pull request because this repo is now archived. If this PR is still active, please submit a new one at https://github.com/Polymer/tools.

PR #833, opened 2018-01-08
Cleanup handling of legacy (Polymer 1.0 and hybrid syntax) element declarations

rictic:


rictic:
Recommend reviewing with ?w=1 in the url, as the diff is much nicer.

This change does two things:
  * removes some code for handling class declarations inside of Polymer() calls.
    * this code was untested
    * this code constructed an element object, but as far as I can tell there is no code that would cause us to recognize and emit that element object, because the element is only emitted when we exit the call expression. It also handles class declarations rather than the expected class expressions, and the class declaration would have to both exist inside of the Polymer() call.
    * in short, this code probably made sense at some point, but through progressive refactoring and a lack of tests for it, it is a very complicated noop. our main class handling code is still in class-scanner.
  * is more explicit about handling Polymer 1.0 style object literals
    * rather than matching on all object literals at any depth inside the Polymer() call, we look at the argument to Polymer() directly. this is more robust

src/test/polymer/polymer-element_test.ts
```diff
@@ -28,6 +28,7 @@ suite('PolymerElement', () => {
   async function getElements(filename: string): Promise<Set<PolymerElement>> {
     const result = (await analyzer.analyze([filename])).getDocument(filename);
     if (!result.successful) {
+      console.log(result.error);
```
aomarks: Did you mean to leave this `console.log`?

src/test/polymer/polymer-element_test.ts
```diff
@@ -156,6 +157,22 @@ suite('PolymerElement', () => {
     ]);
   });
 
+  test.only('Handles inner classes properly', async () => {
```
aomarks: Remove `test.only`.

aomarks:
I ran this through `gen-typescript-declarations` just to be sure, passes there too.

Issue #832, opened 2018-01-06
Declaring class inside function makes `polymer build` fail.


TimvdLippe:
_From @kkimdev on December 25, 2017 3:17_

### Description
On Polymer 3.0, declaring class inside function makes `polymer build` fail.

#### Steps to Reproduce

1. Follow https://www.polymer-project.org/blog/2017-08-23-hands-on-30-preview
2. edit myapp.js so that
```javascript
// ...
  constructor() {
    super();
    class Foo{};
    this.name = '3.0 preview';
  }
// ...
```

#### Expected Results
`polymer build` succeeds.

#### Actual Results
`polymer build` fails.
```
polymer build
info:    Clearing build/ directory...
info:    (default) Building...


    <script type="module" src="src/my-app.js"></script>
                              ~~~~~~~~~~~~~~~

index.html(9,30) error [could-not-load] - Unable to load import: Cannot read property 'properties' of null
error:   Promise rejection: Error: 1 error(s) occurred during build.
error:   Error: 1 error(s) occurred during build.
    at BuildAnalyzer._done (/usr/local/share/.config/yarn/global/node_modules/polymer-build/lib/analyzer.js:262:36)
    at BuildAnalyzer.<anonymous> (/usr/local/share/.config/yarn/global/node_modules/polymer-build/lib/analyzer.js:222:26)
    at Generator.next (<anonymous>)
    at fulfilled (/usr/local/share/.config/yarn/global/node_modules/polymer-build/lib/analyzer.js:17:58)
    at <anonymous>
```

### Versions
$ polymer --version
1.5.7

_Copied from original issue: Polymer/polymer#5005_

TimvdLippe:
_From @kkimdev on December 25, 2017 3:18_

@farQtech

rictic:
Standalone repro:
```html
<base href="https://polygit.org/polymer+:master/webcomponents+:master/shadycss+webcomponents+:master/components/">
<script src="webcomponentsjs/webcomponents-lite.js"></script>
<link rel="import" href="polymer/polymer.html">

<hello-world></hello-world>

<script>
  window.addEventListener('WebComponentsReady', () => {
    class HelloWorld extends Polymer.Element {
      constructor() {
        super();
        class Foo {};
      }
      static get is() {return 'hello-world';}
    };
    customElements.define(HelloWorld.is, HelloWorld);
  });
</script>
```

[...]

PR #835, opened 2018-01-08
Make error handling during parsing or scanning more robust.

rictic:
Pass along info on why we failed rather than producing `[Internal Error]`.

This would have helped us diagnose #832 more easily, which I think has been around for a while.

Fixes https://github.com/Polymer/polymer-analyzer/issues/834

### DIFF
--- a/src/core/analysis-context.ts
+++ b/src/core/analysis-context.ts
@@ -554,7 +554,7 @@ function makeRequestedWithoutLoadingWarning(resolvedUrl: ResolvedUrl) {
     },
     code: 'unable-to-analyze',
     message: `[Internal Error] Document was requested ` +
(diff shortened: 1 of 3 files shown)


src/core/analysis-context.ts
```diff
@@ -516,7 +554,11 @@ function makeRequestedWithoutLoadingWarning(resolvedUrl: ResolvedUrl) {
     },
     code: 'unable-to-analyze',
     message: `[Internal Error] Document was requested ` +
-        ` before loading and scanning finished. This should never happen.`,
+        `before loading and scanning finished. This usually indicates an` +
```
aomarks: missing a space at the end of this string

Issue #713, opened 2017-08-08
Event description is not used


veith:
It would be nice if the event description is used instead of the event name. The description of an event is very helpful / important. It is also recommended in the documentation (https://www.polymer-project.org/2.0/docs/tools/documentation#events).

*current output in analysis.json*
```
"events": [
  {
        "type": "CustomEvent",
        "name": "request",
        "description": "request",
        "metadata": {}
   },...
``` 

*current output in iron-doc-element*
``` 
request: CustomEvent
request
```

*desired output*
```
"events": [
  {
        "type": "CustomEvent",
        "name": "request",
        "description": "Fired when a response is received. ",
        "metadata": {}
   },...
``` 
``` 
request: CustomEvent
Fired when a response is received. 
```
 
*example (iron-ajax)*
https://github.com/PolymerElements/iron-ajax/blob/master/iron-ajax.html#L56
https://www.webcomponents.org/element/PolymerElements/iron-ajax/elements/iron-ajax#event-request

veith:
@FredKSchott This is the Issue, we talked about in Kopenhagen

davidrleonard:
# Analyzer event issue

We’re also facing this issue. We’re analyzing our components with polymer-analyzer and passing the resulting data to iron-doc-element for our documentation on predix-ui.com. But the event descriptions are not coming through.

Here’s an example event I want to parse inside one of my Polymer components:

```js
    /**
     * Fired when the search input field's content changes.
     * @event px-typeahead-search-input-change
     */
```

I think I found the problem:

After a file is with a Polymer component is parsed, its bits of jsdoc attached to AST nodes are peeled off and prepped for analysis. The events found in a Polymer component are parsed here in `getEventComments`:

https://github.com/Polymer/polymer-analyzer/blob/18d68db38f927c72f59ba7160a0a3a0f82838ddb/src/javascript/esutil.ts#L124-L145

On line 140, the jsdoc `comment` for an event is sent to `jsdoc.parseJsdoc`:

https://github.com/Polymer/polymer-analyzer/blob/18d68db38f927c72f59ba7160a0a3a0f82838ddb/src/javascript/jsdoc.ts#L27-L44

parseJsdoc uses the `doctrine.parse` function to parse stuff. What comes back from doctrine looks like this:

```json
{
  "description": "Event fired when search input field changes.\n",
  "tags": [
    {
      "title": "event",
      "description": "px-typeahead-search-input-change"
    }
  ]
}
```

[...]

PR #838, opened 2018-01-11
Detect event descriptions correctly

### LINKED ISSUE #404: Multiple @event annotations in one comment
https://github.com/PolymerElements/iron-ajax/blob/master/iron-ajax.html#L56

We should either recognize both or warn.

### LINKED ISSUE #836: Descriptions for @event annotations are not included in output.
The event info in the output does not include the comment text as the description for the event. For example, [these comments](https://github.com/PolymerElements/iron-selector/blob/6e843058d505f41fb088e6e988fa1dfe34aabd0a/iron-selectable.html#L20-L46) result in this output:
```json
/\/\/\/\/\/\/
          "events": [
            {
              "type": "CustomEvent",
              "name": "iron-activate",
              "description": "iron-activate",
              "metadata": {},
              "inheritedFrom": "Polymer.IronSelectableBehavior"
            },
            {
              "type": "CustomEvent",
              "name": "iron-deselect",
              "description": "iron-deselect",
              "metadata": {},
              "inheritedFrom": "Polymer.IronSelectableBehavior"
            },
            {
              "type": "CustomEvent",
              "name": "iron-items-changed",
              "description": "iron-items-changed",
              "metadata": {},
              "inheritedFrom": "Polymer.IronSelectableBehavior"
            },
            {
              "type": "CustomEvent",
              "name": "iron-select",
              "description": "iron-select",
              "metadata": {},
              "inheritedFrom": "Polymer.IronSelectableBehavior"
            },
/\/\/\/\/\/\/
```
(polymer-cli v1.5.7)

43081j:
* Use the annotation description before trying the tag description (Fixes #836)

### DIFF
--- a/src/test/javascript/esutil_test.ts
+++ b/src/test/javascript/esutil_test.ts
@@ -13,12 +13,61 @@
  */
 
 import * as babel from 'babel-types';
+import {parse} from 'babylon';
 import {assert} from 'chai';
 
-import {objectKeyToString} from '../../javascript/esutil';
+import {objectKeyToString, getEventComments} from '../../javascript/esutil';
 
 
 // See analysis_test for tests of generateElementMetadata
+
+suite('getEventComments', () => {
+  test('returns events from a comment', () => {
+    const node = parse(`
+        class Foo {
+          /**
+           * This is an event
+           *
+           * @event event-name
+           * @param {Event} event
+           */
+           myMethod() { }
+
+           /**
+            * @event descriptionless-event
+            */
+           anotherMethod() { }
+        }`);
+    const events = [...getEventComments(node).values()];
+    const eventMatches = events.map((ev) => ({
+      description: ev.description,
+      name: ev.name,
+      params: ev.params,
+      warnings: ev.warnings
+    }));
+
+    assert.deepEqual(eventMatches, [
+      {
+        description: 'descriptionless-event',
+        name: 'descriptionless-event',
+        params: [],
+        warnings: []
+      },
+      {
+        description: 'This is an event',
+        name: 'event-name',
+        params: [
+          {
+            desc: '',
+            name: 'event',
+            type: 'Event'
+          }
+        ],
+        warnings: []
+      }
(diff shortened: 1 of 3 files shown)


davidrleonard:
It appears this would also fix #713. It would be very helpful if this got reviewed and merged. Thanks, maintainers!

aomarks:
I'm not totally sure about supporting multiple events. I tried out this syntax with jsdoc, and only the last `@event` annotation is detected, so this would be a deviation. Do we have more than a few places we would need to support this style? If it's only a few, we could just update the source with a duplicate comment.

Big +1 to fixing the description and adding tests, though. If we agree on what I said above, maybe pull the other two bits into a separate PR?

43081j:
@aomarks  yup i wasn't too sure about jsdoc-validity either. multiple events makes sense logically but i haven't seen it used anywhere other than polymer.

we use it in several places in polymer and its elements from what i remember, but ill have another look tonight to see how often.

if we don't do it much then duplicating comments is a better idea.

regardless, ill split this PR up because it fixes two issues.

43081j:
this has been updated now to only include the description fix and tests @aomarks 

there is a new PR for multiple events at #840

src/test/javascript/esutil_test.ts
```diff
@@ -13,13 +13,62 @@
...
 
+suite('getEventComments', () => {
+  test('returns events from a comment', () => {
+    const node = parse(`
+        class Foo {
+          /**
+           * This is an event
+           *
+           * @event event-name
+           * @param {Event} event
```
aomarks: I guess this was already the case, but how can an event have a param? It's not a function. At http://usejsdoc.org/tags-event.html they use @property, which seems to make more sense to me.

43081j: i meant to ask you about this. its here because in various polymer elements we do this... but i read the jsdoc docs and it isn't mentioned anywhere. it should be a property or type, or something

PR #840, opened 2018-01-17
Scan multiple events in a single annotation

43081j:
This allows multiple events to be scanned from a single annotation.

Btw it does depend on #838 (branched off that).

## Do not merge yet

cc @aomarks 

we can discuss here whether we want this or not. you may be right about just doing PRs at iron-ajax and what not instead, to fix up the annotations.


43081j:
FYI i looked through all of `PolymerElements` elements and **only** iron-ajax does this.

im going to close this and fix it in polymerelements/iron-ajax#307.

PR #841, opened 2018-01-17
Make FSUrlLoader#canLoad honestly report that it can't load files outside its root

usergenic:
* `FSUrlLoader#canLoad` *now* reports _false_ for local urls outside the loader's own root; enables fall-thru support needed for racking in `MultiUrlLoader`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
+* `FSUrlLoader#canLoad` reports false for local urls outside the loader's
+  own root; enables fall-thru support needed for racking in `MultiUrlLoader`.
 * Add `Element#template` for getting the template of an element.
 * In MultiUrlLoader, proxy the first implementation of readDirectory, if any.
 * Use event annotation descriptions over their tag description
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 * `FSUrlLoader#canLoad` reports false for local urls outside the loader's
-  own root; enables fall-thru support needed for racking in `MultiUrlLoader`.
+  own root; enables fall-thru support needed for use with `MultiUrlLoader`.
 * Add `Element#template` for getting the template of an element.
 * In MultiUrlLoader, proxy the first implementation of readDirectory, if any.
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
+* `FSUrlLoader#canLoad` reports false for local urls outside the loader's
+  own root; enables fall-thru support needed for racking in `MultiUrlLoader`.
```
aomarks: racking?

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
+* `FSUrlLoader#canLoad` reports false for local urls outside the loader's
```
aomarks: Why can't you load a file outside your root? Maybe add more context to the method name? `canLoadFromSomething`? Maybe I don't know what "load" means exactly.

usergenic: Well, it was a pre-existing behavior that `FSUrlLoader` restricted loads to its root.  What I've changed here is that `canLoad` used to always return `true` if the url given was a local file, regardless of whether in or out of the root.  However, `load` would throw an Error if it was outside the `root`.  This was lame because it meant we couldn't use a `MultiUrlLoader` with more than one `FSUrlLoader`.  I use the term `racking` in the sense of a `MultiUrlLoader` looking like a rack or stack of processors with fall-throughs; MultiUrlLoader resolved urls in sequence top-to-bottom, returning first defined result:
```
 |  InMemoryLoader 1
 |  FSUrlLoader 1
 |  FSUrlLoader 2
 V  InMemoryLoader 2
```

Maybe the word "racking" is obscure.  I could go with sequencing...

CHANGELOG.md
usergenic: Yeah that wasn't clear.  I'll rename this to sequencing.

usergenic: Better yet I just rewrote "support need for use with `MultiUrlLoader`"

CHANGELOG.md
usergenic: ...the reason we couldn't use them with MultiUrlLoader is the MultiUrlLoader couldn't interrogate the FSUrlLoader about whether a file load was possible.

usergenic: ...and get a correct answer.

aomarks: Thanks for explaining!

PR #843, opened 2018-01-18
Redirect resolvers should resolve redirects.

usergenic:
- `RedirectResolver` resolves URLs which start with its redirect-to.

The problem is that redirect resolvers produce resolved urls which they later return `undefined` when asked to resolve again.  UrlResolvers should essentially be idempotent, in that once a URL is resolved, it should stay resolved.

This fix prevents a particularly awkward case where a `MultiUrlResolver` includes a `RedirectResolver` which produce a resolved URL which points to files outside the package root, which the subsequently racked `PackageUrlResolver` perceives to be a sibling directory and therefore reroutes to the component directory when the resolved url is subsequently given to `UrlResolver.resolve()`.

While it could be argued that resolved urls should never be asked to be resolved again, at this time we really can't guarantee this as there are many intermediate data-structures or serializations which occur that we can only assume are URLs of some form, relative, package-relative or resolved.

### DIFF
--- a/src/url-loader/redirect-resolver.ts
+++ b/src/url-loader/redirect-resolver.ts
@@ -37,12 +37,15 @@ export class RedirectResolver extends UrlResolver {
         this.getBaseAndUnresolved(firstUrl, secondUrl);
     const packageRelativeUrl =
         this.brandAsResolved(urlLibResolver(baseUrl, unresolvedUrl));
-    if (packageRelativeUrl === undefined ||
-        !packageRelativeUrl.startsWith(this._redirectFrom)) {
-      return undefined;
+    if (packageRelativeUrl.startsWith(this._redirectFrom)) {
+      return this.brandAsResolved(
+          this._redirectTo +
+          packageRelativeUrl.slice(this._redirectFrom.length));
     }
-    return this.brandAsResolved(
-        this._redirectTo + packageRelativeUrl.slice(this._redirectFrom.length));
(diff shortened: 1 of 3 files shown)


src/url-loader/redirect-resolver.ts
```diff
@@ -37,12 +37,15 @@ export class RedirectResolver extends UrlResolver {
         this.getBaseAndUnresolved(firstUrl, secondUrl);
     const packageRelativeUrl =
         this.brandAsResolved(urlLibResolver(baseUrl, unresolvedUrl));
-    if (packageRelativeUrl === undefined ||
-        !packageRelativeUrl.startsWith(this._redirectFrom)) {
-      return undefined;
+    if (packageRelativeUrl.startsWith(this._redirectFrom)) {
```
aomarks: are there any issues with missing/extra leading/trailing slashes here, by doing startsWith? (would have been an issue anyway, just curious)

usergenic: There's no extra-slash problem I can think of, but I think there's a potential that a redirect resolver that didn't enforce a trailing slash on the redirect-from could wind up resolving urls in a sibling directory unintentionally like `/folder` redirecting to `/folder3` could redirect `/folder2/page.html` to `/folder32/page.html`.  I think that class of problem should be looked at separately.  Added follow-up https://github.com/Polymer/polymer-analyzer/issues/844

rictic:
Endorsed that url resolving should be idempotent. We often need to re-resolve a url.

PR #850, opened 2018-01-23
add support for getters to behavior scanner

### LINKED ISSUE #849: Behaviours do not scan getters as properties
Seeing as behaviours are scanned by their own scanned, the getters PR to the class scanner didn't make it to the BehaviorScanner.

So a behaviour:

```js
{
  get foo() { return true; }
}
```

results in `foo` being a method rather than a property.

43081j:
Fixes #849.

Let me know if this is the correct solution.

I moved two methods to `esutil` as they are now shared between the class scanner and behavior scanner. Though I really don't want `esutil` to be the catch-all file, so if there's a better place, feel free to say so.


src/javascript/class-scanner.ts
```diff
@@ -838,10 +838,10 @@ export function extractPropertiesFromClass(
 
     for (const val of accessors.values()) {
```
rictic: Hm, there's some other logic here that it would be nice to also have in the behaviors code, like getting `readOnly` correct. What about a method that encapsulates everything with an API like:

```javascript
for (const prop of propertiesFromGettersAndSetters(astNode.body.body)) {
  properties.set(prop.name, prop);
}
```

43081j: hilariously @rictic it looks like this piece of code was fubar all along too. the accessors loop is _inside_ the body loop... we've been doing an iteration of the accessors per method all along /facepalm

43081j:
here you go @rictic. i added some comments where i question myself too

src/javascript/esutil.ts
```diff
@@ -199,6 +199,109 @@ export function getPropertyValue(
...
+
+  for (const [name, accessor] of accessors) {
+    let comment: string | undefined = undefined;
+    const member = (accessor.getter || accessor.setter)!;
+
+    if (accessor.getter) {
+      comment = getAttachedComment(accessor.getter);
+    }
+
+    if (accessor.setter && !comment) {
```
43081j: here we kind of want to say, "use what you can from the getter, otherwise get it from the setter". we're not quite doing that here, though. do you think its worth basically `||`ing all the annotation-based stuff? (e.g. `getReturnFromAnnotation(getter) || getReturnFromAnnotation(setter)`)

src/polymer/behavior-scanner.ts
```diff
@@ -85,10 +85,15 @@ class BehaviorVisitor implements Visitor {
...
     }
 
+    const methods: babel.Method[] = [];
+
     for (const prop of node.properties) {
       if (babel.isSpreadProperty(prop)) {
         continue;
       }
+      if (babel.isMethod(prop)) {
+        methods.push(prop);
```
43081j: i have to do this because an object property can be more than just a `Method`

src/javascript/esutil.ts
```diff
@@ -199,6 +199,109 @@ export function getPropertyValue(
...
+    if (!babel.isMethod(member)) {
+      continue;
+    }
+
+    // TODO: we dont support statics yet
+    if (babel.isClassMethod(member) && member.static) {
+      continue;
+    }
+
+    if (!babel.isIdentifier(member.key) || !member.key.name) {
```
rictic: The member name shouldn't be computed either. e.g. a method defined like this would match I think unless we check for `computed`, and we'd think that the method name was `dynamicName` rather than `foo`:

```javascript
const dynamicName = 'foo';
const obj = {
  [dynamicName]() {
    return 'bar';
  }
};
```

https://astexplorer.net/#/gist/3b8e43eae071c9aab4cbdf243b8a744e/fb585ceff75671e1ffd27e23bf65c4f969ff0acb

src/javascript/esutil.ts
rictic: Doing that kind of merging is hard. What if they have incompatible information? I think only taking the getter's info if there's comments on both is a simple policy that's easy to understand, so this is fine to me.

src/polymer/behavior-scanner.ts
```diff
@@ -105,15 +110,22 @@ class BehaviorVisitor implements Visitor {
       if (name in this.propertyHandlers) {
         this.propertyHandlers[name](prop.value);
       } else if (babel.isMethod(prop) || babel.isFunction(prop.value)) {
-        const method = esutil.toScannedMethod(
-            prop, this.document.sourceRangeForNode(prop)!, this.document);
-        this.currentBehavior.addMethod(method);
+        if (!babel.isMethod(prop) || (prop.kind !== 'get' && prop.kind !== 'set')) {
```
rictic: The `!babel.isMethod(prop)` part here doesn't seem right. We only want to filter out accessors here, don't we?

src/polymer/behavior-scanner.ts
```diff
@@ -105,15 +110,22 @@ class BehaviorVisitor implements Visitor {
...
+          this.currentBehavior.addMethod(method);
+        }
       } else {
         const property = toScannedPolymerProperty(
             prop, this.document.sourceRangeForNode(prop)!, this.document);
         this.currentBehavior.addProperty(property);
       }
     }
+
+    for (const prop of esutil.propertiesFromGettersAndSetters(methods, this.document)) {
```
rictic: Could avoid needing to create `methods` if we expanded the type of the first argument to `propertiesFromGettersAndSetters`. After all, we filter out anything that isn't a method in the body of `propertiesFromGettersAndSetters`.

43081j: i think i did this because `properties` is `ObjectProperty | ObjectMethod | SpreadProperty` and `SpreadProperty` isn't considered a `Method | Property`.

when i tried to `.filter(p => babel.isMethod(p))`, the types weren't picked up right so  iguess i went wrong somewhere.

src/polymer/behavior-scanner.ts
43081j: `!babel.isMethod` is because it could be a `Function`, and the second condition is for if it is a method, we want to check `Method#kind` isn't an accessor.

src/polymer/behavior-scanner.ts
43081j: i see you mean just add `SpreadProperty`, just seemed odd at first but i will as its much easier

src/polymer/behavior-scanner.ts
rictic: I'm saying we should either call `addMethod` for functions (like we are currently), or we should drop the `|| babel.isFunction` part on line 112.

I think functions on the behavior object should be treated as methods, even if they don't use method syntax.

rictic: 👀

43081j: 👋 
so from what you're saying, i _think_ you mean i should put another `else if (babel.isFunction(prop.value))` and treat them as methods?

this double-if makes me twitch too. the logic looks right but its ugly. the `!babel.isMethod` is because, if its not a method, it must be a function... and if it is a method, it has a `kind` so we check if its an accessor

43081j: its like...

```
if (method or function) {
  if (function or method_which_isnt_an_accessor) {
    do_a_method;
  } else {
    do_a_property;
  }
}
```

43081j:
FYI i think this will be replaced by #869 so we gain the benefits from the namespace scanner improvements too.

43081j:
closing, #869 replaces this.

PR #852, opened 2018-01-26
Updated CHANGELOG for 3.0.0-pre.9 release.

usergenic:
## [3.0.0-pre.9] - 2017-01-25
 * [BREAKING] Document#astNode and ParsedDocument#astNode are now an
   `AstNodeWithLanguage`, because we support inline documents in more than just
   HTML, as we've added an HTML-in-JS scanner.
 * `MultiUrlResolver` now delegates the `relative()` method to the first
   `UrlResolver` in its `_resolvers` array that can `resolve()` the
   destination URL.  Makes it possible now to rely on the Analyzer's
   resolver to return a valid `PackageRelativeUrl` from a resolved URL.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
+<!-- Add new, unreleased changes here. -->
+
+## [3.0.0-pre.9] - 2017-01-25
 * [BREAKING] Document#astNode and ParsedDocument#astNode are now an
   `AstNodeWithLanguage`, because we support inline documents in more than just
   HTML, as we've added an HTML-in-JS scanner.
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -6,14 +6,16 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
+<!-- Add new, unreleased changes here. -->
+
+## [3.0.0-pre.9] - 2017-01-25
```
aomarks: It's 2018 🎉 🎈 ! I do the same thing. Also maybe update the ones below :)

usergenic: OMG OMG OMG fix'd!

THANK YOU!

CHANGELOG.md
```diff
@@ -6,24 +6,26 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
```
rictic: Commenting out the Unreleased line gives better output in release PRs, as otherwise the rendered markdown kinda looks like the latest version is unreleased:

## Unreleased
## [3.0.0-pre.9] - 2018-01-25

usergenic: +1 doing it.  Also changing the date to 26th.

PR #853, opened 2018-01-27
RedirectResolver#relative() now correctly represents relative-to-package-root

usergenic:
- In order to get a package relative URL back out of a resolver from a ResolveUrl, we essentially call `relative(url)` and it will convert something like `file:///this/package/has/a/file.js` to `has/a/file.js`.
- Now this is true for the `RedirectResolver` which will reverse the effect of a redirect from the provided `url` so as to return a URL relative to the package URL.

### DIFF
--- a/src/test/url-loader/redirect-resolver_test.ts
+++ b/src/test/url-loader/redirect-resolver_test.ts
@@ -14,9 +14,8 @@
 
 import {assert} from 'chai';
 
-import {ResolvedUrl} from '../../model/url';
 import {RedirectResolver} from '../../url-loader/redirect-resolver';
-import {fileRelativeUrl, packageRelativeUrl, resolvedUrl} from '../test-utils';
+import {packageRelativeUrl, resolvedUrl} from '../test-utils';
 
 
 suite('RedirectResolver', function() {
@@ -56,7 +55,7 @@ suite('RedirectResolver', function() {
       const resolver =
           new RedirectResolver(resolvedUrl`/a/`, 'proto://site/', '/b/');
       const relative = resolver.relative(resolvedUrl`/b/page.html`)!;
-      assert.equal(relative, fileRelativeUrl`proto://site/page.html`);
+      assert.equal(relative, packageRelativeUrl`proto://site/page.html`);
       assert.equal(
           resolver.relative(resolvedUrl`proto://site/page.html`) as string,
           resolvedUrl`proto://site/page.html`);
--- a/src/test/url-loader/redirect-resolver_test.ts
+++ b/src/test/url-loader/redirect-resolver_test.ts
@@ -51,7 +51,7 @@ suite('RedirectResolver', function() {
   });
 
   suite('relative', () => {
-    test('if `to` is no in redirect-to, return as-is', () => {
+    test('if `to` is not in redirect-to, return as-is', () => {
       const resolver = new RedirectResolver(
           resolvedUrl`file:///src/a/`,
           resolvedUrl`proto://site/`,
--- a/src/test/url-loader/redirect-resolver_test.ts
+++ b/src/test/url-loader/redirect-resolver_test.ts
@@ -52,14 +52,14 @@ suite('RedirectResolver', function() {
   });
 
   suite('relative', () => {
-    test('if from not within _redirectTo but to is, reverse redirect', () => {
+    test('if `from` is not in redirect-to, un-redirect the `to`', () => {
       const resolver =
           new RedirectResolver(resolvedUrl`/a/`, 'proto://site/', '/b/');
       const relative = resolver.relative(resolvedUrl`/b/page.html`)!;
       assert.equal(relative, fileRelativeUrl`proto://site/page.html`);
       assert.equal(
-          resolver.relative(relative as string as ResolvedUrl) as string,
-          relative);
+          resolver.relative(resolvedUrl`proto://site/page.html`) as string,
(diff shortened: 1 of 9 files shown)


src/test/url-loader/redirect-resolver_test.ts
```diff
@@ -49,4 +49,30 @@ suite('RedirectResolver', function() {
       assert.equal(resolver.resolve(resolved), resolved);
     });
   });
+
+  suite('relative', () => {
+    test('if `to` is no in redirect-to, return as-is', () => {
```
rictic: `is no` -> `is not`

src/test/url-loader/url-resolver_test.ts
```diff
@@ -25,8 +25,10 @@ class SimplestUrlResolver extends UrlResolver {
     return this.simpleUrlResolve(baseUrl, url);
   }
 
+  relative(to: ResolvedUrl): PackageRelativeUrl;
+  relative(from: ResolvedUrl, to: ResolvedUrl, kind?: string): FileRelativeUrl;
```
rictic: I think for implementations you can leave off the two type lines

src/url-loader/multi-url-resolver.ts
```diff
@@ -47,11 +47,20 @@ export class MultiUrlResolver extends UrlResolver {
    * Delegates to relative method on the first resolver which can resolve the
    * destination URL.
    */
+  relative(to: ResolvedUrl): PackageRelativeUrl;
+  relative(from: ResolvedUrl, to: ResolvedUrl, kind?: string): FileRelativeUrl;
   relative(fromOrTo: ResolvedUrl, maybeTo?: ResolvedUrl, kind?: string):
-      FileRelativeUrl {
+      FileRelativeUrl|PackageRelativeUrl {
+    const from = (maybeTo === undefined) ? undefined : fromOrTo;
```
rictic: What do you think of:

```typescript
const [from, to] = 
    (maybeTo === undefined) ? 
        [undefined, fromOrTo] : 
        [fromOrTo, maybeTo];
```

src/test/url-loader/url-resolver_test.ts
usergenic: For some reason it doesn't like it.

```
src/test/url-loader/url-resolver_test.ts(20,7): error TS2415: Class 'SimplestUrlResolver' incorrectly extends base class 'UrlResolver'.
  Types of property 'relative' are incompatible.
    Type '(fromOrTo: ResolvedUrl, maybeTo?: ResolvedUrl | undefined, _kind?: string | undefined) =>FileRel...' is not assignable to type '{ (to: ResolvedUrl): PackageRelativeUrl; (from: ResolvedUrl,to: ResolvedUrl, kind?: string | und...'.
      Type 'FileRelativeUrl | PackageRelativeUrl' is not assignable to type 'PackageRelativeUrl'.
        Type 'FileRelativeUrl' is not assignable to type 'PackageRelativeUrl'.
          Type 'FileRelativeUrl' is not assignable to type 'PackageRelativeUrlBrand'.
            Property 'PackageRelativeUrlBrand' is missing in type 'string & FileRelativeUrlBrand'.

20 class SimplestUrlResolver extends UrlResolver {
         ~~~~~~~~~~~~~~~~~~~
```

src/url-loader/multi-url-resolver.ts
usergenic: +1 done.

src/test/url-loader/redirect-resolver_test.ts
usergenic: Done.

usergenic:
@rictic PTAL

src/test/url-loader/url-resolver_test.ts
rictic: Hm, but `resolve` is written this way. Oh, maybe it's because they vary based on return type? Ok, yeah, I guess this is the simplest way to write this.

src/url-loader/multi-url-resolver.ts
```diff
@@ -47,11 +47,20 @@ export class MultiUrlResolver extends UrlResolver {
...
+    const to = (maybeTo === undefined) ? fromOrTo : maybeTo;
     for (const resolver of this._resolvers) {
-      if (resolver.resolve((maybeTo || fromOrTo) as any)) {
-        return resolver.relative(fromOrTo, maybeTo, kind);
+      if (!resolver.resolve(this.brandAsPackageRelative(to))) {
```
rictic: Should for undefined explicitly, rather than coercing string to boolean.

I mean, it should probably never resolve to an empty string, but it's easier for the reader and the VM to be explicit here I think.

rictic:
Arg my second comment is collapsed by default? GitHub… y u do dis to me.

Pasting comment in here so that it's seen:

> Should check for undefined explicitly, rather than coercing string to boolean.

> I mean, it should probably never resolve to an empty string, but it's easier for the reader and the VM to be explicit here I think.

src/url-loader/multi-url-resolver.ts
usergenic: Ok.  Will check for undefined before merging.

usergenic:
Meh Typescript 2.7.1 throwing an actual error but only with Node 6 for this branch + master.  Researching...

usergenic:
Rebased from master and force pushed.

usergenic:
And magically the problem goes away.  :-/  Squash+merge...

PR #855, opened 2018-01-30
trim whitespace when parsing jsdoc

### LINKED ISSUE #854: JSDoc with windows line endings don't get parsed correctly
With a jsdoc like this:

```ts
/**
 * foo
 * @customElement
 */
```

Assuming windows line endings (CRLF), this results in a `description` of `foo * @customElement` or some such thing.

The reason is that in [removeLeadingAsterisks](https://github.com/Polymer/polymer-analyzer/blob/49592f848c0d5972b5d0836dcdab332a748a82be/src/javascript/jsdoc.ts#L79).

We split on `\n`, it could be changed to `split(/\r?\n/)` or something. Or we find out why the regex in there doesn't match (it doesn't strip the leading asterisks...).

43081j:
Fixes #854.

There's some funkiness going on with the pattern here so I just trim it to drop any leftover newlines or CRs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 <!-- ## Unreleased -->
+* Support Windows line endings in JSDoc annotations
 <!-- Add new, unreleased changes here. -->
 
 ## [3.0.0-pre.9] - 2018-01-26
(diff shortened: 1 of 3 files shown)


rictic:
I like it, it needs a test. Simplest way would be to parse from a string with an InMemoryOverlayUrlLoader so that you can write the file under test as a string literal with `\r\n`, to be sure that git doesn't mess with the line endings.

43081j:
Would it not be simpler (and doesn't it make more sense) to add the test to the existing `parseJsdoc` suite?

we already have `jsdoc_test.ts`.

rictic:
Oh, sure yes!

43081j:
Here you go :)

PR #858, opened 2018-01-31
parse binary expressions of string literals

### LINKED ISSUE #656: Warning when analyzing observers, if line is splitted due to being too long
Just upgraded `polymer-build` to 1.2.0 which pulled in analyzer `2.0.0-alpha.40`, and there is a following warning (severity "info") when analyzing iron-ajax:
```
'_requestOptionsChanged(url, method, params.*, headers, contentType, ' +
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  'body, sync, handleAs, jsonPrefix, withCredentials, timeout, auto)'
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
vendor/iron-ajax/iron-ajax.html(329,6) info [unanalyzable-polymer-expression] - 
Can only analyze databinding expressions in string literals.
```
I think this can be a common case due to code style limitations. Could analyzer recognize this?

43081j:
This fixes #656.

Is this approach fine?

### DIFF
--- a/src/javascript/ast-value.ts
+++ b/src/javascript/ast-value.ts
@@ -28,8 +28,28 @@ function literalToValue(literal: babel.Literal): LiteralValue {
   if (babel.isNullLiteral(literal)) {
     return null;
   }
+  if (babel.isTemplateLiteral(literal)) {
+    return templateLiteralToValue(literal);
+  }
   // Any other literal value is treated as undefined.
   return undefined;
+}
+
+function templateLiteralToValue(literal: babel.TemplateLiteral): LiteralValue {
+  const len = literal.quasis.length - 1;
+  let value: LiteralValue = '';
+
+  for (let i = 0; i < len; i++) {
+    value += literal.quasis[i].value.raw;
+    const v = expressionToValue(literal.expressions[i]);
+    if (v !== undefined) {
+      value += `${v}`;
+    }
+  }
+
+  value += literal.quasis[len].value.raw;
+
+  return value;
 }
 
 /**
@@ -230,15 +250,3 @@ export function getNamespacedIdentifier(
 }
 
 export const CANT_CONVERT = 'UNKNOWN';
-
-export function isLiteralExpression(expr: babel.Node): boolean {
-  if (babel.isLiteral(expr)) {
-    return true;
-  }
-
-  if (babel.isBinaryExpression(expr)) {
-    return isLiteralExpression(expr.left) && isLiteralExpression(expr.right);
-  }
-
-  return false;
-}
--- a/src/javascript/ast-value.ts
+++ b/src/javascript/ast-value.ts
@@ -231,18 +231,14 @@ export function getNamespacedIdentifier(
 
 export const CANT_CONVERT = 'UNKNOWN';
 
-export function isLiteralBinaryExpression(expr: babel.Node): boolean {
-  if (!babel.isBinaryExpression(expr)) {
-    return false;
+export function isLiteralExpression(expr: babel.Node): boolean {
+  if (babel.isLiteral(expr)) {
+    return true;
   }
 
-  if (!babel.isLiteral(expr.left) && !isLiteralBinaryExpression(expr.left)) {
-    return false;
+  if (babel.isBinaryExpression(expr)) {
+    return isLiteralExpression(expr.left) && isLiteralExpression(expr.right);
   }
 
-  if (!babel.isLiteral(expr.right) && !isLiteralBinaryExpression(expr.right)) {
-    return false;
-  }
-
-  return true;
+  return false;
 }
--- a/src/javascript/ast-value.ts
+++ b/src/javascript/ast-value.ts
@@ -230,39 +230,3 @@ export function getNamespacedIdentifier(
 }
 
 export const CANT_CONVERT = 'UNKNOWN';
-
-/**
- * Converts a binary expression into a string
- *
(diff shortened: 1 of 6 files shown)


src/polymer/expression-scanner.ts
```diff
@@ -492,8 +492,9 @@ export function parseExpressionInJsStringLiteral(
     warnings
   };
   const sourceRangeForLiteral = document.sourceRangeForNode(stringLiteral)!;
+  const binaryExpressionString = astValue.binaryExpressionToString(stringLiteral);
 
-  if (!babel.isLiteral(stringLiteral)) {
+  if (binaryExpressionString === null && !babel.isLiteral(stringLiteral)) {
```
43081j: we are explicit here instead of just a falsy check because i suppose it could be an empty string

43081j:
turns out my first attempt overcomplicated it.

we already have all the logic for turning a binary expression into a string. so i've thrown the function i added away and just kept the conditional change.

src/javascript/ast-value.ts
```diff
@@ -230,3 +230,19 @@ export function getNamespacedIdentifier(
 }
 
 export const CANT_CONVERT = 'UNKNOWN';
+
+export function isLiteralBinaryExpression(expr: babel.Node): boolean {
```
merlinnot: You can make it even more simple:
```ts
export function isCoercibleToString(expr: babel.Node): boolean {
  if (babel.isLiteral(expr)) {
    return true;
  }

  return babel.isBinaryExpression(expr)) && [expr.left, expr.right].every(isCoercibleToString);
```

WDYT?

src/javascript/ast-value.ts
```diff
@@ -230,3 +230,15 @@ export function getNamespacedIdentifier(
...
+export function isLiteralExpression(expr: babel.Node): boolean {
+  if (babel.isLiteral(expr)) {
+    return true;
+  }
+
+  if (babel.isBinaryExpression(expr)) {
+    return isLiteralExpression(expr.left) && isLiteralExpression(expr.right);
+  }
+
+  return false;
```
43081j: @merlinnot how about this? just a little more readable and figured it may as well become a "literal expression" check (simplifies the calling point too)

43081j: in actual fact, i don't think we need this after all (@rictic?).

if we had an observer for some crazy insane reason, `'onProp' + [1] + 'Changed'`, its technically a literal in the end. maybe we should just care less and accept any `BinaryExpression` which begins with a literal?

merlinnot: Hm... I’d love to sneak in template literals too, so one can do (I tend to do that actually):
```ts
static get observers() {
  return [
    `f(
      sooooo,
      many,
      props,
      here
    )`,
  ];
}
```

src/polymer/expression-scanner.ts
```diff
@@ -493,7 +493,7 @@ export function parseExpressionInJsStringLiteral(
   };
   const sourceRangeForLiteral = document.sourceRangeForNode(stringLiteral)!;
 
-  if (!babel.isLiteral(stringLiteral)) {
+  if (!astValue.isLiteralExpression(stringLiteral)) {
```
rictic: What if, instead of this check, we wrote it more like:

```typescript
const expressionText = astValue.expressionToValue(stringLiteral);
if (expressionText === undefined) {
  // emit the unanalyzable expression warning
  return result;
}
if (typeof expressionText !== 'string') {	   
  // emit invalid expression warning
  return result;
}
```

Because `astValue.expressionToValue` already has a good idea of what it can and can't handle.

src/javascript/ast-value.ts
rictic: +1 to adding untagged template literals. supporting them in expressionToValue should be pretty straightforward

src/polymer/expression-scanner.ts
43081j: yup i was thinking the same. i dont think i need the new function i added. because if we pass a binary expression we don't understand, we already get `undefined` back

43081j:
here you go @rictic 
we have template literals supported now and i removed my unnecessary binary literal function

src/javascript/ast-value.ts
```diff
@@ -28,10 +28,30 @@ function literalToValue(literal: babel.Literal): LiteralValue {
...
 }
 
+function templateLiteralToValue(literal: babel.TemplateLiteral): LiteralValue {
+  const len = literal.quasis.length - 1;
+  let value: LiteralValue = '';
+
+  for (let i = 0; i < len; i++) {
+    value += literal.quasis[i].value.raw;
+    const v = expressionToValue(literal.expressions[i]);
+    if (v !== undefined) {
```
rictic: if `v === undefined` then this function must return `undefined`, as we can't give a confident evaluation of the template literal

... should really rewrite this family of functions to be a Result<LiteralValue>

rictic:
LGTM! Just needs a rebase

43081j:
sorry @rictic  this somehow slipped past me.

i rebased, it seems to look fine

rictic:
Nice!

PR #859, opened 2018-02-01
Prepare CHANGELOG for 3.0.0-pre.10 release.

usergenic:
## [3.0.0-pre.10] - 2018-02-01
 * [BREAKING] The `UrlResolver#relative()` method now returns a
   `PackageRelativeUrl` when called with only one argument, since the only
   use of this call form is to reverse a resolved URL back into a simpler
   package relative form.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,12 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
-## Unreleased
+<!-- ## Unreleased -->
+<!-- Add new, unreleased changes here. -->
+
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -5,12 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
-## Unreleased
+<!-- ## Unreleased -->
+<!-- Add new, unreleased changes here. -->
+
+## [3.0.0-pre.10] - 2018-02-01
 * [BREAKING] The `UrlResolver#relative()` method now returns a
```
rictic: In the past we haven't considered TypeScript type changes breaking, provided that runtime behavior is unchanged.

usergenic: Okay.  Updating.

PR #864, opened 2018-02-05
Resolve aliased super-classes

### LINKED ISSUE #863: Module Import Aliases Not Recognized
The analyzer seems to assume that the export name === the import name. So aliases created by imports are not followed.

Example:

```js
import {Element as PolymerElement} from '@polymer/polymer-element.js';
class Foo extends PolymerElement {}
```

43081j:
Fixes #863.

Just tries to find import aliases when resolving the superclass of a given class, instead of assuming the scanned reference is correct.

```ts
import { foo as bar } from './foo';
class Test extends bar {}
```

Also added a test to make sure these work in the JS import scanner, though i suppose it wasn't needed (really i slipped up and thought that's where this needed fixing but was wrong 🤐 )

cc @chadkillingsworth

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 <!-- ## Unreleased -->
+* Support super-classes referenced from aliased imports
 <!-- Add new, unreleased changes here. -->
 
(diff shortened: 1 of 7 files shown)


rictic:
I'm working on a unified scope-based resolution system that'll follow imports/reassignment/reexports etc.

Ideally we won't add symbols to the global class name based lookup system if they're declared in a module and they don't explicitly write to `window.Foo` or `global.Foo`

rictic:
Closing this PR as we're working on a different design right now.

PR #867, opened 2018-02-10
Be more principled in computing names.

rictic:
`objectKeyToString` looks like it was originally intended for getting propery names, but then a bunch of other code started using it in ways that they probably shouldn't, and it got extended to handle namespaced identifiers. So we ended up conflating getting identifier names with getting property keys, leading to all sorts of strangeness.

This PR switches the cases where we want an identifier name to call `getIdentifierName`, and redesigns `objectKeyToString` as `getPropertyName`. `getPropertyName` takes a more specific type, and handles computed values correctly.

One motivating example of where we get things wrong today: https://polymer.github.io/analyzer-playground/?input=%3Cbase+href%3D%22https%3A%2F%2Fpolygit.org%2Fpolymer%2B%3Amaster%2Fwebcomponents%2B%3Amaster%2Fshadycss%2Bwebcomponents%2B%3Amaster%2Fcomponents%2F%22%3E%0A%3Cscript+src%3D%22webcomponentsjs%2Fwebcomponents-lite.js%22%3E%3C%2Fscript%3E%0A%3Clink+rel%3D%22import%22+href%3D%22polymer%2Fpolymer.html%22%3E%0A%0A%3Cdom-module+id%3D%22hello-world%22%3E%0A++%3Ctemplate%3E%0A++++%3Cstyle%3E%0A++++++%3Ahost+%7B%0A++++++++display%3A+block%3B%0A++++++%7D%0A++++%3C%2Fstyle%3E%0A++++%3C%21--+Ideally+this+wouldn%27t+warn%2C+but+it%27s+ok+that+we+do+for+now.+--%3E%0A++++%3Cdiv%3EHello+%7B%7Bplace%7D%7D%21%3C%2Fdiv%3E%0A%0A++++%3C%21--+This+should+definitely+warn.+--%3E%0A++++%3Cdiv%3EHello+%7B%7BpropName%7D%7D%21%3C%2Fdiv%3E%0A++%3C%2Ftemplate%3E%0A++%3Cscript%3E%0A++++const+propName+%3D+%27place%27%3B%0A++++Polymer%28%7B%0A++++++is%3A+%27hello-world%27%2C%0A++++++properties%3A+%7B%0A++++++++%5BpropName%5D%3A+String%0A++++++%7D%0A++++%7D%29%3B%0A++%3C%2Fscript%3E%0A%3C%2Fdom-module%3E%0A

### DIFF
--- a/src/test/javascript/esutil_test.ts
+++ b/src/test/javascript/esutil_test.ts
@@ -13,13 +13,12 @@
  */
 
 import * as babel from 'babel-types';
+import * as babylon from 'babylon';
 import {parse} from 'babylon';
 import {assert} from 'chai';
 
-import {objectKeyToString, getEventComments} from '../../javascript/esutil';
+import {getEventComments, getPropertyName} from '../../javascript/esutil';
 
-
-// See analysis_test for tests of generateElementMetadata
 
 suite('getEventComments', () => {
   test('returns events from a comment', () => {
@@ -40,11 +39,11 @@ suite('getEventComments', () => {
         }`);
     const events = [...getEventComments(node).values()];
     const eventMatches = events.map((ev) => ({
-      description: ev.description,
-      name: ev.name,
-      params: ev.params,
-      warnings: ev.warnings
-    }));
+                                      description: ev.description,
+                                      name: ev.name,
+                                      params: ev.params,
+                                      warnings: ev.warnings
+                                    }));
 
     assert.deepEqual(eventMatches, [
       {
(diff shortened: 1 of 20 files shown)


src/test/javascript/esutil_test.ts
```diff
@@ -56,99 +55,35 @@ suite('getEventComments', () => {
...
+      'foo': 1,
+      bar: 2,
+      [10]: 3,
+      [10 + 20]: 4,
+      ['hi' + ' there']: 5,
+      [identifier]: 6,
+    });
+    `;
+    const statement = babylon.parse(objectLiteralCode).program.body[0];
+    if (!babel.isExpressionStatement(statement)) {
```
aomarks: why not use the assert lib here?

rictic: TypeScript can't reason about type narrowing with Assert

PR #868, opened 2018-02-10
Namespace properties support

### LINKED ISSUE #866: getters not recognized in namespaces
In this example:

```js
    /**
     * ...
     * @namespace
     * @memberof Polymer
     */
    Polymer.IronScrollManager = {
      /**
       * The current element that defines the DOM boundaries of the
       * scroll lock. This is always the most recently locking element.
       *
       * @memberof Polymer.IronScrollManager
       * @return {!Node|undefined}
       */
      get currentLockingElement() {
        return this._lockingElements[this._lockingElements.length - 1];
      },

      ...
```

(from https://github.com/PolymerElements/iron-overlay-behavior/blob/master/iron-scroll-manager.html)

The `currentLockingElement` getter is scanned as a function, instead of a readonly property.

43081j:
Fixes #866.

Also resolves some issues in PolymerElements/iron-overlay-behavior#262 and Polymer/gen-typescript-declarations#69.

Refactors some pieces of code #850 also touches. Depending on which is merged first, i'll have to update the other.

---

Introduces property scanning to namespaces. This includes getters/setters, too.

A namespace now has a `properties` map which is populate at the time the namespace is scanned.

We detect both object expression properties and `memberof` properties defined outside the initial namespace definition.

We ignore functions/methods because the function scanner picks those up.

cc @aomarks


src/javascript/namespace-scanner.ts
```diff
@@ -29,16 +30,133 @@ export class NamespaceScanner implements JavaScriptScanner {
       document: JavaScriptDocument,
       visit: (visitor: Visitor) => Promise<void>) {
     const visitor = new NamespaceVisitor(document);
```
rictic: Now that there are two visitors, let's call this one `namespaceVisitor` or `namespaceFinder`

src/javascript/namespace-scanner.ts
```diff
@@ -29,16 +30,133 @@ export class NamespaceScanner implements JavaScriptScanner {
...
+
     return {
-      features: Array.from(visitor.namespaces),
+      features: [...visitor.namespaces.values()],
       warnings: visitor.warnings
     };
   }
 }
 
+class NamespacePropertyFinder implements Visitor {
```
rictic: Why isn't this logic just in NamespaceVisitor? It seems like a lot of extra work to find namespaces twice, the first time just to know that the namespace exists, the second time to find their properties, and then reconcile. Why not apply this logic whenever NamespaceVisitor finds a namespace object?

src/javascript/esutil.ts
```diff
@@ -428,3 +428,153 @@ function* _getMethods(node: babel.Node) {
...
+    jsdoc: jsdocAnn,
+    sourceRange: document.sourceRangeForNode(method)!,
+    description,
+    privacy,
+    warnings: [],
+    readOnly,
+  };
+}
+
+export function extractPropertiesFromNode(
```
rictic: How about `extractPropertiesFromAccessorsAndObjectLiteralBodies`?

I know it's an absurdly long name, but we have _so many_ ways of determining properties, it's a challenge to tell when reading a method what any given call is trying to do.

src/javascript/esutil.ts
```diff
@@ -428,3 +428,153 @@ function* _getMethods(node: babel.Node) {
     }
   }
 }
+
+export function extractPropertyFromGetterOrSetter(
+    method: babel.ClassMethod|babel.ObjectMethod,
+    jsdocAnn: jsdoc.Annotation|undefined,
+    document: JavaScriptDocument): ScannedProperty|null {
+  if (babel.isClassMethod(method) && method.static) {
```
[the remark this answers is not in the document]
43081j: yup that's right. will add a comment

src/javascript/esutil.ts
43081j: > I know it's an absurdly long name

You don't say 😂 

i've changed it to `extractPropertiesFromClassOrObjectBody` (is that fine?)

43081j:
@rictic  i have moved the new visitor into the namespace visitor now.

iirc i originally extracted it into its own visitor to avoid ordering issues. but i now realise namespaces should always come before their properties anyway. so we can safely assume if the namespace isn't already scanned by the time we see a property, it never will be.

src/javascript/esutil.ts
rictic: +1

src/test/static/namespaces/namespace-properties.js
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * @memberof PropertiesNamespace
+ */
+PropertiesNamespace.notAPropertyDefinedLater = function() {};
+
+/**
+ * @memberof PropertiesNamespace.prototype
+ * @type {string}
+ */
+PropertiesNamespace.prototype.instanceProperty;
```
rictic: This isn't recognized yet, yeah?

PR #869, opened 2018-02-11
Add support for behaviour accessors (v2)

43081j:
This is an alternative approach for #850, this time based on the improvements introduced in #868.

@rictic do you think i should close #850 and adopt this one instead? i've basically extracted the properties iteration into its own function so the class scanner and behavior scanner can share it.

its a little tricky to see what i've changed because its based on my other PR branch. so maybe we should get that merged before we try code review


rictic:
Rebased into #872 and merged!

PR #871, opened 2018-02-12
Fix issue with imports in one module script tag leaking to other module script tag inline document

usergenic:
- `Document#getFeatures()` now supports an `excludeBackreferences`
  option for use with inline documents to exclude the container/importer
  document and its features from the results.
- [BREAKING] Removed the `ScriptTagBackReferenceImport` and replaced it
  with new general-purpose `DocumentBackreference` feature used to
  link inline/imported documents back to their containers/importers.


src/html/html-script-tag.ts
```diff
@@ -27,7 +27,10 @@ export class ScriptTagImport extends Import { readonly type = 'html-script'; }
  * A synthetic import that provides the document containing the script tag to
  * the javascript document defined/referenced by the script tag.
  */
+// TODO(usergenic): Reconsider whether this should behave like the
```
justinfagnani: Let's file issues for TODOs, and reference the issue.

src/model/inline-document.ts
```diff
@@ -45,6 +45,15 @@ export type AstNodeWithLanguage = {
   containingDocument: JavaScriptDocument,
 };
 
+export class ContainingDocumentBackreference extends ScannedFeature {
```
justinfagnani: doc comment

src/model/inline-document.ts
```diff
@@ -45,6 +45,15 @@ export type AstNodeWithLanguage = {
...
+  kinds = new Set(['containing-document-backreference']);
+  identifiers: Set<string>;
+  constructor(public readonly document: Document) {
+    super(undefined, undefined, undefined, undefined, undefined);
+    this.identifiers = new Set([...document.identifiers]);
```
justinfagnani: It's not clear to me why this should have the same identifiers as the container document. Can you clarify that, and add a comment?

usergenic: You know, I went and dropped the identifiers altogether, essentially because there should only ever be one of these backreference items in a given inline document so identifiers are extraneous.  I think I defaulted to cloning them at the time because I didn't know what else they should be; upon reflection, I decided they shouldn't be anything.

src/model/inline-document.ts
usergenic: Added.

src/html/html-script-tag.ts
usergenic: Done.  Also removed the TODO about reconsidering behavior.  I considered it and decided it makes sense as an Import.

usergenic:
@justinfagnani feedback addressed; PTAL

usergenic:
Thanks @justinfagnani

usergenic:
@justinfagnani PTAL: per our discussion in-person, I've removed the `ScriptTagBackReferenceImport` class altogether, renamed `ContainingDocumentBackreference` to `DocumentBackreference` and am using this class for connecting inline documents back to containing documents *as well as* connecting imported documents back to their importers.

usergenic:
@justinfagnani I also added assertion to existing test demonstrate the difference between `excludeBackreferences: true` and not.

PR #880, opened 2018-02-14
Updated CHANGELOG and package-lock.json for 3.0.0-pre.11 release.

usergenic:
## [3.0.0-pre.11] - 2018-02-14 💘
 * Support Windows line endings in JSDoc annotations
 * Support scanning of properties in namespaces
 * Support scanning of accessors in behaviors
 * `Document#getFeatures()` now supports an `excludeBackreferences`
   option for use with inline documents to exclude the container/importer
   document and its features from the results.
 * Support @return annotations on getters/setters on legacy Polymer function
   calls, and fix bug where readonly was inverted.
 * Add `IndirectUrlResolver`, a URL resolver for very large codebases where
   source files map in an arbitrary but predetermined fashion onto URL space.
   This resolver is still experimental.
   * It is generally true that if something has to be imported from
     `polymer-analyzer/lib/...` then it's not yet stable. caveat importer.
 * [BREAKING] Removed the `ScriptTagBackReferenceImport` and replaced it
   with new general-purpose `DocumentBackreference` feature used to
   link inline/imported documents back to their containers/importers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
-## Unreleased
+<!-- ## Unreleased -->
+<!-- Add new, unreleased changes here. -->
+
+## [3.0.0-pre.11] - 2018-02-14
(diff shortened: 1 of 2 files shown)


usergenic:
polymer-analyzer 3.0.0-pre.11 published 🎉

PR #883, opened 2018-02-15
Infer when functions return void.

### LINKED ISSUE #753: Feature request: infer void return type

aomarks:
Functions and methods will now be automatically inferred as returning `void` in certain cases. This occurs when all of the following are true:

1) It has no `@return` or `@returns` JSDoc annotation.
2) It is not  async or a generator.
3) Its body contains no return statements with  arguments.

Fixes #753
Related to #826 

Also contains some minor refactoring to the function scanner.


src/javascript/esutil.ts
```diff
@@ -284,6 +284,33 @@ export function getReturnFromAnnotation(jsdocAnn: jsdoc.Annotation):
...
 }
 
+/**
+ * Examine the body of a function to see if we can infer something about its
+ * return type. This currently only handles the case where a function definitely
+ * returns void.
+ */
+export function inferReturnFromBody(node: babel.Function): {type: string}|
+    undefined {
+  if (node.async === true || node.generator === true) {
```
rictic: need to early exit if node is an ArrowFunctionExpression whose body isn't a BlockStatement:

https://astexplorer.net/#/gist/5b1da2824ee4b9b919fb3f7ae4e1e628/2778b732d6cd4ceaf2f4d3ae141a9814c3d8e2f5

src/javascript/esutil.ts
```diff
@@ -284,6 +284,33 @@ export function getReturnFromAnnotation(jsdocAnn: jsdoc.Annotation):
...
+    return undefined;
+  }
+  let returnsVoid = true;
+  estraverse.traverse(node, {
+    enterReturnStatement(statement: babel.ReturnStatement) {
```
rictic: Can have fewer false negatives if you do something like:

```typescript
enterFunction() {
  return estraverse.VisitorOption.Skip;
}
```

src/javascript/function-scanner.ts
```diff
@@ -118,12 +115,14 @@ class FunctionVisitor implements Visitor {
...
   private _initFunction(
-      node: babel.Node, analyzedName?: string, _fn?: babel.Function) {
-    const comment = getAttachedComment(node);
-    if (!comment) {
-      return;
-    }
-    const docs = jsdoc.parseJsdoc(comment);
+      node: babel.Function|babel.ObjectProperty,
+      analyzedName?: string,
+      docNode?: babel.Node,
```
rictic: hm, not sure if I like the way clang-format formats a trailing comma in a parameter list

src/test/static/namespaces/function-return-inference.js
```diff
@@ -0,0 +1,37 @@
...
+/** @global */
+function* isGenerator() {
+}
+
+/**
+ * @global
+ * @return {string}
+ */
+function annotationOverride() {
+}
```
rictic: More tests:

```typescript
/** @global */
const voidArrowFunction = () => {reutn};

/** @global */
const nonVoidArrowFunction = () => 10;

/** @global */
const voidWithNonvoidInside = function() {
  let f = function () { return 10;};
  function g() { return 20;};
  class H { method() { return 30; } };
  const i = () => {return 40;};
};
```

src/test/static/namespaces/memberof-functions.js
```diff
@@ -13,7 +13,7 @@ Polymer.aaa = aaa;
  * @memberof Polymer
  */
 Polymer.bbb = function bbb() {
-
+  return;
```
rictic: Why add this?

src/javascript/esutil.ts
aomarks: Done.

src/javascript/esutil.ts
aomarks: Done.

src/javascript/function-scanner.ts
aomarks: Done.

src/test/static/namespaces/function-return-inference.js
aomarks: Done.

src/test/static/namespaces/memberof-functions.js
aomarks: Was to get more test coverage from this suite. But then I added my own suite, so I don't need it here too. Removed.

aomarks:
PTAL

PR #886, opened 2018-02-20
Fix #477 on master

### LINKED ISSUE #477: Unable to analyze polymer expressions which are not legal javascript
Notably `{{foo.*}}` and `{{arr.0}}` both work (though the latter only kinda, and by accident).

Currently the expression parser just ignores all such expressions entirely, which can lead to false negatives and false positives in linting.

boromisp:
This is a quick-and-dirty fix for #477.

It's based on property-effects parser code of Polymer (https://github.com/Polymer/polymer/blob/10aded461b1a107ed1cfc4a1d630149ad8508bda/lib/mixins/property-effects.html#L864).

It converts polymer expressions to valid javascript, before passing it to the AST parser (eg.:
```js
> transformPolymerExprToJS(
    "_compute(_myObject.kebab-array-prop.0, 'some-string', 1.0, _otherObj.*)"
  )
< _compute(_myObject['kebab-array-prop']['0'], 'some-string', 1.0, _otherObj)
```

### DIFF
--- a/src/polymer/expression-scanner.ts
+++ b/src/polymer/expression-scanner.ts
@@ -459,13 +459,71 @@ function findDatabindingInString(str: string) {
   return expressions;
 }
 
+function transformPath(expression: string) {
+  return expression
+    // replace .0, .123, .kebab-case with ['0'], ['123'], ['kebab-case']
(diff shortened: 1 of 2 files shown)


boromisp:
I signed it!

rictic (approved):

PR #887, opened 2018-02-20
Fix #477 on the 2.x branch

### LINKED ISSUE #477: Unable to analyze polymer expressions which are not legal javascript

boromisp:
This is a quick-and-dirty fix for #477.

It's based on property-effects parser code of Polymer (https://github.com/Polymer/polymer/blob/10aded461b1a107ed1cfc4a1d630149ad8508bda/lib/mixins/property-effects.html#L864).

It converts polymer expressions to valid javascript, before passing it to the AST parser (eg.:
```js
> transformPolymerExprToJS(
    "_compute(_myObject.kebab-array-prop.0, 'some-string', 1.0, _otherObj.*)"
  )
< _compute(_myObject['kebab-array-prop']['0'], 'some-string', 1.0, _otherObj)
```

### DIFF
--- a/src/polymer/expression-scanner.ts
+++ b/src/polymer/expression-scanner.ts
@@ -463,13 +463,70 @@ function findDatabindingInString(str: string) {
   return expressions;
 }
 
+function transformPath(expression: string) {
+  return expression
+    // replace .0, .123, .kebab-case with ['0'], ['123'], ['kebab-case']
+    .replace(/\.([a-zA-Z_$]([\w:$*]*-+[\w:$*]*)+|[1-9][0-9]*|0)/g, "['$1']")
+    // remove .* and .splices from the end of the paths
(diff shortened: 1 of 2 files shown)


boromisp:
I signed it!

rictic:
Do you have a need for this in the 2.x branch? Generally we haven't been backporting to 2.x unless something is high priority.

boromisp:
I can't really judge how useful it is for others, but it proved useful to me in my current project.
I had hundreds of false positive warnings with polymer lint (so I had to disable that rule), and this helped me find the few real bugs.

boromisp:
This PR is no longer relevant.

PR #888, opened 2018-02-21
Add support for bare module specifiers

justinfagnani:
Adds a constructor option to JavaScriptImportScanner that enables node-style module resolution. Resolution happens during scanning so that the generated ScannedImports can have URLs instead of un-resolved specifiers.

Module resolution option is currently disabled by default. It will need to be enabled for the time-being by Analyzer embedders. This should be done by exposing an option or flag to end users. The plan is to add a "web" option for resolution when such a thing exists.


src/test/javascript/javascript-import-scanner_test.ts
```diff
@@ -35,26 +36,81 @@ suite('JavaScriptImportScanner', () => {
...
   test('finds dynamic imports', async () => {
-    const {features} = await runScanner(
+    const {features, warnings} = await runScanner(
         analyzer,
-        new JavaScriptImportScanner(),
+        new JavaScriptImportScanner({moduleResolution: 'node'}),
         'javascript/dynamic-import.js');
 
+    if (warnings.length !== 1) {
+      console.log('AAA');
```
usergenic: Maybe you left in a quick-n-dirty-wtf-debug-output?

src/test/javascript/javascript-import-scanner_test.ts
```diff
@@ -35,26 +36,81 @@ suite('JavaScriptImportScanner', () => {
...
+  test('resolves bare specifiers', async () => {
+    const {features, warnings} = await runScanner(
         analyzer,
-        new JavaScriptImportScanner(),
+        new JavaScriptImportScanner({moduleResolution: 'node'}),
         'javascript/module-with-named-import.js');
 
-    assert.equal(features.length, 0);
+    if (warnings.length !== 0) {
+      console.log('AAA');
```
usergenic: I read this like 😱🙀"AAA!"

src/javascript/javascript-import-scanner.ts
```diff
@@ -36,44 +50,109 @@ export class JavaScriptImportScanner implements JavaScriptScanner {
...
+        sourceRange: document.sourceRangeForNode(node)!,
+        severity: Severity.WARNING,
+        code: 'cant-resolve-module-specifier',
+        parsedDocument: document,
+      }));
+      return undefined;
+    }
+
+    const documentPath = documentURL.pathname;
+    console.log('documentURL.pathname', documentURL.pathname);
```
usergenic: Drop the console.log

src/javascript/javascript-import-scanner.ts
```diff
@@ -36,44 +50,109 @@ export class JavaScriptImportScanner implements JavaScriptScanner {
...
+        message: 'Cannot resolve module specifier with no module resolution ' +
+            'algorithm set',
+        sourceRange: document.sourceRangeForNode(node)!,
+        severity: Severity.WARNING,
+        code: 'cant-resolve-module-specifier',
+        parsedDocument: document,
+      }));
+      return undefined;
+    }
+    const documentURL = new URL(document.url);
```
usergenic: If the JavaScriptDocument is an inline document from an HTML script tag, I think we would need to honor `baseUrl` in computing the relative URL from resolved module file of just `url`.  Does this document instance have a `baseUrl`?  If not, I think it might be in `document.parsedDocument.baseUrl`...

src/core/analyzer.ts
```diff
@@ -38,6 +38,8 @@ export interface Options {
    */
   lazyEdges?: LazyEdgeMap;
 
+  moduleResolution?: 'node';
```
rictic: Add a comment for the reader about what happens when this isn't specified

justinfagnani:
Sorry @usergenic I should have mentioned top-level that I'm pushing some commits to try to debug the failure on AppVeyor. I'll remove those commits when it's fixed.

src/javascript/javascript-import-scanner.ts
justinfagnani: Good point, I'll add a test case.

rictic: +1, yeah, and in cases where there isn't a baseUrl, it will be the same as document.url.

src/javascript/javascript-import-scanner.ts
```diff
@@ -36,44 +50,109 @@ export class JavaScriptImportScanner implements JavaScriptScanner {
...
+        message: 'Cannot resolve module specifier in non-local document',
+        sourceRange: document.sourceRangeForNode(node)!,
+        severity: Severity.WARNING,
+        code: 'cant-resolve-module-specifier',
+        parsedDocument: document,
+      }));
+      return undefined;
+    }
+
+    const documentPath = documentURL.pathname;
```
rictic: You want `vscode-uri`'s URI#fsPath here. I think it will solve the windows issues you're seeing.

src/javascript/javascript-import-scanner.ts
```diff
@@ -36,44 +50,112 @@ export class JavaScriptImportScanner implements JavaScriptScanner {
...
+            severity: Severity.WARNING,
+            code: 'non-literal-import',
+            parsedDocument: document,
+          }));
           return;
         }
         imports.push(new ScannedImport(
             'js-import',
-            source,
+            scanner._resolveSpecifier(
```
rictic: Why not emit a warning and return if we can't resolve the specifier?

Do we have a use for a ScannedImport without a url?

justinfagnani: The import still exists, we just couldn't determine a valid URL for it. It should be useful for linting. We don't produce an `Import` if `url` is undefined, which is consistent with current behavior, but I feel a little weird about that too.

PR #894, opened 2018-03-08
Fixed resolution of protocol-relative URLs.

usergenic:
- Fixes https://github.com/Polymer/polymer-analyzer/issues/893 where `UrlResolver#simpleUrlResolve` treated protocol-relative URL like `//something` as a path.

### DIFF
--- a/src/test/url-loader/package-url-resolver_test.ts
+++ b/src/test/url-loader/package-url-resolver_test.ts
@@ -72,7 +72,7 @@ suite('PackageUrlResolver', function() {
           resolvedUrl`http://abc.xyz/foo.html`);
       assert.equal(
           r.resolve(packageRelativeUrl`//abc.xyz/foo.html`),
-          resolvedUrl`file://abc.xyz/foo.html`);
+          resolvedUrl`//abc.xyz/foo.html`);
     });
 
     test(`resolves a URL with the right hostname`, () => {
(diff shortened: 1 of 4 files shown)


src/test/url-loader/package-url-resolver_test.ts
```diff
@@ -72,7 +72,7 @@ suite('PackageUrlResolver', function() {
           resolvedUrl`http://abc.xyz/foo.html`);
       assert.equal(
           r.resolve(packageRelativeUrl`//abc.xyz/foo.html`),
-          resolvedUrl`file://abc.xyz/foo.html`);
+          resolvedUrl`//abc.xyz/foo.html`);
```
rictic: Two more tests I'd like to see:

```typescript
       assert.equal(
          resolvedUrl`https://foo.com/bar.html`,
          r.resolve(fileRelativeUrl`//foo.com/bar.html`, resolvedUrl`https://baz.com/qux.js`));

       assert.equal(
          resolvedUrl`https://foo.com/bar.html`,
          r.resolve(fileRelativeUrl`https://foo.com/bar.html`, resolvedUrl`//baz.com/qux.js`));
```

rictic: The `//foo.com/bar.html` url isn't a resolved URL. I'd prefer that we resolve to `https://foo.com/bar.html` with an option to specify the protocol (via e.g. a server root url option)

PR #897, opened 2018-03-08
Parse JS to a `File` rather than a `Program`

rictic:
We need the `File` for babel traverse to do scope analysis.

This is a breaking change, but there's ~0% chance that anyone is actually broken by this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
-<!-- ## Unreleased -->
+## Unreleased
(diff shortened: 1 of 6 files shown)


rictic:
Oh, right, the stringification test. That's fixed by #896 I believe

CHANGELOG.md
```diff
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
-<!-- ## Unreleased -->
+## Unreleased
+* [breaking] JavascriptDocument#ast is now a `babel.File` rather than a
```
aomarks: below we yell `BREAKING`

rictic: I was deliberately quieter here because this is not going to break anyone :)

PR #898, opened 2018-03-09
Use a unified reference resolution system.

rictic:
Now using the same logic for resolving super classes, mixins, and behaviors.

Also parameterized the ScannedReference and Reference types for clarity.

 - [x] CHANGELOG.md not updated, internal refactoring.


rictic:
Thanks for the super quick review!

PR #902, opened 2018-03-09
Fix bug in package URL resolution due to startsWith containment check.

aomarks:
If a dependency name began with the name of the package it was being resolved from, we would not resolve the dependency to the components directory, because we were checking for containment naively with string `startsWith`.

Real life example:

  - `/repos/iron-icons/a.html` imports `../iron-iconset-svg/b.html`
  - We test `"/repos/iron-iconset-svg/b.html".startsWith("/repos/iron-icons")` and did not rewrite the url into the components directory, because it looked like it was already inside the package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,6 +93,7 @@
     "jsonschema": "^1.1.0",
     "minimatch": "^3.0.4",
     "parse5": "^4.0.0",
+    "path-is-inside": "^1.0.2",
     "polymer-project-config": "^3.6.0",
     "resolve": "^1.5.0",
     "shady-css-parser": "^0.1.0",
--- a/src/core/utils.ts
+++ b/src/core/utils.ts
@@ -12,7 +12,6 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
-import * as pathlib from 'path';
 import {parse as parseUrl_, Url} from 'url';
 
 const unspecifiedProtocol = '-:';
@@ -32,17 +31,6 @@ export function trimLeft(str: string, char: string): string {
     leftEdge++;
   }
   return str.substring(leftEdge);
-}
-
-/**
- * Returns whether the given file path points to a location inside the given
- * directory. Also return true if the paths are the same.
- */
-export function isPathInside(directory: string, file: string): boolean {
-  // If the path from the directory to the file does not require traversing up,
-  // then it must be either a descendent or the same directory. Note this is
-  // case-insensitive on Windows (which is what we want).
-  return !pathlib.relative(directory, file).startsWith('..');
 }
 
 export class Deferred<T> {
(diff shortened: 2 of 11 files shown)


src/url-loader/fs-url-loader.ts
```diff
@@ -30,15 +30,13 @@ export class FSUrlLoader implements UrlLoader {
   root: string;
 
   constructor(root: string = '') {
-    if (root.endsWith('/')) {
```
aomarks: Seemed like this was both wrong (shouldn't it be **not** endsWith?) and also unnecessary, because resolve normalizes the slash away anyway.

src/url-loader/fs-url-loader.ts
```diff
@@ -30,15 +30,13 @@ export class FSUrlLoader implements UrlLoader {
...
-      root += '/';
-    }
     this.root = pathlib.resolve(root);
   }
 
   canLoad(url: ResolvedUrl): boolean {
-    return url.startsWith('file://') &&
-        isPathInside(this.root, Uri.parse(url).fsPath);
+    const parsed = Uri.parse(url);
+    return parsed.scheme === 'file' && !parsed.authority &&
```
aomarks: Previously this was testing using `startsWith`, so if there was a host/authority (which we have test for), there would be a missing slash, and it worked out. Now we need to explicitly check there is no authority.

src/core/utils.ts
```diff
@@ -35,13 +36,13 @@ export function trimLeft(str: string, char: string): string {
...
-export function isPathInside(directory: string, filePath: string): boolean {
-  if (process.platform === 'win32') {
-    return filePath.toLowerCase().startsWith(directory.toLowerCase());
-  }
-  return filePath.startsWith(directory);
+export function isPathInside(directory: string, file: string): boolean {
+  // If the path from the directory to the file does not require traversing up,
+  // then it must be either a descendent or the same directory. Note this is
+  // case-insensitive on Windows (which is what we want).
+  return !pathlib.relative(directory, file).startsWith('..');
```
rictic: A small issue that probably won't come up:

I think `isPathInside('c:\\foo', 'd:\\bar')` returns true right now. Maybe we should add an extra check if we're on windows?

aomarks: Good catch. I've just switched to pull in path-is-inside, which already handles this case correctly. Note that it has the parameters in reverse order (which is the more intuitive order to me anyway). Also filed https://github.com/domenic/path-is-inside/pull/7 to add typings there, but added custom ones here for now.

package.json
```diff
@@ -93,6 +93,7 @@
     "jsonschema": "^1.1.0",
     "minimatch": "^3.0.4",
     "parse5": "^4.0.0",
+    "path-is-inside": "^1.0.2",
```
rictic: Cool, zero dependencies and written by Domenic

aomarks: Yup!

PR #906, opened 2018-03-11
Add specialized subclass JS Imports.

rictic:
This is a good place to add the original module specifiers, and may be useful hook for caching reference lookups.

### DIFF
--- a/src/javascript/javascript-import-scanner.ts
+++ b/src/javascript/javascript-import-scanner.ts
@@ -88,8 +88,10 @@ export class JavascriptImport extends Import {
   constructor(
       url: ResolvedUrl, originalUrl: FileRelativeUrl, type: string,
       document: Document, sourceRange: SourceRange|undefined,
-      urlSourceRange: SourceRange|undefined, ast: any, warnings: Warning[],
-      lazy: boolean, specifier: string) {
+      urlSourceRange: SourceRange|undefined,
+      ast: babel.ImportDeclaration|babel.CallExpression|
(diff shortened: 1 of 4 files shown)


src/javascript/javascript-import-scanner.ts
```diff
@@ -15,26 +15,95 @@
...
+export class JavascriptImport extends Import {
+  /**
+   * The original text of the specifier. Unlike `this.url`, this may not
+   * be a URL, but may be a bare module specifier, like 'jquery'.
+   */
+  readonly specifier: string;
+  constructor(
+      url: ResolvedUrl, originalUrl: FileRelativeUrl, type: string,
+      document: Document, sourceRange: SourceRange|undefined,
+      urlSourceRange: SourceRange|undefined, ast: any, warnings: Warning[],
```
aomarks: why is ast any?

rictic: copy-pasta from the definition of `Import`.

We should probably do a sweep of all uses of `any` in the analyzer.

src/model/import.ts
```diff
@@ -27,7 +27,7 @@ import {Severity, Warning} from './warning';
  * @template N The AST node type
  */
 export class ScannedImport implements Resolvable {
-  type: 'html-import'|'html-script'|'html-style'|'js-import'|string;
+  readonly type: 'html-import'|'html-script'|'html-style'|'js-import'|string;
```
justinfagnani: `keyof ImportFeatureKindMap` would be useful here. `FeatureKindMap` can extend it, and we'd reduce some duplication.

PR #908, opened 2018-03-11
Resolve references through imports

rictic:
This has all of the other PRs merged in. Will rebase to undo the merge commits after those other PRs land, or we can just review everything all at once here.

Scoped reference resolution could be enhanced further. e.g. it wouldn't be hard to resolve through const assignments. I've got that working in one of my prototype branches actually.

### DIFF
--- a/src/model/reference.ts
+++ b/src/model/reference.ts
@@ -14,6 +14,7 @@
 
 import {NodePath} from 'babel-traverse';
 import * as babel from 'babel-types';
+import * as util from 'util';
 
 import * as esutil from '../javascript/esutil';
 import {Annotation} from '../javascript/jsdoc';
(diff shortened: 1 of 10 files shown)


src/model/reference.ts
```diff
@@ -56,39 +59,149 @@ export class ScannedReference<K extends keyof FeatureKindMap> extends
...
-    let feature: undefined|FeatureKindMap[DK];
-    [feature] = features;
     return new Reference<FeatureKindMap[K]>(this, feature, warnings);
   }
 }
 
+function resolveScopedAt<K extends keyof FeatureKindMap>(
+    path: NodePath, identifier: string, document: Document, kind: K):
+    Result<FeatureKindMap[K], Warning|undefined> {
+  if (isSomeKindOfImport(path)) {
```
justinfagnani: I'd add some comments to describe the cases that this is working though

src/model/reference.ts
```diff
@@ -56,39 +59,149 @@ export class ScannedReference<K extends keyof FeatureKindMap> extends
...
+  const features = document.getFeatures({kind, id: identifier, statement});
+  if (features.size > 1) {
+    // TODO(rictic): narrow down by identifier? warn?
+    return {successful: false, error: undefined};
+  }
+  const [feature] = features;
+  if (feature !== undefined) {
+    return {successful: true, value: feature};
+  }
+  if (/^[^\.]+\.[^\.]+$/.test(identifier)) {
```
justinfagnani: put the regex in a variable whose name describes what it matches

PR #909, opened 2018-03-12
Calculate identifiers for re-exports correctly.

rictic:
- [x] CHANGELOG.md not updated, nothing interesting for a user yet.

### DIFF
--- a/src/javascript/javascript-export-scanner.ts
+++ b/src/javascript/javascript-export-scanner.ts
@@ -54,11 +54,8 @@ export class Export implements Resolvable, Feature {
     this.statementAst = statementAst;
     let exportedIdentifiers;
     if (exportingAllFrom !== undefined) {
-      exportedIdentifiers = (function*() {
-        for (const export_ of exportingAllFrom) {
-          yield* export_.identifiers;
-        }
-      })();
+      exportedIdentifiers =
+          flatMap(exportingAllFrom, (export_) => export_.identifiers);
     } else {
       exportedIdentifiers = esutil.getBindingNamesFromDeclaration(astNode);
     }
@@ -93,6 +90,14 @@ export class Export implements Resolvable, Feature {
     // arbitrarily long chain of references.
 
     return this;
+  }
+}
+
+function*
+    flatMap<In, Out>(inputs: Iterable<In>, map: (input: In) => Iterable<Out>):
+        Iterable<Out> {
+  for (const input of inputs) {
+    yield* map(input);
   }
 }
 
--- a/src/javascript/javascript-export-scanner.ts
+++ b/src/javascript/javascript-export-scanner.ts
@@ -48,10 +48,21 @@ export class Export implements Resolvable, Feature {
 
   constructor(
       astNode: ExportNode, statementAst: babel.Statement,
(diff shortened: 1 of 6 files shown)


src/javascript/javascript-export-scanner.ts
```diff
@@ -48,24 +48,50 @@ export class Export implements Resolvable, Feature {
...
       astNode: ExportNode, statementAst: babel.Statement,
-      sourceRange: SourceRange|undefined, nodePath: NodePath<babel.Node>) {
+      sourceRange: SourceRange|undefined, nodePath: NodePath<babel.Node>,
+      exportingAllFrom?: Iterable<Export>) {
     this.astNode = astNode;
     this.statementAst = statementAst;
-    for (const name of esutil.getBindingNamesFromDeclaration(astNode)) {
+    let exportedIdentifiers;
+    if (exportingAllFrom !== undefined) {
+      exportedIdentifiers = (function*() {
```
justinfagnani: I don't think we're using IterableX here yet, but `flatMap` would be understandable here for functional folk.

src/javascript/javascript-export-scanner.ts
```diff
@@ -48,24 +48,50 @@ export class Export implements Resolvable, Feature {
...
-    //   resolution algorithm here. Especially in cases of re-export.
-    //   would need to separate out ScannedExport from Export in that case.
+  // TODO: Could potentially get a speed boost by doing the Reference
+  //   resolution algorithm here, rather than re-doing it every single place
+  //   this export is referenced.
+  resolve(document: Document): Feature|undefined {
+    if (babel.isExportAllDeclaration(this.astNode)) {
+      const [import_] =
+          document.getFeatures({kind: 'import', statement: this.statementAst});
+      if (import_ === undefined) {
```
justinfagnani: I'd like to make this not be the case... I think Imports should be features whether they resolve or not, but generate a warning and have an undefined URL. Worth taking not of where we make this assumption.

src/test/javascript/javascript-export-scanner_test.ts
```diff
@@ -40,7 +40,31 @@ suite('JavaScriptExportScanner', () => {
...
+      ['anotherValue'],
+      ['someValue']
     ]);
   });
+
+  test('re-exports across multiple files correctly', async () => {
+    const features = await getExports('javascript/re-export-all.js');
+    // Like the list above, but flattened as they're all exported from one
+    // export statement.
+    assert.containSubset([...features].map((f) => [...f.identifiers]), [[
```
justinfagnani: is `containSubset` the right check? Why not filter for js-exports and assert set equality?

src/javascript/javascript-export-scanner.ts
rictic: Done, added a local definition of flatMap and called that here.

src/javascript/javascript-export-scanner.ts
rictic: I'm cool with that. The type checker will save us here when we make that change.

src/test/javascript/javascript-export-scanner_test.ts
rictic: Good catch, this should just be deepEqual. Changed to deepEqual here and elsewhere.

PR #910, opened 2018-03-12
Do scoped resolution of same-file references.

rictic:
This covers just the simplest kinds of scoped references. It doesn't follow reassignments, imports, or re-exports. Will add each of these in follow up PRs.

### DIFF
--- a/src/model/reference.ts
+++ b/src/model/reference.ts
@@ -12,11 +12,13 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
-import {NodePath} from 'babel-traverse';
+import {Binding, NodePath} from 'babel-traverse';
 import * as babel from 'babel-types';
 
+import * as esutil from '../javascript/esutil';
 import {Annotation} from '../javascript/jsdoc';
 
+import {Result} from './analysis';
 import {Document} from './document';
 import {Feature, ScannedFeature} from './feature';
 import {FeatureKindMap} from './queryable';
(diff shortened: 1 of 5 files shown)


src/model/reference.ts
```diff
@@ -56,39 +58,70 @@ export class ScannedReference<K extends keyof FeatureKindMap> extends
...
+    let feature: undefined|FeatureKindMap[DK];
     const warnings = [...this.warnings];
-    if (this.sourceRange) {
-      if (features.size === 0) {
-        let message = `Could not resolve reference to ${this.kind}`;
-        if (kind === 'behavior') {
-          message += `. Is it annotated with @polymerBehavior?`;
+
+    const binding = this.astPath.scope.getBinding(this.identifier);
+    if (binding !== undefined) {
```
justinfagnani: We should file an issue to make this more principled. We can know at scanning time whether a binding was to a locally defined name, and only look up via scope analysis in that case. We can also attempt both look up styles and report when resolution is ambiguous, though you pointed out that this would be verbose for classes, which always add themselves to the global identifier map. We should also file an issue to fix that and only add to namespaces if annotated with `@memberof`, etc.

src/model/reference.ts
```diff
@@ -56,39 +58,70 @@ export class ScannedReference<K extends keyof FeatureKindMap> extends
...
     return new Reference<FeatureKindMap[K]>(this, feature, warnings);
   }
 }
 
+function resolveBinding<K extends keyof FeatureKindMap>(
+    binding: Binding, document: Document, kind: K):
+    Result<FeatureKindMap[K], Warning|undefined> {
+  const statement = esutil.getCanonicalStatement(binding.path);
+  if (!statement) {
+    return {successful: false, error: undefined};
```
justinfagnani: Add some kind of an error? In the above code it'll be ignored, but in a case where a caller does expect this to be resolved, they might want to know why it failed.

src/model/reference.ts
```diff
@@ -56,39 +58,70 @@ export class ScannedReference<K extends keyof FeatureKindMap> extends
...
-    [feature] = features;
     return new Reference<FeatureKindMap[K]>(this, feature, warnings);
   }
 }
 
+function resolveBinding<K extends keyof FeatureKindMap>(
+    binding: Binding, document: Document, kind: K):
+    Result<FeatureKindMap[K], Warning|undefined> {
+  const statement = esutil.getCanonicalStatement(binding.path);
+  if (!statement) {
```
justinfagnani: do an explicit comparison

src/model/reference.ts
rictic: Filed https://github.com/Polymer/polymer-analyzer/issues/918 and https://github.com/Polymer/polymer-analyzer/issues/917

src/model/reference.ts
rictic: Did the hookup here, filed https://github.com/Polymer/polymer-analyzer/issues/919 for actually emitting them, which might mean passing more context down.

rictic:
Comments addressed in #908 which subsumes this PR.

PR #914, opened 2018-03-15
Stop warning for unloadable documents if we dont expect to be able to load them

usergenic:
* Fixed an issue where warnings were generated for imports of resolved URLs
  which are not loadable and external to the package.  Now we warn only when
  unable to load a document we expect to be loadable.

### DIFF
--- a/src/html/html-script-tag.ts
+++ b/src/html/html-script-tag.ts
@@ -33,6 +33,7 @@ export class ScannedScriptTagImport extends ScannedImport {
         !document._analysisContext.loader.canLoad(resolvedUrl)) {
       return;
     }
+    console.log('i have a resolvedUrl', resolvedUrl);
 
     // TODO(justinfagnani): warn if the same URL is loaded from more than one
     // non-module script tag
--- a/src/html/html-script-tag.ts
+++ b/src/html/html-script-tag.ts
@@ -12,7 +12,7 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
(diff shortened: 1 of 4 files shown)


usergenic:
Side note: this is a problem because polymer-build is sensitive to (aka fails the build) when analyzer warns that it `could-not-load` an import.

This is a new behavior in a way because it used to be that the resolver would return undefined if something wasn't expected to be loadable.

usergenic:
*Removed review requests* because I added a test and this isn't working still.

usergenic:
Added back review requests because test is good now.  (I forgot in memory overlay URL loader always reports canLoad = true so I had to patch its canLoad method in test.)

usergenic:
@rictic PTAL.  I also attempted to DRY up various warning logic around imports.

src/html/html-script-tag.ts
```diff
@@ -24,13 +24,11 @@ export class ScriptTagImport extends Import { readonly type = 'html-script'; }
...
   resolve(document: Document): ScriptTagImport|undefined {
-    if (this.url === undefined) {
-      return;
-    }
-    const resolvedUrl = document._analysisContext.resolver.resolve(
-        document.parsedDocument.baseUrl, this.url, this);
-    if (resolvedUrl === undefined) {
-      return;
+    const resolvedUrl = this.getLoadableUrlOrWarn(document);
+    if (this.url === undefined || resolvedUrl === undefined) {
```
rictic: One wait to avoid this duplication would be to return this.url _and_ resolvedUrl from getLoadableUrlOrWarn

rictic:
+1 to the DRYing

PR #915, opened 2018-03-16
Prepare CHANGELOG for 3.0.0-pre.16 release.

usergenic:
## [3.0.0-pre.16] - 2018-03-15
 * Import URLs which are resolved, but for which the URL loader returns
   `canLoad() === false` result in a `not-loadable` Warning instead of
   `could-not-load`, which is reserved for other failures, such as parser
   errors or load errors.


usergenic:
Travis down.  Merging anyway after successful Appveyor test run and local test run.

PR #927, opened 2018-03-26
Import features will now be produced when a document has a broken import

rictic:


src/test/html/html-script-scanner_test.ts
```diff
@@ -126,25 +124,31 @@ suite('HtmlScriptScanner', () => {
           `import * as somethingElse from './javascript/other-module-with-export.js';`);
     });
 
-    test('follows import statements in modules', async () => {
+    test.only('follows import statements in modules', async () => {
```
TimvdLippe: Remove the `only` here

PR #928, opened 2018-03-27
Extract node module resolution from JavascriptImport into a separate module

### LINKED ISSUE #921: import `./foo` doesn't convert to `./foo.js`
Need to use node resolution algorithm for resolving relative modules - `./foo` could be `./foo.js` or `./foo/index.js`. Affects polyserve.

cc @usergenic

justinfagnani:
Fixes #921 

### DIFF
--- a/src/javascript/javascript-import-scanner.ts
+++ b/src/javascript/javascript-import-scanner.ts
@@ -13,8 +13,7 @@
  */
 
 import * as babel from 'babel-types';
-import {dirname, relative} from 'path';
-import {URL} from 'whatwg-url';
+import {parseURL, URL} from 'whatwg-url';
 
 import {Document, FileRelativeUrl, Import, ResolvedUrl, ScannedImport, Severity, Warning} from '../model/model';
 
@@ -22,9 +21,9 @@ import {Visitor} from './estree-visitor';
 import * as esutil from './esutil';
 import {JavaScriptDocument} from './javascript-document';
 import {JavaScriptScanner} from './javascript-scanner';
+import {resolve} from './resolve-specifier-node';
 
 import isWindows = require('is-windows');
-import resolve = require('resolve');
 import {SourceRange} from '../model/model';
 import {DeclaredWithStatement} from '../model/document';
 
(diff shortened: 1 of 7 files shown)


src/javascript/javascript-import-scanner.ts
```diff
@@ -254,32 +236,34 @@ export class JavaScriptImportScanner implements JavaScriptScanner {
...
+      try {
+        // TODO (justinfagnani): we need to pass packageName and componentDir,
+        // or maybe the UrlResolver, to get correct relative URLs for
+        // named imports from the root package to its dependencies
+        return resolve(specifier, documentPath);
+      } catch (e) {
+        if (!isPathSpecifier(specifier)) {
+          // Don't warn if the specifier was already a path, even though we do
+          // resolve paths, because maybe the user is serving it some other way.
+          console.warn(`Could not resolve module specifier "${specifier}"`);
```
rictic: The comment says that we don't warn, but the code does warn. Is this debug code

IMO we should not log to console, we should emit a Warning or throw an Error to express a problem.

src/javascript/resolve-specifier-node.ts
```diff
@@ -0,0 +1,97 @@
...
+              return packageJson;
+            },
+      });
+
+      let relativeSpecifierUrl =
+          relative(dirname(documentPath), resolvedSpecifier) as FileRelativeUrl;
+
+      if (componentInfo !== undefined) {
+        const {packageName, rootDir, componentDir} = componentInfo;
+        console.log(
```
rictic: console log

src/javascript/resolve-specifier-node.ts
```diff
@@ -0,0 +1,97 @@
...
+/**
+ * Resolves module specifiers using node module resolution.
+ *
+ * Full URLs - those parsable by the WHATWG URL spec - are returned as-is.
+ * Absolute and relative paths are resolved, even though they are valid
+ * HTML-spec module specifiers, because node resolution supports directories
+ * and omitting extensions. If a specifier doesn't resolve, it's returned as-is.
+ */
+export const resolve =
+    (specifier: string, documentPath: string, componentInfo?: {
```
rictic: Document what `componentInfo` means?

src/javascript/javascript-import-scanner.ts
aomarks: The comment applies to the if statement (maybe clearer to say "Only warn if the specifier was a name")

justinfagnani:
PTAL

src/javascript/resolve-specifier-node.ts
```diff
@@ -0,0 +1,102 @@
...
+ *
+ * Full URLs - those parsable by the WHATWG URL spec - are returned as-is.
+ * Absolute and relative paths are resolved, even though they are valid
+ * HTML-spec module specifiers, because node resolution supports directories
+ * and omitting extensions. If a specifier doesn't resolve, it's returned as-is.
+ *
+ * @param componentInfo An object describing a "component-style" URL layout. In
+ *   this layout, cross-package URLs reach out of the package directory to
+ *   sibling packages, rather than into the component directory. When given,
+ *   this parameter causes relative paths to be returns for this style.
```
aomarks: to be returned

src/javascript/javascript-import-scanner.ts
```diff
@@ -254,32 +236,30 @@ export class JavaScriptImportScanner implements JavaScriptScanner {
...
-      relativeSpecifierUrl =
-          relativeSpecifierUrl.replace(/\\/g, '/') as FileRelativeUrl;
+    if (this.moduleResolution === 'node') {
+      try {
+        // TODO (justinfagnani): we need to pass packageName and componentDir,
+        // or maybe the UrlResolver, to get correct relative URLs for
+        // named imports from the root package to its dependencies
+        return resolve(specifier, documentPath);
+      } catch (e) {
+        // If the specifier was a name, we'll emit a warning below.
```
rictic: The warning emitted below though is too specific, it's just for when there's no module resolution algorithm.

src/javascript/resolve-specifier-node.ts
rictic: This has really heavy overlap with PackageUrlResolver. Can we unify?

PR #929, opened 2018-03-27
Support behaviors declared in export statements.

rictic:
Also cleans up some super sketchy code that was almost definitely silently crashing on some code.

### DIFF
--- a/src/polymer/behavior-scanner.ts
+++ b/src/polymer/behavior-scanner.ts
@@ -290,13 +290,24 @@ class BehaviorVisitor implements Visitor {
  * gets the expression representing a behavior from a node.
  */
 function behaviorExpression(node: babel.Node): babel.Node|null|undefined {
+  if (babel.isVariableDeclarator(node)) {
+    return node.init;
+  }
+  if (babel.isAssignmentExpression(node)) {
+    return node.right;
+  }
+  if (babel.isExportDefaultDeclaration(node) ||
+      babel.isExportNamedDeclaration(node)) {
+    return behaviorExpression(node.declaration);
+  }
   if (babel.isExpressionStatement(node)) {
-    // need to cast to `any` here because ExpressionStatement is super
-    // super general. this code is suspicious.
-    return (node as any).expression.right;
+    return behaviorExpression(node.expression);
   }
   if (babel.isVariableDeclaration(node)) {
-    return node.declarations.length > 0 ? node.declarations[0].init : undefined;
+    return behaviorExpression(node.declarations[0]);
+  }
+  if (babel.isObjectExpression(node) || babel.isArrayExpression(node)) {
+    return node;
   }
 }
 
--- a/src/test/polymer/behavior-scanner_test.ts
+++ b/src/test/polymer/behavior-scanner_test.ts
@@ -130,4 +130,18 @@ suite('BehaviorScanner', () => {
           {name: 'getterSetter', type: undefined, readOnly: false}
         ]);
   });
+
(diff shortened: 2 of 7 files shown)


src/test/polymer/behavior-scanner_test.ts
```diff
@@ -130,4 +130,18 @@ suite('BehaviorScanner', () => {
...
+    const analysis = await analyzer.analyze(['uses-behaviors.js']);
+    const elements = [...analysis.getFeatures({kind: 'polymer-element'})];
+    assert.deepEqual(
+        elements.map((e) => e.tagName),
+        ['uses-basic-behavior', 'uses-array-behavior']);
+
+    assert.deepEqual(
+        elements.map((e) => [...e.methods.keys()]),
+        [['method1'], ['method1', 'method2']]);
+  });
```
aomarks: can you check that we also know explicitly which behaviors are applied?

src/polymer/behavior-scanner.ts
```diff
@@ -290,13 +290,24 @@ class BehaviorVisitor implements Visitor {
  * gets the expression representing a behavior from a node.
  */
 function behaviorExpression(node: babel.Node): babel.Node|null|undefined {
+  if (babel.isVariableDeclarator(node)) {
```
aomarks: So this function is only called when the node has been explicitly annotated with `@polymerBehavior`, right?

aomarks: It would be nice to have test coverage for all these syntaxes.

rictic: All of these did show up in our various tests, though not all in the behaviors tests

rictic: Yes

src/test/polymer/behavior-scanner_test.ts
rictic: Done

src/polymer/behavior-scanner.ts
aomarks: But no other tests changed so how did this work before?

src/test/polymer/behavior-scanner_test.ts
```diff
@@ -130,4 +130,24 @@ suite('BehaviorScanner', () => {
...
+    const analysis = await analyzer.analyze(['uses-behaviors.js']);
+    const elements = [...analysis.getFeatures({kind: 'polymer-element'})];
+    assert.deepEqual(
+        elements.map((e) => e.tagName),
+        ['uses-basic-behavior', 'uses-array-behavior']);
+
+    // Get the toplevel behaviors.
+    assert.deepEqual(
+        elements.map((e) => e.behaviorAssignments.map(ba => ba.identifier)),
+        [['BasicBehavior1'], ['ArrayOfBehaviors']]);
```
aomarks: I think I would have expected the second one to be flattened to `[BasicBehavior1, BasicBehavior2]`, since `ArrayOfBehaviors` is not a behavior. At least, the TypeScript generator assumes that this is flat, so it won't work well there. Will Analyzer let me look up what `ArrayOfBehaviors` is somehow?

src/polymer/behavior-scanner.ts
rictic: That's a damn good question. I added an explicit test of a default exported behavior.

src/test/polymer/behavior-scanner_test.ts
rictic: ArrayOfBehaviors _is_ a Behavior.

aomarks: Ah I see.

PR #935, opened 2018-03-31
Update to the latest version of babel.

rictic:

The biggest change here is that SpreadProperty was renamed to SpreadElement, but the types haven't been updated yet, so I needed to swap some of the type comparisons around from filtering out to filtering in.

### DIFF
--- a/src/javascript/esutil.ts
+++ b/src/javascript/esutil.ts
@@ -92,6 +92,10 @@ export function getPropertyName(prop: PropertyOrMethod): string|undefined {
   return undefined;
 }
 
+/**
+ * Yields properties and methods, filters out spread expressions or anything
+ * else.
+ */
 export function* getSimpleObjectProperties(node: babel.ObjectExpression) {
   for (const property of node.properties) {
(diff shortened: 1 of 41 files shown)


src/javascript/esutil.ts
```diff
@@ -96,6 +92,14 @@ export function getPropertyName(prop: PropertyOrMethod): string|undefined {
   return undefined;
 }
 
+export function* getSimpleObjectProperties(node: babel.ObjectExpression) {
```
aomarks: document what "simple" means here, or call it something more meaningful

src/javascript/esutil.ts
```diff
@@ -598,17 +600,15 @@ export function extractPropertiesFromClassOrObjectBody(
...
 
   for (const member of body) {
-    if (babel.isSpreadProperty(member)) {
+    if (!babel.isMethod(member) && !babel.isObjectProperty(member)) {
       continue;
     }
 
-    if ((babel.isMethod(member) || babel.isObjectProperty(member)) &&
-        member.computed) {
+    const name = getPropertyName(member)!;
```
aomarks: do you need the `!` since you check undefined below?

PR #941, opened 2018-04-07
Fix own-package component directory handling in node resolution.

aomarks:
I'm not 100% sure this is right. This does make the tests in https://github.com/Polymer/polyserve/pull/266 pass (after I added another case where I think the golden was wrong). I changed the approach slightly, and tried to think about it more directly in terms of what polyserve is doing:

If you were served from a `/components/` URL (which we can tell because a `componentInfo` was passed in), but your document path is not inside the component dir on _disk_, then this must be the special "own package" case where you are serving some package root as though it were a component sibling. We can then get the right relative path just by imagining that the file on disk actually was located in the components directory, and then computing the relative path as normal.

Part of https://github.com/Polymer/polymer-cli/issues/993

### DIFF
--- a/src/javascript/resolve-specifier-node.ts
+++ b/src/javascript/resolve-specifier-node.ts
@@ -70,16 +70,20 @@ export const resolve =
       let effectiveDocumentPath = documentPath;
       if (componentInfo !== undefined) {
         const {packageName, rootDir, componentDir} = componentInfo;
-        const isOwnPackageRequest = !pathIsInside(documentPath, componentDir);
-        if (isOwnPackageRequest) {
+        const isRootPackageRequest = !pathIsInside(documentPath, componentDir);
+        if (isRootPackageRequest) {
+          // Special handling for servers like Polyserve which, when serving a
+          // package "foo", will map the URL "/components/foo" to the root
+          // package directory, so that "foo" can make correct relative path
+          // references to its dependencies.
+          const rootRelativePath = relative(rootDir, documentPath);
           effectiveDocumentPath =
-              join(componentDir, packageName, relative(rootDir, documentPath));
+              join(componentDir, packageName, rootRelativePath);
         }
       }
 
-      let relativeSpecifierUrl =
-          relative(dirname(effectiveDocumentPath), resolvedSpecifier) as
-          FileRelativeUrl;
+      let relativeSpecifierUrl = relative(
+          dirname(effectiveDocumentPath), resolvedSpecifier) as FileRelativeUrl;
 
       if (isWindows()) {
         // normalize path separators to URL format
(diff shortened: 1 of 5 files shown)


src/javascript/resolve-specifier-node.ts
```diff
@@ -67,26 +67,20 @@ export const resolve =
...
-              relative(join(rootDir, packageName), rootDir);
-          const relativePathFromComponentDir =
-              relative(componentDir, resolvedSpecifier);
-          relativeSpecifierUrl =
-              join(
-                  componentDirRelativeToPackage,
-                  relativePathFromComponentDir) as FileRelativeUrl;
+        const isOwnPackageRequest = !pathIsInside(documentPath, componentDir);
+        if (isOwnPackageRequest) {
+          effectiveDocumentPath =
```
justinfagnani: let's add a comment here explaining the component URL mapping. Something like "component requests map root package URLs so that they appear to be (siblings of dependencies | paths within the components folder)."

src/javascript/resolve-specifier-node.ts
```diff
@@ -67,26 +67,20 @@ export const resolve =
...
+              join(componentDir, packageName, relative(rootDir, documentPath));
```
justinfagnani: I'd break this into two statements so the variable names help document what's going on. First:

```ts
rootRelativePath = relative(rootDir, documentPath);
```

src/javascript/resolve-specifier-node.ts
```diff
@@ -67,26 +67,20 @@ export const resolve =
...
-        if (pathIsInside(resolvedSpecifier, componentDir)) {
-          const componentDirRelativeToPackage =
-              relative(join(rootDir, packageName), rootDir);
-          const relativePathFromComponentDir =
-              relative(componentDir, resolvedSpecifier);
-          relativeSpecifierUrl =
-              join(
-                  componentDirRelativeToPackage,
-                  relativePathFromComponentDir) as FileRelativeUrl;
+        const isOwnPackageRequest = !pathIsInside(documentPath, componentDir);
```
justinfagnani: "own" doesn't feel like the right term here. How about `isRootPackageRequest`?

src/javascript/resolve-specifier-node.ts
aomarks: Done.

src/javascript/resolve-specifier-node.ts
aomarks: Done.

src/javascript/resolve-specifier-node.ts
aomarks: Done.

aomarks:
> I do think we need a regression test at least, and I'd prefer a first unit test on resolve, which will make it easier to move into a separate package in the monorepo if we want.

Done.

